PR #5, opened 2026-02-23
fix(gateway): add 64KB request body size limit

amilz:
## Summary
- Enforce a 64 KB cap on POST request bodies in the JSON-RPC gateway to prevent OOM under large payloads
- Two-layer defense: Content-Length pre-check (fast path) + `http_body_util::Limited` wrapper (catches chunked / missing headers)
- Error downcast distinguishes `LengthLimitError` (413) from genuine IO errors (400); unparseable `Content-Length` headers are logged as warnings
- 413 responses now include a JSON-RPC error body with the limit value for client debuggability

## Test Plan
Five integration tests using raw TCP streams (no mock required):
- `rejects_content_length_over_64kb` — oversized Content-Length header rejected before reading body
- `rejects_oversized_body_without_content_length` — chunked body over limit rejected by Limited wrapper
- `accepts_body_at_exactly_64kb` — boundary: exactly 64 KB is NOT rejected as 413
- `rejects_oversized_body_despite_small_content_length` — lying Content-Length (100) with 65 KB body is rejected
- `accepts_normal_sized_request` — normal request passes size gate (502 since no backend in tests)

Run with: `cargo test -p contra-gateway`

Closes PRO-890

### DIFF
--- a/gateway/src/lib.rs
+++ b/gateway/src/lib.rs
@@ -80,7 +80,11 @@ impl Gateway {
             Some(bytes) => {
                 builder = builder.header("Content-Type", "application/json");
                 builder
-                    .body(Full::new(bytes).map_err(|never| match never {}).boxed_unsync())
+                    .body(
+                        Full::new(bytes)
+                            .map_err(|never| match never {})
+                            .boxed_unsync(),
+                    )
                     .unwrap()
             }
             None => builder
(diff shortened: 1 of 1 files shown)


gateway/src/lib.rs
```diff
@@ -220,19 +255,46 @@ impl Gateway {
...
-                    )
-                    .body(Empty::new().map_err(|never| match never {}).boxed_unsync())
-                    .unwrap())
+                Ok(self.error_response(StatusCode::BAD_GATEWAY, None))
             }
         }
     }
 }
 
+#[cfg(test)]
```
dev-jodee: think this should be in the cfg test at the bottom, keep everything test related together

PR #14, opened 2026-02-24
feat(indexer): add startup reconciliation against on-chain escrow ATA balances [PRO-920]

dev-jodee:
## Summary

- On escrow indexer startup, before any data processing, optionally verify that per-mint DB totals (completed deposits − completed withdrawals) match the actual token balances in the escrow ATAs on-chain
- Mismatches exceeding `mismatch_threshold_raw` emit a `reconciliation_alert=true` tracing event and block startup; mismatches within tolerance log a warning and continue
- Opt-in (`enabled: false` by default) so existing deployments are unaffected — no config changes required to upgrade

## Changes

**New files**
- `indexer/src/indexer/reconciliation.rs` — core reconciliation logic with unit tests
- `indexer/src/storage/common/storage/get_mint_balances_for_reconciliation.rs` — storage routing

**Modified files**
- `indexer/src/bin/indexer.rs` — wires `ReconciliationConfig` from TOML/env (`INDEXER_RECONCILIATION_*`)
- `indexer/src/config.rs` — adds `ReconciliationConfig` struct and field on `IndexerConfig`
- `indexer/src/error/indexer.rs` — adds `ReconciliationError` enum and `IndexerError::Reconciliation` variant
- `indexer/src/indexer/indexer.rs` — calls `run_startup_reconciliation` after schema init, before backfill/datasource
- `indexer/src/storage/common/models.rs` — adds `MintDbBalance` struct
- `indexer/src/storage/common/storage.rs` / `mock.rs` / `postgres/db.rs` — storage layer support
- `indexer/config/example-indexer.toml` — documents `[indexer.reconciliation]` section

## Notes

No metrics library exists in the codebase; alerting is via structured tracing (`reconciliation_alert = true` field) which is filterable in any log aggregator.

### DIFF
--- a/indexer/src/indexer/reconciliation.rs
+++ b/indexer/src/indexer/reconciliation.rs
@@ -1,0 +1,947 @@
+//! Startup reconciliation of DB state against on-chain escrow ATA balances.
+//!
+//! On startup, before processing any new data, the escrow indexer verifies that its
+//! stored deposit/withdrawal totals match the actual token balances held in the escrow
+//! instance's Associated Token Accounts (ATAs).
+//!
+//! The DB-side formula mirrors exactly what is on-chain:
+//!   `db_expected = all_indexed_deposits − completed_withdrawals`
+//!
+//! Deposits increase the ATA balance on-chain the moment they are observed, regardless of
+//! the operator's contra minting status (`pending`/`processing`/`completed`/`failed`).
+//! Only completed withdrawals (`release_funds`) reduce the ATA balance.
+//!
+//! Flow:
+//! 1. Query the DB for per-mint aggregate balances (all deposits − completed withdrawals).
+//! 2. Derive the escrow ATA for each mint (instance PDA + mint + token program).
+//! 3. Fetch the live ATA balance via RPC.
+//! 4. If any |on_chain - db_expected| > threshold → log error, emit alert, abort startup.
+//! 5. If any mismatch ≤ threshold (but > 0) → log warning, continue.
+//! 6. If all balanced → log info, continue.
+
+use crate::{
+    config::{ProgramType, ReconciliationConfig},
+    error::{IndexerError, ReconciliationError},
+    operator::{
+        account_util::find_instance_pda, rpc_util::RpcClientWithRetry, RetryConfig, RetryPolicy,
+    },
+    storage::Storage,
+};
+use contra_core::rpc::error::INVALID_PARAMS_CODE;
+use solana_sdk::{commitment_config::CommitmentConfig, pubkey::Pubkey};
+use spl_associated_token_account::get_associated_token_address_with_program_id;
+use std::str::FromStr;
+use tracing::{error, info, warn};
+
+/// Per-mint result produced during reconciliation.
+#[derive(Debug, Clone)]
+pub struct MintReconciliation {
+    pub mint: String,
+    /// Expected balance according to DB: all indexed deposits − completed withdrawals.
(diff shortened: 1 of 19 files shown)


indexer/src/indexer/reconciliation.rs
```diff
@@ -0,0 +1,757 @@
...
+/// Uses structured `ErrorKind` inspection rather than raw string matching:
+/// - Primary: error code `-32602` (JSON-RPC "Invalid params"), which is what
+///   Solana validators return for a missing account via `getTokenAccountBalance`.
+/// - Fallback: substring match on the error message for non-standard RPC
+///   providers that may emit the same wording with a different code.
+fn is_account_not_found(e: &solana_rpc_client_api::client_error::Error) -> bool {
+    use solana_rpc_client_api::{client_error::ErrorKind, request::RpcError};
+
+    if let ErrorKind::RpcError(RpcError::RpcResponseError { code, message, .. }) = &e.kind {
+        if *code == -32602 {
```
amilz: I think this should be `3230000`? [ref](https://github.com/anza-xyz/kit/blob/69a380bada7881be085b9c49bddabbb026edd7f7/packages/errors/src/codes.ts#L84C58-L84C65). can add hardcode to `core/src/rpc/error.rs` for better naming/readability.
otherwise, want to use `core/src/rpc/error.rs` imports from `jsonrpsee::types::error`, `INVALID_PARAMS_CODE`?

per my comment above...i worry this could result in a false trigger above depending on how we catch the error here.

indexer/src/indexer/reconciliation.rs
```diff
@@ -0,0 +1,757 @@
...
+    let rpc = rpc_client.rpc_client.clone();
+    let ata = *ata;
+
+    let result = rpc_client
+        .with_retry("get_token_account_balance", RetryPolicy::Idempotent, || {
+            let rpc = rpc.clone();
+            async move {
+                match rpc.get_token_account_balance(&ata).await {
+                    Ok(ui_amount) => Ok(Some(ui_amount)),
+                    Err(e) if is_account_not_found(&e) => Ok(None),
```
amilz: is there any reason that `is_account_not_found` could be triggered for some DB issue rather than uninitialized account (see comment below)?
if so, we might need to at least warn...but i think if we have clearer code handling we'll be good.

indexer/src/indexer/reconciliation.rs
```diff
@@ -0,0 +1,757 @@
...
+        total_mints = results.len(),
+        balanced,
+        within_tolerance = within_tolerance.len(),
+        "Startup reconciliation passed"
+    );
+
+    Ok(())
+}
+
+#[cfg(test)]
```
amilz: want to add `Pubkey::from_str` -> `ReconciliationError::InvalidPubkey` tets?

indexer/src/indexer/reconciliation.rs
```diff
@@ -0,0 +1,757 @@
...
+    }
+
+    fn account_not_found_error() -> (i32, &'static str) {
+        (-32602, "Invalid param: could not find account")
+    }
+
+    fn make_mint_balance(
+        mint_address: &str,
+        total_deposits: i64,
+        total_withdrawals: i64,
```
amilz: from what i can tell, all tests run `total_withdrawals = 0` so i don't think `total_deposits - total_withdrawals` or the `SQL WHERE status = 'completed'` get tested

amilz (approved):

PR #15, opened 2026-02-24
feat(core): persist dedup state across restarts [PRO-906]

dev-jodee:
## Summary

Fixes invariant C3: MUST NOT allow two transactions with the same signature to both execute.

The dedup cache was entirely in-memory. On restart it was empty, creating two bugs:
- **Safety**: a successfully executed transaction could be re-submitted within its expiration window and execute again
- **Liveness**: transactions in-flight at restart time were all rejected because their `recent_blockhash` was no longer in the empty `live_blockhashes` list

## Approach

`BlockInfo` already lives in the `blocks` table for both Postgres and Redis. Adding `transaction_recent_blockhashes: Vec<Hash>` (with `#[serde(default)]` for backward compat) means we persist everything needed for recovery without a schema migration and without breaking either backend.

On startup, `load_dedup_state()` loads the last `max_blockhashes` blocks in a single query and rebuilds both `live_blockhashes` and `dedup_cache`. Falls back to empty state on a fresh node or DB error so startup is never blocked.

## Changes

**`core/src/accounts/traits.rs`** — Add `transaction_recent_blockhashes` to `BlockInfo` (`#[serde(default)]`, backward-compatible)

**`core/src/stages/settle.rs`** — Collect each transaction's `recent_blockhash` in all three match arms (Executed, FeesOnly, Err) and store in `BlockInfo`

**`core/src/accounts/get_blocks_in_range.rs`** — New batch loader: single `WHERE slot BETWEEN` for Postgres, single `MGET` for Redis (avoids ~300 sequential round-trips)

**`core/src/stages/dedup.rs`** — Add `initial_live_blockhashes` / `initial_dedup_cache` to `DedupArgs`; add `load_dedup_state()`

**`core/src/nodes/node.rs`** — Load dedup state via short-lived read-only DB connection before `start_dedup()`; add `Clone` to `NodeConfig`

**`integration/tests/contra/rpc/test_dedup_persistence.rs`** — New self-contained integration test: start node → mint + transfer → restart → replay identical transaction → assert balances and tx count unchanged

**`integration/tests/contra/rpc/utils.rs`** — Add `restart_contra()` helper

**`integration/tests/contra/rpc/test_context.rs`** — Fix `check_transaction_exists` retry budget: 3×100ms → 20×100ms (previous budget was shorter than worst-case settle + DB commit cycle, causing intermittent failures)

## Linear

Closes [PRO-906](https://linear.app/solana-fndn/issue/PRO-906/persist-dedup-state-across-restarts)

### DIFF
--- a/core/src/accounts/get_latest_slot.rs
+++ b/core/src/accounts/get_latest_slot.rs
@@ -1,6 +1,6 @@
 use {
     super::{postgres::PostgresAccountsDB, redis::RedisAccountsDB, traits::AccountsDB},
-    anyhow::{anyhow, Context, Result},
+    anyhow::{anyhow, Result},
     redis::AsyncCommands,
 };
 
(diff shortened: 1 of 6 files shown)


amilz (approved):
couple of comments. lmk

PR #22, opened 2026-02-25
test(core): add pipeline invariant unit tests (C3-C9) [PRO-914]

amilz:
## Summary

- Adds 20 unit tests directly verifying core pipeline invariants C3-C9
- C3/C4: dedup stage correctly rejects duplicate signatures and unknown/expired blockhashes, including window eviction boundary conditions
- C5: sigverify rejects transactions signed by the wrong keypair or with a tampered signature
- C6: `send_transaction_impl` rejects transactions invoking disallowed program IDs before they enter the pipeline
- C7/C8/C9: sigverify rejects admin instructions without admin signer, mixed admin+normal transactions, and empty transactions
- Each invariant is tested at its actual enforcement point in the pipeline

## Test Plan

```
cargo test -p contra-core
# 65 passed, 0 failed
```

Closes PRO-914

### DIFF
--- a/core/src/rpc/send_transaction_impl.rs
+++ b/core/src/rpc/send_transaction_impl.rs
@@ -138,8 +138,6 @@ mod tests {
         (WriteDeps { dedup_tx }, rx)
     }
 
-    // --- C6: disallowed program ID must be rejected --------------------
-
     #[tokio::test]
     async fn disallowed_program_rejected() {
         let payer = Keypair::new();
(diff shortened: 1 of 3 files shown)


amilz:
Also moved the inline `PACKET_DATA_SIZE` constant from `send_transaction_impl.rs` to `rpc/constants.rs` — same module where `MAX_BODY_SIZE`, `MAX_SLOT_RANGE`, etc. already live. Minor cleanup, no behavior change.

dev-jodee (approved):

PR #51, opened 2026-03-02
feat(core): add admin CLI with slot truncation and integration tests [PRO-913]

dev-jodee:
## Summary
- Add `admin` binary with `truncate` subcommand for pruning old block, transaction, and account_history data from Postgres
- Add `first_available_block` metadata lookup in `get_first_available_block` so truncated databases report the correct starting slot
- Add `truncate` module in `core/src/accounts/` with batch-based deletion, backup validation, dry-run mode, and VACUUM support
- Add end-to-end integration tests (`test_truncate_apply_mode_e2e`, `test_truncate_dry_run_e2e`) using Postgres testcontainers
- Copy `admin` binary into the Docker image

## Fixes applied during implementation
- Replace `.context()` with `.map_err(|e| anyhow!(...))` on `PostgresAccountsDB::new()` — `Box<dyn std::error::Error>` doesn't implement `std::error::Error` itself, so `anyhow::Context` can't be called on it
- Use `fetch_optional` + `.flatten()` instead of `fetch_one` when querying `first_available_block` metadata in dry-run test — no row exists in dry-run mode
- Convert manual async fn (`-> impl Future`) to `async fn` to satisfy clippy `manual_async_fn` lint

## Test plan
- [x] `cargo test -p contra-integration --test truncate_integration -- --nocapture` — both tests pass
- [x] `cargo clippy` passes on `core` and `integration` (truncate test) with `-D warnings`
- [ ] CI pipeline on `hardening/pre-release-audit`

### DIFF
--- a/core/src/accounts/truncate.rs
+++ b/core/src/accounts/truncate.rs
@@ -12,6 +12,8 @@ use {
 
 const FIRST_AVAILABLE_BLOCK_KEY: &str = "first_available_block";
 const ACCOUNT_HISTORY_TABLE: &str = "account_history";
+const TRUNCATE_ADVISORY_LOCK_ID: i64 = 0x434F4E_54525543; // "CONTRUC" as hex
+const MAX_BIND_PARAMS: usize = 60_000;
 
 #[derive(Debug, Clone)]
 pub struct TruncateOptions {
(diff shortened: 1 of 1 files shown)


core/src/accounts/truncate.rs
```diff
@@ -0,0 +1,528 @@
...
+        Self {
+            wal_archive_ok: false,
+            wal_archive_reason: "Skipped: no rows eligible for truncation".to_string(),
+            pg_dump_ok: false,
+            pg_dump_reason: "Skipped: no rows eligible for truncation".to_string(),
+        }
+    }
+}
+
+pub async fn truncate_slots(
```
amilz: I think we might want to use `pg_try_advisory_lock` to prevent 2 admins from running at the same time?

core/src/accounts/truncate.rs
```diff
@@ -0,0 +1,528 @@
...
+    } else {
+        0
+    };
+    report.account_history_rows_deleted = account_history_rows_deleted;
+
+    report.first_available_block = set_first_available_block_metadata(
+        pool.as_ref(),
+        query_first_available_slot(pool.as_ref()).await?,
+    )
+    .await?;
```
amilz: probably fine, but i think might make sense to wrap these 3 in a single operation to ensure we don't have only 1 state update if  there's a crash between operations

dev-jodee: Yeah was by design that it's fine if it crashes in between because you can just rerun the same command safely, and running 3 huge commands might break more often than running them 1 by 1 worst case

core/src/accounts/truncate.rs
dev-jodee: Added — `truncate_slots` now acquires a `pg_try_advisory_lock` at entry and releases it on exit. If the lock is already held, it returns an error immediately.

amilz (approved):

PR #53, opened 2026-03-03
tests: add coverage tests

amilz:
## Summary

- Add test coverage across `core` and `indexer` crates, targeting the highest-impact 0% files first
- Prune ~39 low-value tests (string formatting, Default impls, tautological mock assertions) so coverage numbers reflect real safety signal
- Bring `core` line coverage from **30% to 68%** and `indexer` from **~50% to 68.7%** (honest, no padding)
- Add `llvm-cov --ignore-filename-regex` excludes for files that skew coverage without test value (error boilerplate, constants, thin RPC wrappers, server wiring, Redis backend)
- Add wiremock-based webhook retry tests and testcontainers-backed DB integration tests for both crates
- Harden `MockStorage` so it records inputs and respects `should_fail` instead of returning hardcoded values

### What changed

**Core — DB integration tests** (`tests/db_tests.rs`)
- Account CRUD, block store/get/range, transaction write_batch roundtrip, blockhash/slot metadata, performance samples, read-only noop, truncation (validation, empty DB, dry run, actual deletion with tx cleanup, backup verification failure)

**Core — unit tests**
- `webhook.rs`: wiremock tests covering post_json retry loop (success, retry-then-succeed, max-attempts exhaustion, empty body, connection refused) plus backoff math and clamping
- `vm/admin.rs`: malformed SPL input tests (empty data, short data guard, unsupported instruction, valid InitializeMint end-to-end)
- `stages/settle.rs`: FeesOnly branch, executed transaction with account settlement, slot incrementing, writable-vs-readonly filtering, failed tx signature recording
- `rpc/mod.rs`: ~30 tests covering 15+ RPC impl files via testcontainers
- `accounts/bob.rs`: GC/eviction state machine (12 tests covering sync, tombstone, ahead-state, age-based eviction, concurrent preload+settle races)
- `stages/dedup.rs`: block window filtering, out-of-window rejection, mismatched lengths, multi-block live list
- `stages/sequencer.rs`: single tx batching, empty input, channel closure
- Plus: gasless_callback, gasless_rent_collector, transactions

**Indexer — Postgres integration tests** (`postgres_db_test.rs`)
- Schema lifecycle, single/batch inserts, pending/lock queries, status updates, checkpoints, mint upserts, reconciliation balances, withdrawal nonces

**Indexer — unit tests**
- `smt_util.rs`: exclusion/inclusion proofs, idempotent insert, order independence, reset, removal, leaf collision
- `rpc_util.rs`: retry policy (none vs idempotent), backoff, attempt exhaustion
- `instruction_util.rs`: MintToBuilder instruction generation, missing field errors, memo inclusion
- `mint_util.rs`: cache miss/hit, Token-2022 detection, RPC fallback, invalid owner, prefetch
- `config.rs`: validation for escrow/withdraw/RPC/Yellowstone configs
- `transaction_util.rs`: error mapping (MintNotInitialized, InvalidSmtProof, InvalidNonce)
- `rpc.rs`: RPC batch test flake fix with body-matched mocks + `.expect(1)` assertions

**MockStorage hardening**
- `insert_db_transaction`: records input, increments IDs, respects `should_fail`
- `update_transaction_status`: records all 4 params, respects `should_fail`
- `get_pending_db_transactions`: filters by `transaction_type`, respects `limit`
- `get_and_lock_pending_transactions`: filters by type, respects limit, only drains matched items

**Noise test cleanup**
- Deleted tests that only verified string formatting, Default impls, builder getter chains, struct field assignment, enum variant distinctness, and tautological mock returns
- Added `COV_IGNORE` to both Makefiles excluding error enums, constants, CLI entrypoints, thin RPC wrappers, and infra files from coverage reports

**Test helper consolidation**
- Moved `create_test_bob` into `test_helpers.rs` with a `BOB::new_test` constructor, eliminating duplicate dummy-pool setup across modules

## Test plan

- [x] `cargo test -p contra-core --lib` (167 pass)
- [x] `cargo test -p contra-core --test db_tests` (33 pass, needs Docker)
- [x] `cd indexer && cargo test --test postgres_db_test -- --test-threads=1` (24 pass)
- [x] `cargo test -p contra-indexer --lib --features datasource-rpc` (261 pass)
- [x] `cargo test -p contra-indexer --lib --features datasource-yellowstone` (261 pass)
- [x] `make fmt` clean

Closes DEV-210

### DIFF
--- a/indexer/src/operator/sender/proof.rs
+++ b/indexer/src/operator/sender/proof.rs
@@ -212,6 +212,7 @@ mod tests {
             retry_max_attempts: 3,
             rotation_retry_queue: Vec::new(),
             pending_rotation: None,
+            program_type: crate::config::ProgramType::Escrow,
         }
     }
 


core/src/accounts/bob.rs
```diff
@@ -324,6 +324,34 @@ impl TransactionProcessingCallback for BOB {
     }
 }
 
+#[cfg(test)]
```
dev-jodee: These things should not be in their own cfg test tag, if it's utils that are reused we should put in test util, or if it's part of the test for that file then it should be in the same test mod below

PR #74, opened 2026-03-12
fix(rpc): query accounts database directly in getAccountInfo

0xKyungmin:
## Summary

`getAccountInfo` always returned `null` for user accounts because it constructed a fresh `BOB` instance with an empty in-memory cache instead of querying the accounts database directly.

- `BOB` is the write-node's execution-time account cache. A newly created `BOB` only contains precompile accounts (system program, SPL Token, etc.), never user data.
- `getTokenAccountBalance` already uses the correct pattern: `read_deps.accounts_db.get_account_shared_data()`.
- This fix aligns `getAccountInfo` with that same pattern.

**Root cause:** `BOB::get_account_shared_data()` (via `TransactionProcessingCallback` trait) checks only its in-memory HashMap — it has no database fallback. When called from a fresh instance on the read-node, it always returns `None` for any non-precompile account.

## Changes

- Replace `BOB` instantiation with direct `AccountsDB` query (1 line)
- Remove unused imports (`BOB`, `TransactionProcessingCallback`, `mpsc`, `tracing::info`)
- Remove verbose per-request account data logging

## Test plan

- [x] Verified `getAccountInfo` returns account data (base64 + jsonParsed encoding) for both mint and token accounts on devnet
- [x] Verified `getTokenAccountBalance` still works correctly
- [x] Verified non-existent accounts still return `null`

### DIFF
--- a/core/src/rpc/get_account_info_impl.rs
+++ b/core/src/rpc/get_account_info_impl.rs
@@ -1,9 +1,6 @@
-use crate::{
-    accounts::bob::BOB,
-    rpc::{
-        error::{custom_error, INVALID_PARAMS_CODE, JSON_RPC_SERVER_ERROR},
-        ReadDeps,
-    },
+use crate::rpc::{
+    error::{custom_error, INVALID_PARAMS_CODE, JSON_RPC_SERVER_ERROR},
+    ReadDeps,
 };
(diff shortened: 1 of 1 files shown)


Huzaifa696:

PR #76, opened 2026-03-12
fix(rpc): return parsed JSON for token accounts in getAccountInfo jsonParsed encoding

0xKyungmin:
## Summary
- `getAccountInfo` and `simulateTransaction` with `encoding: "jsonParsed"` silently return base64 for SPL token accounts instead of parsed JSON
- Root cause: `encode_ui_account` receives `None` for `additional_data`, but the upstream parser (`parse_token_v3`) requires mint decimals to parse token accounts — on failure it falls back to base64
- Fix: when `jsonParsed` is requested and the account is owned by SPL Token, fetch the mint account to obtain decimals and pass them via `AccountAdditionalDataV3`

## Reproduction
```
curl -X POST http://localhost:8899 -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","id":1,"method":"getAccountInfo",
       "params":["<any-token-ATA>",{"encoding":"jsonParsed"}]}'
```

Before fix: `["aDOoHy...", "base64"]`
After fix: `{"parsed": {"info": {"mint": "...", "owner": "...", "tokenAmount": {...}}, "type": "account"}, "program": "spl-token"}`

Note: mint accounts already parse correctly (no decimals needed for that path).

## Scope
This fix covers classic SPL Token accounts only. Token-2022 is not registered in Contra's SVM (`processor.rs`) and `spl-token-2022` is not a dependency of `contra-core`, so Token-2022 accounts cannot exist on L2. Token-2022 support can be added when the program is registered in the SVM.

## Test plan
- [x] `cargo check -p contra-core` passes
- [x] `getAccountInfo` with `jsonParsed` on token account returns parsed JSON
- [x] `getAccountInfo` with `jsonParsed` on mint account still works
- [x] `getAccountInfo` with `base64` encoding unchanged
- [x] Non-token accounts (system) unaffected

### DIFF
--- a/core/src/rpc/get_account_info_impl.rs
+++ b/core/src/rpc/get_account_info_impl.rs
@@ -6,13 +6,18 @@ use crate::{
     },
 };
 use jsonrpsee::core::RpcResult;
-use solana_account_decoder::encode_ui_account;
+use solana_account_decoder::{
+    encode_ui_account,
+    parse_account_data::{AccountAdditionalDataV3, SplTokenAdditionalDataV2},
+};
 use solana_account_decoder_client_types::{UiAccount, UiAccountEncoding};
 use solana_client::{
     rpc_config::RpcAccountInfoConfig,
     rpc_response::{Response, RpcResponseContext},
 };
-use solana_sdk::pubkey::Pubkey;
+use solana_sdk::{account::AccountSharedData, account::ReadableAccount, pubkey::Pubkey};
+use spl_token::solana_program::program_pack::Pack;
+use spl_token::state::{Account as TokenAccount, Mint};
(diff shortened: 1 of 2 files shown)


Huzaifa696:

Huzaifa696:
Still reproduces on main. get_account_info_impl.rs:93 and simulate_transaction_impl.rs:109 both still pass None for additional_data, so jsonParsed silently falls back to base64 for token accounts. Please rebase onto main and I will review.

0xKyungmin:
@Huzaifa696 Rebased to main. Both paths are covered by tests that fail without the fix. Ready for review.

PR #77, opened 2026-03-12
fix(indexer): use StateWithExtensions for Token-2022 account unpacking in reconciliation

0xKyungmin:
## Summary
- `fetch_on_chain_balances` fetches token accounts from both `spl_token` and `spl_token_2022` programs, but uses `Pack::unpack` which requires exactly 165 bytes
- Token-2022 accounts include an `AccountType` discriminator (minimum 166 bytes), causing `InvalidAccountData` error that aborts the entire reconciliation run
- Use `StateWithExtensions::<Token2022Account>::unpack` for Token-2022 program accounts, which correctly handles the base state + extensions layout

## Changes
- Branch on `token_program_id` to select the appropriate unpacker
- Add `spl_token_2022::extension::StateWithExtensions` and `spl_token_2022::state::Account` imports (crate already in dependencies)
- Add 3 unit tests proving the bug and verifying the fix:
  - `test_spl_token_account_unpack` - SPL Token 165-byte unpack works
  - `test_spl_token_unpack_rejects_token2022_data` - Pack::unpack rejects 166-byte data
  - `test_state_with_extensions_unpacks_token2022_data` - StateWithExtensions handles it

## Test plan
- [x] 62 existing + 3 new unit tests pass
- [x] Docker build succeeds

### DIFF
--- a/indexer/src/operator/escrow_sweep.rs
+++ b/indexer/src/operator/escrow_sweep.rs
@@ -15,6 +15,8 @@ use solana_client::rpc_request::TokenAccountsFilter;
 use solana_sdk::pubkey::Pubkey;
 use spl_token::solana_program::program_pack::Pack;
 use spl_token::state::Account as TokenAccount;
+use spl_token_2022::extension::StateWithExtensions;
+use spl_token_2022::state::Account as Token2022Account;
 use std::collections::HashMap;
 use std::str::FromStr;
 use tracing::warn;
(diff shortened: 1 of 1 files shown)


Huzaifa696:

Huzaifa696:
Still reproduces on main. escrow_sweep.rs:139 iterates both token programs but line 175 still calls TokenAccount::unpack(&decoded), so any Token-2022 account with an extension aborts the reconciliation run. Please rebase onto main and I will review.

Huzaifa696:
Superseded by #323, which rebases this onto current `main` and carries your commit with authorship intact.

The rebase needed one adjustment beyond the conflict: `main`'s test module drifted after you opened this, so the new test no longer compiled (`EMPTY_BODY` is now `empty_body(slot)`, `result_body` takes a slot, the sweep returns `CustodySnapshot`, and there's a slot-agreement retry across the two token-program calls). The production change merged into `sweep_once` unmodified.

Thanks for the fix.

PR #95, opened 2026-04-09
RPC method getSignaturesForAddress implemented

sebatustra:
# feat(rpc): implement `getSignaturesForAddress`

Implements the `getSignaturesForAddress` Solana JSON-RPC method end-to-end.

## What changed

- **`address_signatures` table**: new table `(address, slot, signature)` indexed via composite PK. Written atomically with `transactions` so they're always consistent.
- **`get_signatures_for_address.rs`**: single LEFT JOIN query, `slot DESC` ordering, `before`/`until` cursor pagination for Postgres (Redis returns an error for cursor params — non-trivial with score-only sorted sets).
- **`get_signatures_for_address_impl.rs`**: address validation, limit clamped to `[1, 1000]`, typed `RpcSignaturesForAddressConfig`.
- **Gateway**: method added to `KNOWN_RPC_METHODS` and `ACCOUNT_GATED_METHODS` (JWT required, Phase 2 ownership check for user-role callers).

## Tests

| Location | Coverage |
|---|---|
| `core/src/accounts/traits.rs` | found, empty, same-slot ordering, before/until cursors, unknown cursor → error |
| `core/src/rpc/mod.rs` | found, empty, invalid address, limit clamping, newest-first ordering |
| `gateway/src/auth.rs` + `gateway/tests/auth_integration.rs` | no token → 401, operator proceeds, user role → ownership check, missing pubkey → 400 |

## Notes

- Unknown cursor signatures return an explicit error (Solana mainnet silently falls back — we consider that worse behavior).
- `mint.rs` idempotency check now actually works (previously hit `-32601 method not found`).
### Coverage Report

| Core | 9,043 | 10,447 | 86.6% | [rust-unit-coverage-reports](https://github.com/solana-foundation/contra/actions/runs/24755029365) |
| Indexer | 13,533 | 15,800 | 85.7% | [rust-unit-coverage-reports](https://github.com/solana-foundation/contra/actions/runs/24755029365) |
| Gateway | 991 | 1,115 | 88.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/contra/actions/runs/24755029365) |
| Auth | 541 | 596 | 90.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/contra/actions/runs/24755029365) |
| Withdraw Program | 118 | 230 | 51.3% | [unit-coverage-reports](https://github.com/solana-foundation/contra/actions/runs/24755029361) |
| Escrow Program | 1,170 | 1,951 | 60.0% | [unit-coverage-reports](https://github.com/solana-foundation/contra/actions/runs/24755029361) |
| E2E Integration | 8,219 | 12,228 | 67.2% | [e2e-coverage-reports](https://github.com/solana-foundation/contra/actions/runs/24755029365) |
| **Total** | **33,615** | **42,367** | **79.3%** | |

> Last updated: 2026-04-22 01:42:14 UTC by E2E Integration


core/src/accounts/traits.rs
```diff
@@ -428,6 +442,259 @@ mod tests {
...
+    #[tokio::test(flavor = "multi_thread")]
+    async fn get_signatures_for_address_found() {
+        use crate::test_helpers::create_test_sanitized_transaction;
+        use solana_sdk::signature::{Keypair, Signer};
+        use solana_svm::account_loader::LoadedTransaction;
+        use solana_svm::transaction_execution_result::{
+            ExecutedTransaction, TransactionExecutionDetails,
+        };
+        use solana_svm::transaction_processing_result::ProcessedTransaction;
+        use std::collections::HashMap;
```
Huzaifa696: nit: can we move the imports to the top of test mod

core/src/accounts/get_signatures_for_address.rs
```diff
@@ -0,0 +1,144 @@
...
+    Ok(results)
+}
+
+async fn get_signatures_for_address_redis(
+    _db: &RedisAccountsDB,
+    _address: &Pubkey,
+    _limit: usize,
+) -> Result<Vec<RpcConfirmedTransactionStatusWithSignature>> {
+    Ok(vec![])
+}
```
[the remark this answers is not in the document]
sebatustra: Resolved — the Redis backend is now fully implemented with addr_sigs:{pubkey} sorted set indexing in write_batch_redis and a  
  real query in get_signatures_for_address_redis; Ok(vec![]) is gone.

core/src/rpc/get_signatures_for_address_impl.rs
```diff
@@ -0,0 +1,42 @@
...
+    let pubkey = Pubkey::from_str(&address)
+        .map_err(|e| custom_error(INVALID_PARAMS_CODE, format!("Invalid address: {}", e)))?;
+
+    let limit = config
+        .as_ref()
+        .and_then(|c| c.get("limit"))
+        .and_then(|v| v.as_u64())
+        .map(|v| v as usize)
+        .unwrap_or(DEFAULT_LIMIT)
+        .clamp(1, MAX_LIMIT);
```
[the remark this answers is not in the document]
sebatustra: Resolved — config is now typed as RpcSignaturesForAddressConfig; before and until are parsed and forwarded to the DB layer where Postgres enforces them as pagination cursors.

core/src/accounts/get_signatures_for_address.rs
```diff
@@ -0,0 +1,144 @@
...
+    // address_signatures, and `data` only from transactions.
+    let rows = match sqlx::query(
+        "SELECT address_signatures.signature,
+                address_signatures.slot,
+                transactions.data
+         FROM address_signatures
+         LEFT JOIN transactions ON address_signatures.signature = transactions.signature
+         WHERE address_signatures.address = $1
+         ORDER BY address_signatures.slot DESC
+         LIMIT $2",
```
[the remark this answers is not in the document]
sebatustra: Resolved — ORDER BY slot DESC, signature DESC is now explicit, and a dedicated test verifies stable ordering for same-slot transactions.

core/src/accounts/traits.rs
sebatustra: Resolved

gateway/src/auth.rs
```diff
@@ -61,7 +61,11 @@ const OPERATOR_ONLY_METHODS: &[&str] = &["getBlock", "getTransaction", "simulate
 /// JSON-RPC spec:
 ///   getAccountInfo:         params: [pubkey, {encoding, ...}]
 ///   getTokenAccountBalance: params: [pubkey]
-const ACCOUNT_GATED_METHODS: &[&str] = &["getAccountInfo", "getTokenAccountBalance"];
+const ACCOUNT_GATED_METHODS: &[&str] = &[
+    "getAccountInfo",
+    "getTokenAccountBalance",
+    "getSignaturesForAddress",
```
dev-jodee, line 75: I think there's a corner case that's going to fail if we set on getSignaturesForAddress the current auth flow:

[`Here the check for account exists is before account owned by user`](https://github.com/solana-foundation/contra/blob/5a7a988f28fc2d9333f38edfd8cf1c633c98510f/gateway/src/lib.rs#L256)

Thinking out loud here, but if some accounts get closed, like closes TokenAccount, then that first if would fail, and user couldn't get signatures for an account he used to have ? Might not be an issue but curious to see what you think

core/src/accounts/get_signatures_for_address.rs
```diff
@@ -0,0 +1,292 @@
...
+        return Err(anyhow::anyhow!(
+            "before/until cursors are not yet supported by the Redis backend"
+        ));
+    }
+
+    let mut conn = db.connection.clone();
+    let key = format!("addr_sigs:{}", address);
+
+    // ZRANGE ... BYSCORE REV returns members with the highest score (most recent
+    // slot) first, matching the Postgres ORDER BY slot DESC behaviour.
```
dev-jodee, line 240: Is this going to have a different behavior than postgres since postgres has on slot and then on signature (https://github.com/solana-foundation/contra/pull/95/changes#diff-e1e9b8a701448e47decc131fc65967e2bc6d0b951829b2877bf49c2abc573736R138)

dev-jodee (commented):
two questions / discussions

gateway/src/auth.rs
sebatustra: Good catch, you're right. Once a `TokenAccount` is closed, `check_account_data_ownership` falls back to `is_wallet_owned_by_user` on the TA address directly, which always fails.                                                                                                 
                                                                                                                                        
I see three options:                                                                                                                        
                                                                                                                                        
1. Accept the limitation. No real incentive for users to close these accounts in our context (no rent to claim).                      
2. Optional mint param. Derive the ATA and check if it matches (partial solution, ATAs only).                                         
3. Snapshot ownership at index time. Store (address -> owner_wallet) in write_batch.rs during ingest. Complete solution but requires schema changes and extra work at ingest.

I'm leaning toward 1 for now, but curious about your thoughts?

core/src/accounts/get_signatures_for_address.rs
sebatustra: Thanks, fixed by storing members as hex instead of base58, which preserves byte ordering and matches Postgres's ORDER BY signature DESC

gateway/src/auth.rs
dev-jodee: Option #1 for now seems to make sense, if possible to add a comment mentioning that that is the case

sebatustra: Added in gateway/src/auth.rs at the ACCOUNT_GATED_METHODS const (commit b7280d7). Also merged main and adapted the address_signatures insert to the new bulk UNNEST pattern from #99

dev-jodee (approved):

PR #97, opened 2026-04-13
Fix/escrow support transfer fee [EXO-8]

sebatustra:
# fix(escrow): support Token2022 TransferFeeConfig extension

## Summary

Adds proper support for SPL Token 2022 mints with the `TransferFeeConfig` extension in both the `deposit` and `release_funds` instructions.

Previously, both processors used `Transfer` (the basic Token2022 CPI), which is rejected by the SPL Token 2022 program with error `Custom(31)` ("Mint required for this account to transfer tokens, use `transfer_checked`") whenever the mint has any extensions configured. Additionally, the processors assumed a 1:1 relationship between the requested amount and the tokens actually transferred — an assumption that breaks with transfer fee mints, where fees are withheld at the destination ATA.

### Program (`deposit.rs`, `release_funds.rs`)

- **`Transfer` → `TransferChecked`**: Both processors now use `TransferChecked`, which requires passing the `mint` account and `decimals`. This is required by SPL Token 2022 for any mint with extensions.

- **Balance-delta pattern**: Instead of asserting `balance_after == balance_before ± args.amount` (which fails when fees are withheld), both processors now compute the actual transferred amount from the observed balance change:
  - Deposit: `received = escrow_balance_after - escrow_balance_before`
  - Release: `released = escrow_balance_before - escrow_balance_after`

  The delta is used in the emitted event and replaces the strict equality check that was previously returning `InvalidEscrowBalance`.

### Tests (`test_deposit/mod.rs`, `test_release_funds/mod.rs`, `utils.rs`)

- Added `create_mint_2022_with_transfer_fee` helper in `utils.rs` that creates a Token2022 mint with `TransferFeeConfig` using real on-chain instructions (`create_account` → `initialize_transfer_fee_config` → `initialize_mint`). The order matters — the extension must be initialized before the base mint.

- Fixed `get_token_balance` in `utils.rs` to use `StateWithExtensions::<Token2022Account>::unpack` instead of `Token2022Account::unpack`. The plain unpack fails on accounts that have the `TransferFeeAmount` extension attached (which is added automatically to every ATA for a fee-bearing mint).

- Added `test_deposit_token_2022_transfer_fee_success` and `test_release_funds_token_2022_transfer_fee_success`.

**Note on test assertions**: These two tests do **not** use the shared `assert_get_or_deposit` / `assert_deposit_balances` helpers used by the rest of the suite. Those helpers assert that the escrow balance increases by exactly `deposit_amount`, which is incorrect for fee-bearing mints (escrow only receives `amount - fee`). The transfer fee tests use inline balance assertions with explicit fee calculations instead:

```rust
let expected_fee = (DEPOSIT_AMOUNT as u128 * TRANSFER_FEE_BASIS_POINTS as u128 / 10_000) as u64;
let expected_received = DEPOSIT_AMOUNT - expected_fee;
assert_eq!(instance_balance_after, instance_balance_before + expected_received);
```

The release funds test also raises the compute unit limit to `1_200_000` (vs the default `200_000`) because SMT proof verification is expensive enough to exceed the default budget.
### Coverage Report

| Core | - | - | - | - |
| Indexer | - | - | - | - |
| Gateway | - | - | - | - |
| Auth | - | - | - | - |
| Withdraw Program | 118 | 230 | 51.3% | [unit-coverage-reports](https://github.com/solana-foundation/contra/actions/runs/24371021559) |
| Escrow Program | 1,170 | 1,951 | 60.0% | [unit-coverage-reports](https://github.com/solana-foundation/contra/actions/runs/24371021559) |
| E2E Integration | 7,895 | 11,473 | 68.8% | [e2e-coverage-reports](https://github.com/solana-foundation/contra/actions/runs/24371021558) |
| **Total** | **9,183** | **13,654** | **67.3%** | |

> Last updated: 2026-04-13 23:13:31 UTC by E2E Integration

### DIFF
--- a/contra-escrow-program/program/src/processor/deposit.rs
+++ b/contra-escrow-program/program/src/processor/deposit.rs
@@ -4,7 +4,7 @@ use crate::{
     error::ContraEscrowProgramError,
     events::DepositEvent,
     processor::{
-        get_token_account_balance,
+        get_mint_decimals, get_token_account_balance,
         shared::{
             account_check::{verify_signer, verify_system_program},
             event_utils::emit_event,
@@ -18,7 +18,9 @@ use crate::{
 };
 use pinocchio::{account::AccountView, error::ProgramError, Address, ProgramResult};
 
-use pinocchio_token_2022::{instructions::Transfer as Transfer2022, ID as TOKEN_2022_PROGRAM_ID};
+use pinocchio_token_2022::{
+    instructions::TransferChecked as TransferChecked2022, ID as TOKEN_2022_PROGRAM_ID,
+};
 
 /// Processes the Deposit instruction.
 ///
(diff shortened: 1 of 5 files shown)


dev-jodee (approved):
lgtm

PR #102, opened 2026-04-22
fix(operator): prevent operator crash on malformed transaction

Huzaifa696:
## Summary

A single unprocessable withdrawal row previously tore down the whole operator processor task. The resulting cascade left the operator as a zombie container — process alive, pipeline dead, no `/healthz` to tell Docker otherwise.

This PR converts that crash path into a **halt-and-quarantine** recovery: the poison row is moved to `ManualReview`, the fetcher channel is drained, and all active withdrawals for that `program_type` are swept to `ManualReview` so nothing is submitted on-chain against a stale nonce. Deposits are unaffected.

## Changes

- **Processor** — per-transaction errors no longer tear down the task; offending rows are classified and quarantined in place, and panics are replaced with typed errors.
- **Operator supervisor** — critical tasks are watched via `select!`; an unexpected exit triggers a clean shutdown instead of leaving a zombie container.
- **Storage** — new cascade sweep that flips all active withdrawals to `ManualReview` in one DB transaction, with the poison row excluded so it is not re-alerted.
- **Metrics** — quarantine and task-exit counters so dashboards can page without tailing logs.

## Impact

- Operator no longer crash-loops or wedges on a malformed withdrawal row.
- No on-chain nonce gap introduced by the recovery path.

### Coverage Report

| Core | 9,043 | 10,447 | 86.6% | [rust-unit-coverage-reports](https://github.com/solana-foundation/contra/actions/runs/24783254831) |
| Indexer | 14,402 | 16,742 | 86.0% | [rust-unit-coverage-reports](https://github.com/solana-foundation/contra/actions/runs/24783254831) |
| Gateway | 991 | 1,115 | 88.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/contra/actions/runs/24783254831) |
| Auth | 541 | 596 | 90.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/contra/actions/runs/24783254831) |
| E2E Integration | 8,292 | 12,429 | 66.7% | [e2e-coverage-reports](https://github.com/solana-foundation/contra/actions/runs/24783254831) |
| **Total** | **33,269** | **41,329** | **80.5%** | |

> Last updated: 2026-04-22 14:31:11 UTC by E2E Integration

### DIFF
--- a/indexer/src/metrics.rs
+++ b/indexer/src/metrics.rs
@@ -134,6 +134,28 @@ gauge_vec!(
     &["program_type"]
 );
 
+// Poison-pill: a single transaction that could not be sent on-chain was
+// quarantined to ManualReview so the pipeline could keep moving.  The `reason`
+// label mirrors the error classification (`invalid_pubkey`, `invalid_builder`,
+// `missing_nonce`, ...) so dashboards can distinguish systemic bugs from
+// one-off bad rows.
+counter_vec!(
+    OPERATOR_TRANSACTION_QUARANTINED,
+    "contra_operator_transaction_quarantined_total",
+    "Transactions quarantined to ManualReview by the processor",
+    &["program_type", "reason"]
+);
+
+// Supervision: a critical task inside the operator exited.  The supervisor
+// aborts the process immediately when this increments; the counter exists
+// so dashboards can alert even if the restart is fast.
+counter_vec!(
+    OPERATOR_TASK_EXIT,
+    "contra_operator_task_exit_total",
+    "Critical operator task exits observed by the supervisor",
+    &["program_type", "task"]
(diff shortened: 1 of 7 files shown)


Huzaifa696:
**Regarding the Greptile comments:**

**P2 — Dead metric labels**: Dropped unused metrics.

**P2 — Double ManualReview webhook**: Skipped the poison row in the sweep.

**P2 — `critical_exit` false positive on ctrl-c**: Added `biased;` to the `select!`.

**P1 — Unscoped SQL sweep**: No `instance_pda` column today and the fetcher's scope is the same.

dev-jodee (approved):

PR #126, opened 2026-05-19
fix(operator): block deposits for non-allowlisted mints (SOLA2-20)

Huzaifa696:
## Summary

Addresses SOLA2-20 finding. The deposit operator would issue private-channel tokens for any mint present on a `transactions` row, without checking whether the mint had ever been authorized on-chain via `AllowMint`. A foreign mint slipping past the indexer's checks (or inserted manually) would get mirrored on the private channel with no Mainnet escrow backing, and balance reconciliation wouldn't catch it (the balance check only iterates `mints`).

This PR adds the in-line allowlist gate, plus a reconciliation orphan check that flags any deposit row that should never have been written. Both compose: the gate prevents the next bad mint, the orphan check detects anything that already slipped past or arrives via a future indexer bug.

## What's added

- **Deposit-side gate.** New `MintCache::assert_mint_allowlisted` is called from `process_deposit_funds` before any builder construction. Non-allowlisted rows quarantine to `ManualReview`; the deposit loop continues.
- **Reconciliation orphan check.** Each tick queries deposit row IDs whose mint has no `mints` entry and logs them with per-row dedup (baseline once, error on each new id, silent in steady state). Startup logs a full snapshot.
- **Runbook + drill.** `deposit_manual_review.md` adds Path E (indexer-gap backfill / investigate-then-delete / foreign-instance Tier 3). `drill_15` pins the runbook against source.

### Coverage Report

| Core | 9,620 | 11,100 | 86.7% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/26760804325) |
| Indexer | 16,440 | 19,181 | 85.7% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/26760804325) |
| Gateway | 994 | 1,164 | 85.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/26760804325) |
| Auth | 717 | 831 | 86.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/26760804325) |
| Withdraw Program | 120 | 232 | 51.7% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/26760804411) |
| Escrow Program | 1,185 | 1,963 | 60.4% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/26760804411) |
| DvP Swap Program | 191 | 1,044 | 18.3% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/26760804411) |
| E2E Integration | 10,225 | 12,681 | 80.6% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/26760804325) |
| **Total** | **39,301** | **47,152** | **83.3%** | |

> Last updated: 2026-06-01 14:53:22 UTC by E2E Integration


indexer/tests/runbook_drills.rs
```diff
@@ -1056,6 +1064,158 @@ async fn drill_14_deposit_manual_review_post_jit_recovery_flows(
...
+    // ── Step 3: cross-signal contract with reconciliation ────────────
+    // Path E's recovery for the terminal branch tells the on-call that
+    // the reconciliation orphan alert is the same incident — so the
+    // runbook must reference that alert, and reconciliation must
+    // actually run the orphan-mint check. Anchor each side concretely.
+    let reconciliation_path = workspace_root.join("indexer/src/operator/reconciliation.rs");
+    let reconciliation = std::fs::read_to_string(&reconciliation_path)
+        .unwrap_or_else(|e| panic!("read {reconciliation_path:?}: {e}"));
+    assert!(
+        reconciliation.contains("check_orphan_deposit_mints"),
```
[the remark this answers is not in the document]
Huzaifa696: fixed

indexer/src/operator/reconciliation.rs
```diff
@@ -185,9 +190,80 @@ async fn perform_reconciliation_check(
         info!("Balance reconciliation successful: all mints within tolerance");
     }
 
+    // Balance reconciliation iterates `mints`, so deposit rows whose mint
+    // has no `mints` entry never show up as a mismatch. Surface them
+    // explicitly here, with dedup so we don't re-log on every tick.
+    check_orphan_deposit_rows(storage, previously_alerted_orphans).await;
```
dev-jodee: if `send_webhook_alert` fails with ?, this will be skipped completely

indexer/src/operator/reconciliation.rs
```diff
@@ -185,9 +190,80 @@ async fn perform_reconciliation_check(
...
+                .collect();
+
+            if new_orphans.is_empty() {
+                // Steady state, silent to avoid log spam on every tick.
+                return;
+            }
+
+            error!(
+                new_row_count = new_orphans.len(),
+                "Reconciliation found {} new orphan deposit row(s) (deposit rows with no AllowMint record)",
```
dev-jodee: should this print the ids as well so the operator can debug?

dev-jodee (commented):
Two questions, and also just in general I think we should really limit the number of comments, every PR has had a lot of new comment lines and I think a lot of it will just be noise, the code itself should be readable and understandable enough that we don't need multi line comments that often, unless it's something very "monkey patched"

indexer/src/operator/reconciliation.rs
Huzaifa696: moved the orphan check to run first so a downstream ? can't skip it.

indexer/src/operator/reconciliation.rs
Huzaifa696: Added ids to the error log.

indexer/src/operator/processor.rs
```diff
@@ -2265,4 +2306,138 @@ mod tests {
...
+        assert!(
+            sender_rx.try_recv().is_err(),
+            "no Mint builder should be forwarded for an unknown mint",
+        );
+    }
+
+    /// Happy-path guard: a deposit whose mint IS registered in `mints`
+    /// produces a `Mint` builder with the expected recipient/amount.
+    /// Protects against a future regression that accidentally always
+    /// quarantines the deposit path.
```
gitteri: do we actually need this? are there tests above that cover this?

indexer/src/operator/reconciliation.rs
```diff
@@ -188,6 +196,73 @@ async fn perform_reconciliation_check(
...
+            // Baseline pass, informational log only.
+            if orphans.is_empty() {
+                info!("Reconciliation baseline: no orphan deposit rows detected");
+            } else {
+                info!(
+                    row_count = orphans.len(),
+                    "Reconciliation baseline: {} orphan deposit row(s) currently present \
+                     (no AllowMint record); subsequent ticks will only log on new entries",
+                    orphans.len()
+                );
```
gitteri: if orphans is not empty why aren't we logging an error?

docs/runbooks/deposit_manual_review.md
```diff
@@ -202,6 +211,111 @@ UPDATE transactions SET status = 'pending', updated_at = NOW()
  WHERE id = :transaction_id;
 ```
 
+### Path E - mint not in `AllowMint` allowlist
```
gitteri, line 214: how would an operator cancel the deposit and refund the user without minting on the private channel?

indexer/src/operator/processor.rs
Huzaifa696: You're right, once the older happy-path tests were updated with insert_mint_row(...), they cover the same ground. The only extra assertion here was "no quarantine update emitted," which is implicit in the existing tests. 
Removed.

indexer/src/operator/reconciliation.rs
Huzaifa696: info! was deliberate to avoid re-firing on restart, but since this is log-only the cost is one line, changed to error!.

docs/runbooks/deposit_manual_review.md
Huzaifa696: Verified on-chain: process_deposit requires the AllowedMint PDA, so a deposit for a non-allowlisted mint can't actually land, the escrow program rejects it, no funds move. 
The "User error / manual release_funds" bullet in 3b is therefore impossible; I have droped it.

indexer/src/operator/reconciliation.rs
```diff
@@ -188,6 +196,74 @@ async fn perform_reconciliation_check(
...
+                info!("Reconciliation baseline: no orphan deposit rows detected");
+            } else {
+                error!(
+                    row_count = orphans.len(),
+                    orphan_ids = ?orphans,
+                    "Reconciliation baseline: {} orphan deposit row(s) currently present \
+                     (no AllowMint record); subsequent ticks will only log on new entries",
+                    orphans.len()
+                );
+            }
```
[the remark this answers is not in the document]
Huzaifa696: fixed

indexer/src/operator/reconciliation.rs
gitteri: it's still info

indexer/src/operator/processor.rs
```diff
@@ -127,6 +127,7 @@ enum ErrorDisposition {
 fn classify_processor_error(err: &OperatorError) -> ErrorDisposition {
     match err {
         OperatorError::InvalidPubkey { .. } => ErrorDisposition::Quarantine("invalid_pubkey"),
+        OperatorError::MintNotAllowed { .. } => ErrorDisposition::Quarantine("mint_not_allowed"),
```
gitteri, line 130: missing `MintNotAllowed` test in classify_processor_error_covers_every_variant

indexer/src/operator/reconciliation.rs
```diff
@@ -188,6 +196,73 @@ async fn perform_reconciliation_check(
...
+                .into_iter()
+                .filter(|id| !seen.contains(id))
+                .collect();
+
+            if new_orphans.is_empty() {
+                // Steady state, silent to avoid log spam on every tick.
+                return;
+            }
+
+            error!(
```
sebatustra, line 269: does an orphan row means the indexer wrote a deposit for an unauthorized mint? in that case maybe it's worthy of a webhook altert? (like in balance mismatch in `perform_reconciliation_check`)

indexer/src/operator/reconciliation.rs
Huzaifa696: Changed info! to error! in [86d4fad](https://github.com/solana-foundation/solana-private-channels/pull/126/commits/86d4fadc5b0db012e78b68b8f6ba55399829eb8a).

indexer/src/operator/processor.rs
Huzaifa696: added test in https://github.com/solana-foundation/solana-private-channels/pull/126/commits/3b1e32e32c7d894d1a1f7b3a5e976288db1e630d

indexer/src/operator/reconciliation.rs
Huzaifa696: Yes, an orphan row means the indexer persisted a deposit for a mint with no AllowMint record.
Wired a webhook alert.

Huzaifa696:
To make the orphan reconciliation check future-proof for upcoming BlockMint support and scenarios where a mint may transition between allowed and blocked states, I made additional changes to incorporate mint history when determining whether a deposit should be classified as an orphan.

docs/runbooks/deposit_manual_review.md
```diff
@@ -202,6 +211,109 @@ UPDATE transactions SET status = 'pending', updated_at = NOW()
...
+
+```sql
+INSERT INTO mints
+  (mint_address, decimals, token_program, is_pausable, has_permanent_delegate, created_at)
+VALUES
+  (:mint, :decimals, :token_program, :is_pausable, :has_permanent_delegate, NOW());
+
+UPDATE transactions SET status = 'pending', updated_at = NOW()
+ WHERE id = :transaction_id;
+```
```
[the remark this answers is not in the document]
Huzaifa696: Done: Step 3a now inserts the mint_status_history row too

Huzaifa696:
Scenarios where allow-mint and block-mint operations occur within the same slot are out of scope for this PR. If needed, we can track support for those cases in a separate ticket.

docs/runbooks/deposit_manual_review.md
```diff
@@ -202,6 +211,117 @@ UPDATE transactions SET status = 'pending', updated_at = NOW()
...
+incident record and delete it. The
+reconciliation orphan check has no status filter and runs with in-memory
+per-id dedup, so a `failed` row stays silent in steady state but
+re-appears in the orphan log — and re-posts a webhook alert via
+`reconciliation_webhook_url` (payload: `orphan_ids`, `row_count`,
+`timestamp`) — on every operator restart. Deleting the row is the only
+way to remove that recurring boot-time noise:
+
+```sql
+DELETE FROM transactions WHERE id = :transaction_id;
```
sebatustra, line 313: this DELETE is status agnostic, is that intentional? is there a scenario where a `completed` transaction is deleted?

Huzaifa696: 3b is a recovery process for which we have already identified the valid orphan transaction_id in step 1 and 2, which is unique. So, its safe.

sebatustra (approved):
lgtm

PR #127, opened 2026-05-19
feat(operator): recover stuck Processing transactions after operator crash

Huzaifa696:
## Summary

Addresses SOLA2-12 finding. When the fetcher flipped a row from `pending` to `processing` and the operator died before the row reached a terminal status (`completed` / `failed` / `manual_review` / `pending_remint`), the row sat in `processing` forever. The fetcher's next loop only selects `pending` rows. For deposits this meant tokens locked on Mainnet with the private-channel mint never happening; for withdrawals, the user's burned private-channel tokens with no matching escrow release.

This PR adds a peer recovery worker that runs on boot and on a 60 s interval, picks up rows that have sat in `processing` past a 5 min stale threshold, asks the chain whether the operator's intended action already happened, and updates each row to its true state.

## What's added

- **Recovery worker.** New `operator/recovery.rs`, spawned as a peer task next to fetcher/processor/sender. Per type: deposits go through an RPC idempotency check (three-way `Landed` / `NotLanded` / `Ambiguous`); withdrawals demote unconditionally and rely on the escrow program's exclusion-proof check to prevent double-release.
- **Storage helpers.** `get_stale_processing_transactions` plus three conditional writes (`try_complete_processing`, `try_requeue_processing`, `try_quarantine_processing`) gated on `updated_at = $expected`. Returns `Ok(false)` instead of erroring when another writer/operator moved the row first.
- **Tightened terminal write.** `update_transaction_status_internal` now carries `AND status = 'processing'` so a lagging in-flight completion cannot stomp a recovery demote. Mirrors `set_pending_remint_internal`.
- **Webhook integration.** Quarantine outcomes emit a `TransactionStatusUpdate` through the shared `storage_tx` channel so the existing `DbTransactionWriter` fires the alert webhook with a runbook-keyed reason string.
- **Metric.** New `OPERATOR_STALE_PROCESSING_RECOVERED{program,outcome,type}` counter. 
- **Runbooks + drills.** Adds Path E to `deposit_manual_review.md` (RPC unreachable) and Path F to `withdrawal_manual_review.md` (corrupt withdrawal row). 
- **Tests.** One new integration suite covering 12 IT scenarios + 8 unit tests inside `recovery.rs`. IT-5 and IT-12 also assert the metric increments.

### Coverage Report

| Core | 9,735 | 11,231 | 86.7% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27134188494) |
| Indexer | 18,753 | 21,734 | 86.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27134188494) |
| Gateway | 994 | 1,164 | 85.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27134188494) |
| Auth | 717 | 831 | 86.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27134188494) |
| E2E Integration | 10,989 | 13,564 | 81.0% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27134188494) |
| **Total** | **41,188** | **48,524** | **84.9%** | |

> Last updated: 2026-06-08 11:50:44 UTC by E2E Integration

### DIFF
--- a/docs/runbooks/deposit_manual_review.md
+++ b/docs/runbooks/deposit_manual_review.md
@@ -52,6 +52,7 @@ to the right Path below.
 | `invalid_builder` | Builder rejected the row's data (e.g. negative amount). |
 | `program_error` | Generic builder error not covered by the specific variants. |
 | `deposit idempotency:` | The recovery worker could not authoritatively decide whether the deposit's mint already landed. Row was quarantined on uncertainty rather than risk a double-mint. Indicates RPC health, not a row defect. See **Path E**. |
+| `recovery requeues without progress` | The recovery worker demoted this `processing` deposit 3 times (mint never landed each cycle) and quarantined rather than loop forever. Row data is fine; the mint keeps failing to land. See **Path G**. |
 
 ### Sender-side post-JIT surface (`sender/mint.rs`)
 
@@ -117,6 +118,7 @@ Correct the columns and re-arm:
 ```sql
 UPDATE transactions
    SET status = 'pending',
+       recovery_requeue_attempts = 0,
        mint = :corrected_mint,
        recipient = :corrected_recipient,
        updated_at = NOW()
@@ -142,7 +144,7 @@ extended to classify this error variant explicitly. Do not patch
 in-place.
 
 ```sql
-UPDATE transactions SET status = 'pending', updated_at = NOW()
+UPDATE transactions SET status = 'pending', recovery_requeue_attempts = 0, updated_at = NOW()
  WHERE id = :transaction_id;
 ```
 
@@ -213,16 +215,19 @@ Only for the recoverable authority case after the underlying authority
 correction has been performed:
 
 ```sql
-UPDATE transactions SET status = 'pending', updated_at = NOW()
+UPDATE transactions SET status = 'pending', recovery_requeue_attempts = 0, updated_at = NOW()
  WHERE id = :transaction_id;
 ```
 
 ### Path E - recovery-worker idempotency lookup failed
 
 `error_message` starts with `deposit idempotency:`. The recovery worker
-intentionally refuses to demote on RPC failure (the live mint path
-fails-open on the same condition; recovery cannot, because recovery has
-no second chance to catch a duplicate).
+quarantines rather than demote when the idempotency lookup fails on RPC
+error, because demoting would re-pick the row and risk a double-mint. This
+is the same fail-closed stance as the live mint path, which on the same RPC
+failure also stops without minting (it routes to a terminal failure via
+`send_fatal_error` rather than minting unverified). Both paths refuse to
+proceed on uncertainty; they differ only in terminal state.
 
 #### Step 1 - check RPC health
 
@@ -258,7 +263,7 @@ The mint did not land. Re-arm to `pending`; the idempotency memo
 prevents duplicate mint on the next attempt.
 
 ```sql
-UPDATE transactions SET status = 'pending', updated_at = NOW()
+UPDATE transactions SET status = 'pending', recovery_requeue_attempts = 0, updated_at = NOW()
  WHERE id = :transaction_id;
 ```
 
@@ -331,7 +336,7 @@ VALUES
   (:mint, 'allowed', :allow_mint_slot, :allow_mint_signature, NOW())
 ON CONFLICT (mint_address, effective_slot) DO NOTHING;
 
-UPDATE transactions SET status = 'pending', updated_at = NOW()
+UPDATE transactions SET status = 'pending', recovery_requeue_attempts = 0, updated_at = NOW()
  WHERE id = :transaction_id;
 ```
 
@@ -376,6 +381,50 @@ The `AllowMint` exists but binds to a different escrow instance than
 ours. The operator should never see foreign-instance rows — this is the
 SOLA2-27 attack surface, not an operations recovery.
 [Escalate](_escalation.md) (Tier 3). No SQL.
+
+### Path G - requeue cap exhausted (mint never landed)
+
+`error_message` contains `recovery requeues without progress`. Each
+recovery pass whose idempotency scan finds no existing mint (`NotLanded`)
+demotes the stuck `processing` deposit back to `pending` for a fresh
+mint. After `MAX_RECOVERY_REQUEUE_ATTEMPTS` (3) such requeues with the
+mint still never landing, recovery quarantines instead of looping. So the
+mint was retried 3 times and never landed. The lookup *succeeded* each
+time and said not-landed (distinct from Path E, where the lookup itself
+*failed* on RPC error); the row data is valid (distinct from Path A).
+
+#### Step 1 - verify on-chain
+
+Run [`_verify_onchain_mint.md`](_verify_onchain_mint.md). Expected
+verdict: `NOT_LANDED`. If `LANDED` -> the idempotency scan missed a landed
+mint; mark `completed` per **Path E** (`LANDED` branch) and
+[escalate](_escalation.md) (Tier 2) - the scan has a defect (e.g. the sig
+is past its lookback window).
+
+#### Step 2 - find why the mint never lands
+
+The deposit mint is fire-and-forget (`spawn_fire_and_store`), so a
+repeating failure points at a deterministic mint rejection (paused/frozen
+mint, `mint_authority` mismatch, compute) or the sender never
+broadcasting. Read the operator logs for this `transaction_id`'s mint send
+error. A deterministic cause means re-arming will loop again ->
+[escalate](_escalation.md) (Tier 2) to engineering.
+
+#### Step 3 - resolve, then re-arm
+
+Fix the root cause first. Then re-arm to `pending` **and reset the requeue
+counter** - re-arming without the reset re-quarantines on the next stall,
+since the counter is already at the cap. The idempotency memo still guards
+against a double-mint on retry:
+
+```sql
+UPDATE transactions
+   SET status = 'pending', recovery_requeue_attempts = 0, updated_at = NOW()
+ WHERE id = :transaction_id;
+```
+
+If the mint can never land (unrecoverable), mark `failed` and
+[escalate](_escalation.md) (Tier 1) for refund coordination.
 
 ## Post-incident artifacts
 
--- a/docs/runbooks/deposit_manual_review.md
+++ b/docs/runbooks/deposit_manual_review.md
@@ -16,7 +16,7 @@ Practically: a single deposit `manual_review` is a single row in trouble.
 Other deposits keep flowing. There is no collateral, no sweep, no halt to
 recover from.
 
-There are **two trigger surfaces** that can land a deposit in
+There are **three trigger surfaces** that can land a deposit in
 `manual_review`:
 
 1. **Processor-side row-data validation** — deterministic per-row error
@@ -25,6 +25,11 @@ There are **two trigger surfaces** that can land a deposit in
    mint-init helper found the on-chain mint in a state it cannot fix by
    re-issuing `mint_to` (wrong authority on the private channel mint, or
    corrupt mint data). See **Path D** for recovery.
+3. **Recovery-worker idempotency lookup failed** — the periodic
+   stuck-row recovery worker found a stale `Processing` deposit, tried
+   the idempotency memo lookup against the PrivateChannel RPC, and the
+   RPC was unreachable. Row data is fine; no on-chain mint attempted.
+   See **Path E**.
 
 ## Symptom
 
@@ -33,8 +38,8 @@ There are **two trigger surfaces** that can land a deposit in
 
 ## Triage
 
-`error_message` distinguishes the two trigger surfaces and dispatches to
-the right Path below.
+`error_message` distinguishes the three trigger surfaces and dispatches
+to the right Path below.
 
 ### Processor-side surface (`processor.rs::process_deposit_funds`)
 
@@ -43,6 +48,7 @@ the right Path below.
 | `invalid_pubkey` | `mint` or `recipient` field is not a valid base58 pubkey. |
 | `invalid_builder` | Builder rejected the row's data (e.g. negative amount). |
 | `program_error` | Generic builder error not covered by the specific variants. |
+| `deposit idempotency:` | The recovery worker could not authoritatively decide whether the deposit's mint already landed. Row was quarantined on uncertainty rather than risk a double-mint. Indicates RPC health, not a row defect. See **Path E**. |
 
 ### Sender-side post-JIT surface (`sender/mint.rs`)
 
@@ -202,6 +208,55 @@ UPDATE transactions SET status = 'pending', updated_at = NOW()
  WHERE id = :transaction_id;
 ```
 
+### Path E - recovery-worker idempotency lookup failed
+
+`error_message` starts with `deposit idempotency:`. The recovery worker
+intentionally refuses to demote on RPC failure (the live mint path
+fails-open on the same condition; recovery cannot, because recovery has
+no second chance to catch a duplicate).
+
+#### Step 1 - check RPC health
+
+Inspect the suffix of `error_message` to identify the underlying RPC
+transport failure (timeout, connection refused, HTTP 5xx, etc.).
+Confirm against the operator's current PrivateChannel RPC endpoint:
+
+```bash
+curl -s -X POST -H "Content-Type: application/json" \
+  -d '{"jsonrpc":"2.0","id":1,"method":"getHealth"}' \
+  <private-channel-rpc-url>
+```
+
+If RPC is unhealthy, address that first. If RPC is healthy now,
+proceed.
+
+#### Step 2 - verify on-chain
+
+Run [`_verify_onchain_mint.md`](_verify_onchain_mint.md). The row was
+quarantined because the recovery worker could not answer this question
+itself; resolving the runbook requires the same answer.
+
+#### Step 3 - branch on verdict
+
+##### `LANDED <signature>` — mint already succeeded
+
+Mark `completed` with the observed signature (same SQL as
+[`deposit_failed.md`](deposit_failed.md) Path B).
+
+##### `NOT_LANDED` — re-arm
+
+The mint did not land. Re-arm to `pending`; the idempotency memo
+prevents duplicate mint on the next attempt.
+
+```sql
+UPDATE transactions SET status = 'pending', updated_at = NOW()
+ WHERE id = :transaction_id;
+```
+
+##### `AMBIGUOUS`
+
+Stop. [Escalate](_escalation.md) (Tier 2). Do not act.
+
 ## Post-incident artifacts
 
 - Transaction id, originating Solana `signature`, `recipient`, `mint`.
@@ -211,4 +266,6 @@ UPDATE transactions SET status = 'pending', updated_at = NOW()
 - For Path D: on-chain `mint_authority` before/after, the
   `spl-token authorize` signature (if applicable), and the engineering
   ticket for any mint-replacement coordination.
+- For Path E: RPC endpoint health snapshot and the
+  `getHealth` / verification responses captured at triage time.
 
--- a/indexer/src/operator/operator.rs
+++ b/indexer/src/operator/operator.rs
@@ -232,8 +232,10 @@ pub async fn run(
     // A task exit increments the OPERATOR_TASK_EXIT metric with a task
     // label so dashboards can tell which one failed without tailing logs.
     //
-    // Non-critical tasks (reconciliation, feepayer monitor) are not watched
-    // here.
+    // The recovery worker is critical: if it dies, stuck-Processing rows stop
+    // being recovered, so an unexpected exit must page and restart like the
+    // pipeline stages. Non-critical tasks (reconciliation, feepayer monitor)
+    // are not watched here.
     //
     // Handles are polled by mutable reference so ownership stays here and
     // they can still be moved into `shutdown_operator` below — awaiting an
@@ -242,6 +244,7 @@ pub async fn run(
     let mut processor_handle = processor_handle;
     let mut sender_handle = sender_handle;
     let mut storage_writer_handle = storage_writer_handle;
+    let mut recovery_handle = recovery_handle;
     let pt_label = program_type.as_label();
 
     // `biased;` makes ctrl-c win on concurrent readiness — avoids a
@@ -263,6 +266,9 @@ pub async fn run(
         }
         _ = &mut storage_writer_handle => {
             critical_exit(pt_label, "storage_writer");
+        }
+        _ = &mut recovery_handle => {
+            critical_exit(pt_label, "recovery");
         }
     }
 
--- a/indexer/src/operator/operator.rs
+++ b/indexer/src/operator/operator.rs
@@ -2,8 +2,8 @@ use crate::config::OperatorConfig;
 use crate::error::OperatorError;
 use crate::metrics;
 use crate::operator::{
-    feepayer_monitor, fetcher, processor, reconciliation, sender, DbTransactionWriter, RetryConfig,
-    RpcClientWithRetry,
+    feepayer_monitor, fetcher, processor, reconciliation, recovery, sender, DbTransactionWriter,
+    RetryConfig, RpcClientWithRetry, SignerUtil,
 };
 use crate::shutdown_utils::shutdown_operator;
 use crate::storage::Storage;
@@ -123,6 +123,7 @@ pub async fn run(
     let sender_commitment = config.rpc_commitment;
     let sender_source_rpc = source_rpc_client.clone();
     let sender_common_config = common_config.clone();
+    let recovery_storage_tx = storage_tx.clone();
     let sender_handle = tokio::spawn(async move {
         if let Err(e) = sender::run_sender(
             &sender_common_config,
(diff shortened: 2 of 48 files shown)


indexer/src/operator/recovery.rs
```diff
@@ -0,0 +1,720 @@
...
+            })))
+            .with_status(200)
+            .with_body(
+                serde_json::json!({
+                    "jsonrpc": "2.0",
+                    "id": 1,
+                    "result": {
+                        "slot": 100,
+                        "blockTime": 1_700_000_000i64,
+                        "meta": {
```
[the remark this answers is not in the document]
Huzaifa696: Fixed, threaded &CancellationToken into recover_once, checked between rows.

indexer/src/operator/recovery.rs
```diff
@@ -0,0 +1,720 @@
...
+        // Captured timestamp that does NOT match the seeded row's updated_at.
+        let stale = row.updated_at - chrono::Duration::seconds(60);
+        route_outcome(
+            &storage,
+            &row,
+            stale,
+            RecoveryAction::Demote,
+            ProgramType::Escrow,
+            &storage_tx,
+        )
```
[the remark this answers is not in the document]
Huzaifa696: replaced let _  with if let Err(e) = ... { warn!(...) }.

indexer/src/storage/postgres/db.rs
```diff
@@ -824,9 +830,154 @@ impl PostgresDb {
         .execute(&self.pool)
         .await?;
 
+        if result.rows_affected() == 0 {
+            warn!(
+                transaction_id,
+                "terminal write skipped: row no longer in processing"
+            );
+        }
```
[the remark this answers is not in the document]
Huzaifa696: update_transaction_status_internal now returns Result<bool>, Webhook still fires unconditionally.

indexer/src/shutdown_utils.rs
```diff
@@ -476,6 +479,24 @@ async fn perform_operator_shutdown_stages(
         }
     }
 
+    info!("Stage 6c: Waiting for recovery worker to drain...");
+    let recovery_result = tokio::time::timeout(
```
dev-jodee: should this happen before the storage writer drain? Since recovery uses storage_tx?

indexer/src/operator/operator.rs
```diff
@@ -173,6 +174,35 @@ pub async fn run(
...
     };
 
+    // Recovery worker. If the operator crashes between claiming a row
+    // (Pending → Processing) and writing its final status, the row would
+    // sit stuck in Processing forever. This worker periodically finds
+    // those stuck rows and resolves them. Runs on every operator since
+    // every operator has the same crash window.
+    let recovery_handle = {
+        let recovery_storage = storage.clone();
+        let recovery_rpc = source_rpc_client
```
dev-jodee: recovery should point to the private channel rpc right ? think source is the solana rpc?

dev-jodee (commented):
two questions and same comment in relation to the amount of comments on the pr, if we can remove the ones that don't add anything, it would be great

indexer/src/shutdown_utils.rs
Huzaifa696: Fixed, moved recovery drain before storage writer.

indexer/src/operator/operator.rs
Huzaifa696: Good catch,fixed. Recovery now uses PrivateChannel's rpc_client directly.

indexer/src/storage/postgres/db.rs
```diff
@@ -824,7 +827,130 @@ impl PostgresDb {
...
+        _reason: String,
+    ) -> Result<bool, sqlx::Error> {
+        let result = sqlx::query(
+            r#"
+            UPDATE transactions
+            SET status = 'manual_review',
+                processed_at = NOW()
+            WHERE id = $1
+              AND status = 'processing'
+              AND updated_at = $2
```
gitteri, line 1065: missing the reason (error_message)

gitteri: also worth adding this check to the unit test route_outcome_quarantine_writes_manual_review_and_sends_alert

integration/tests/indexer/stuck_processing_recovery.rs
```diff
@@ -0,0 +1,915 @@
...
+    let err = update.error_message.as_deref().unwrap_or("");
+    assert!(
+        err.starts_with("deposit idempotency:"),
+        "reason should match runbook substring: {err}"
+    );
+    assert_recovered_increment("escrow", "quarantined", "deposit", metric_before, "IT-5");
+    mock.shutdown().await;
+}
+
+// IT-6 omitted: -32601 arm is unreachable on the PrivateChannel RPC (PR #95).
```
gitteri: what's the reason for leaving the gap?

Huzaifa696: The "method not found" branch in the deposit lookup (getSignaturesForAddress) is unreachable given the current RPC implementation, so it's purely defensive. I've hardened it to quarantine rather than demote (avoids a double-mint if the method were ever missing) and added IT-6 to cover the case.

indexer/src/storage/postgres/db.rs
Huzaifa696: The `transactions` table doesn't have an `error_message` column. The quarantine reason is carried in the webhook alert rather than stored on the transaction row, which is consistent with the other `manual_review` writers.

The `reason` argument isn't used here, so I'm removing it for now.

indexer/src/operator/recovery.rs
```diff
@@ -0,0 +1,697 @@
...
+        row.status = TransactionStatus::Processing;
+        let captured = seed_processing_row(&mock, row.clone()).await;
+        let storage = Storage::Mock(mock.clone());
+        let (storage_tx, _rx) = mpsc::channel(8);
+
+        route_outcome(
+            &storage,
+            &row,
+            captured,
+            RecoveryAction::Complete {
```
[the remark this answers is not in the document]
Huzaifa696: Fixed, dropped the -32601 -> Ok(None) swallow so any RPC failure now returns Err, which recovery maps to Ambiguous -> quarantine instead of demote; covered by IT-6.

indexer/src/operator/recovery.rs
```diff
@@ -0,0 +1,697 @@
...
+pub(crate) const RECOVERY_BATCH_LIMIT: i64 = 100;
+
+/// Three-way outcome: uncertainty must NOT demote (double-mint risk).
+enum DepositOutcome {
+    Landed { signature: String },
+    NotLanded,
+    Ambiguous { reason: String },
+}
+
+/// Two-way outcome: on-chain SMT prevents duplicate release anyway.
```
sebatustra: Are we relying on the re-sent tx to fail to prevent a duplicate release? That feels
fragile. A more robust approach would mirror the remint flow: verify the release actually landed
via get_signature_statuses_with_history before demoting, instead of trusting the replay to be
rejected

Huzaifa696: Good catch. We can't apply that directly because a stuck `Processing` withdrawal has no persisted signature to check; release signatures are only stored at the `PendingRemint` transition, while in-flight signatures are lost on crash.

To address this, I now persist release signatures when they are broadcast. Recovery reads them and performs an on-chain finality check before deciding to complete, demote, wait, or quarantine. If verification is unavailable or inconclusive, we quarantine rather than demote and replay, eliminating the double-payout risk.

I also extracted the remint finality logic into a shared `classify_release_signatures` helper so both paths use the same logic. Existing remint tests remain unchanged and pass.

sebatustra (dismissed):

sebatustra (dismissed):

indexer/src/operator/sender/remint.rs
```diff
@@ -180,6 +180,96 @@ pub async fn execute_deferred_remint(
...
+    if response.value.len() != flat.len() {
+        return SigFinality::Uncertain(format!(
+            "RPC returned {} statuses for {} signatures",
+            response.value.len(),
+            flat.len()
+        ));
+    }
+
+    // Any sig finalized successfully → the release landed.
+    let finalized_success_index = response.value.iter().position(|signature_status| {
```
gitteri, line 221: This batch classifier is the fund-safety core for both remint and recovery, but every test that exercises it (here, in `recovery.rs`, in `stuck_processing_recovery.rs`, and the pre-existing remint tests) uses a single-element `value` array. The multi-signature orderings that actually matter are untested:

- `[finalized-failed, finalized-success]` must return `Landed` (proves success-wins, not first-match)
- `[finalized-failed, confirmed-success-not-yet-finalized]` must return `Live`, never `Dead` (the double-release trap)
- `[expired-null, live-null]` must return `Live`
- length-mismatch must return `Uncertain`

The logic reads correct, but this is exactly where a future refactor flips a withdrawal to double-release with all tests still green. Can we add 3-4 multi-sig cases directly against `classify_release_signatures`?

indexer/src/operator/recovery.rs
```diff
@@ -0,0 +1,984 @@
...
+            return DepositOutcome::Ambiguous {
+                reason: format!("deposit idempotency: rebuild: {e}"),
+            }
+        }
+    };
+    match find_existing_mint_signature(rpc_client, &builder).await {
+        Ok(Some(sig)) => DepositOutcome::Landed {
+            signature: sig.to_string(),
+        },
+        Ok(None) => DepositOutcome::NotLanded,
```
gitteri, line 207: `find_existing_mint_signature` only scans the most recent `MINT_IDEMPOTENCY_SIGNATURE_LOOKBACK_LIMIT` (1000) sigs on the recipient ATA. If the original mint landed but got buried past 1000 sigs on a hot ATA before recovery runs, this returns `Ok(None)` -> `NotLanded` -> demote -> re-mint -> **double-mint**. The 5-min stale window makes this more reachable than the live path (more time to accrue >1000 sigs), and it compounds with the unbounded requeue below.

This is the one spot deposit reprocessing is not truly idempotent. Options: paginate/widen the lookback, or gate completion on a durable mint-sig record (the same pattern this PR adds for releases via `pending_release_signatures`). At minimum a comment plus a tracking ticket.

indexer/src/operator/recovery.rs
```diff
@@ -0,0 +1,984 @@
...
+                Ok(false) => {
+                    debug!(
+                        id = row.id,
+                        "recovery skipped — another writer touched the row first"
+                    );
+                }
+                Err(e) => warn!(id = row.id, "recovery write error: {}", e),
+            }
+        }
+        RecoveryAction::Demote => {
```
gitteri, line 356: Requeue resets no counter and has no cap. The sender's retry budget is in-memory so re-pickup starts from zero, and `finality_check_attempts` doesn't apply to this path. A row that can't make progress (e.g. a deposit repeatedly judged `NotLanded` per the lookback-window comment above, or any persistent pre-mint failure) can cycle `pending -> processing -> (5 min) -> pending` forever, silently (`debug!`).

Recommend a requeue/attempt counter column with a cap that quarantines after N requeues, so a stuck row pages instead of ping-ponging indefinitely.

indexer/src/operator/sender/transaction.rs
```diff
@@ -315,6 +315,23 @@ pub(super) async fn send_and_confirm(
...
+                    if let Err(e) = state
+                        .storage
+                        .insert_release_signature(
+                            txid,
+                            signature.to_string(),
+                            last_valid_block_height as i64,
+                        )
+                        .await
+                    {
+                        warn!("failed to persist release signature for tx {txid}: {e}");
```
gitteri, line 336: Because this insert is best-effort, there's a wedge path worth calling out: release lands on-chain and advances the SMT root -> this insert fails (DB blip) -> operator crashes before the `Completed` write. Recovery then sees zero recorded sigs -> quarantine (correct and safe). But the DB is now missing a completed nonce the chain consumed, so the **next** operator restart hits `SmtRootMismatch` in `initialize_smt_state` and the whole withdrawal pipeline refuses to start until someone manually marks the row `Completed`. A single best-effort insert failure can halt all withdrawals.

Suggest: (a) note this consequence here, (b) add a runbook for the `SmtRootMismatch` startup halt (`processor.rs` already references a `withdrawal_pipeline_halt_runbook.md` that doesn't exist), and (c) an end-to-end test for landed-but-unrecorded -> quarantine -> divergence.

indexer/src/operator/sender/mint.rs
```diff
@@ -427,6 +425,7 @@ pub async fn find_existing_mint_signature(
 }
 
 /// Check recent ATA signatures for an already-confirmed mint carrying the given memo.
+/// Any RPC failure (including `-32601`) is returned as `Err` — callers decide.
```
gitteri, line 428: Removing the `-32601` fail-open changes the live JIT mint path, not just recovery. `find_existing_mint_signature` wraps this, and `transaction.rs` turns `Err` into `send_fatal_error`. So before: `-32601` -> `Ok(None)` -> mint proceeds; after: `-32601` -> deposit goes to `manual_review` without minting. Against any RPC that doesn't implement `getSignaturesForAddress`, every live deposit now halts to manual review.

Safer (no blind double-mint) and I think correct, but it's a steady-state operational change that isn't called out in the summary. Please confirm it's intended. It also makes the `deposit_manual_review.md` Path E "live mint path fails-open" rationale inaccurate (separate comment there).

docs/runbooks/deposit_manual_review.md
```diff
@@ -211,7 +217,56 @@ UPDATE transactions SET status = 'pending', updated_at = NOW()
  WHERE id = :transaction_id;
 ```
 
-### Path E - mint not in `AllowMint` allowlist
+### Path E - recovery-worker idempotency lookup failed
+
+`error_message` starts with `deposit idempotency:`. The recovery worker
+intentionally refuses to demote on RPC failure (the live mint path
+fails-open on the same condition; recovery cannot, because recovery has
```
gitteri: After the `-32601` change in `mint.rs`, the live path no longer fails-open on RPC failure: it fails closed via `send_fatal_error`. The asymmetry this paragraph leans on is gone (both live and recovery now fail closed, just to different terminal states). Reword so on-call isn't reasoning from a false premise.

indexer/src/storage/postgres/db.rs
```diff
@@ -423,6 +424,34 @@ impl PostgresDb {
         .execute(&self.pool)
         .await?;
 
+        // Broadcast release signatures written at send time; recovery reads
+        // them to verify a release landed before demoting (avoids double-payout).
+        sqlx::query(
+            r#"
+            CREATE TABLE IF NOT EXISTS pending_release_signatures (
+                id BIGSERIAL PRIMARY KEY,
+                transaction_id BIGINT NOT NULL REFERENCES transactions(id) ON DELETE CASCADE,
```
gitteri, line 449: This is the first FK into `transactions`, and it cascades. The new `withdrawal_manual_review.md` Path F has the operator run `DELETE FROM transactions WHERE id = ...` on a corrupt withdrawal, which will now silently cascade-delete that row's `pending_release_signatures`, i.e. the on-chain-proof artifacts recovery relies on. More broadly, hard-deleting financial rows by hand during an incident is irreversible, destroys forensic data exactly when it matters, and risks wrong-row deletion.

The only reason DELETE is needed at all is that the orphan-alert query (`get_orphan_deposit_ids_internal`) has no status filter. Consider a `deleted_at` tombstone plus `AND deleted_at IS NULL` on that query instead: keeps the audit trail and the proof rows, and makes "silence the alert" non-destructive. There's already precedent (`mint_status_history` uses a `blocked` tombstone). Probably a follow-up rather than this PR.

Huzaifa696:
I think your two comments share one root cause - so instead of patching each flow's crash-safety separately and living with the trade-offs, I'd suggest we fix it once at the source.

@Ilan - your two comments:

1. **[Withdrawal](https://github.com/solana-foundation/solana-private-channels/pull/127#discussion_r3352845543)** (*"Because this insert is best-effort…"*) - document/pin the best-effort-persist -> `SmtRootMismatch` wedge.
2. **[Deposit](https://github.com/solana-foundation/solana-private-channels/pull/127#discussion_r3352845457)** (*"…only scans the most recent 1000 sigs…"*) - either (a) widen the signature scan, or (b) the withdrawal-style persist-and-route-to-manual-review.

But (2b) doesn't work for deposits. Because the signature is saved *after* the mint is broadcast, a crash can leave a deposit with no saved signature in two different ways that recovery can't tell apart:

- we crashed *before* sending the mint, so nothing went on-chain (safe to re-mint), or
- we sent the mint and it may have landed, but crashed *before* saving its signature (re-minting would double-mint).

Both look identical to recovery (no signature on record), so the only safe move is to send both to manual review. For withdrawals that's fine, since the pipeline handles one at a time. But deposits run at thousands per second, so this would flood manual review. Option (2a), rescanning signature history for every stuck  adds unbounded RPC per row and won't scale imo. BTW SOLA3-8 in v2 audit scan directly points out this problem and its in [WIP](https://linear.app/solana-fndn/issue/EXO-38/ai-scan-sola3-8) but the soution that is being implemented is remint flow specific.

The root cause is **post-broadcast persistence**, partly forced by `sign_and_send` doing both steps in one call. The clean fix for both flows is **pre-broadcast persistence: split `sign` and `send`, persist the signature between them, and fail closed if the insert fails.** Then "no record" provably means "never broadcast," collapsing the two ambiguous crash points into one decidable state - deterministic recovery for deposits and withdrawals alike. It also *eliminates* comment 1's wedge case. I'd unify the deposit/withdrawal pending signature table behind this.

It's a fairly large change (the split touches every broadcast site + a table migration), so I'd appreciate a quick sanity check on direction before I implement.

indexer/src/operator/sender/remint.rs
Huzaifa696: Added 4 multi-sig tests pinning `classify_release_signatures` against adversarial orderings. commit(https://github.com/solana-foundation/solana-private-channels/pull/127/commits/603eaf0f4298b99bdc06a99237b4c5e1f4bc5601)

indexer/src/operator/recovery.rs
Huzaifa696: Per the [discussion](https://github.com/solana-foundation/solana-private-channels/pull/127#issuecomment-4625114229) below, fixing this at the root cause (post-broadcast persistence) in a follow-up PR ([ticket](https://linear.app/solana-fndn/issue/EXO-111/pre-broadcast-signature-persistence-crash-safety-fix-for-operator)) that persists the signature before broadcast, rather than widening the bounded scan here. commit(https://github.com/solana-foundation/solana-private-channels/pull/127/commits/603eaf0f4298b99bdc06a99237b4c5e1f4bc5601)

indexer/src/operator/recovery.rs
Huzaifa696: Added a durable capped requeue counter so a stuck `processing` row can't loop forever.

- **Cap:** new `recovery_requeue_attempts` increments in `try_requeue_processing` under the same CAS for atomicity. `MAX_RECOVERY_REQUEUE_ATTEMPTS = 3`
- **recovery counter reset:** The only valid reset is a deliberate operator re-arm (manual SQL today), so the runbooks set `recovery_requeue_attempts = 0` on every re-arm-to-`pending`.
- **Runbooks:** new Path G (deposit + withdrawal) for the `recovery requeues without progress` reason, plus the reset on every re-arm.
- **Tests:** unit - a requeue at the cap quarantines instead of looping; DB - a requeue bumps the counter, but a stale CAS that no-ops doesn't; integration `it13` - a row seeded at cap-1 quarantines and sends nothing.

commit(https://github.com/solana-foundation/solana-private-channels/pull/127/commits/603eaf0f4298b99bdc06a99237b4c5e1f4bc5601)

indexer/src/operator/sender/transaction.rs
Huzaifa696: Documented and pinned the wedge- the real fix is the pre-broadcast [follow-up](https://linear.app/solana-fndn/issue/EXO-111/pre-broadcast-signature-persistence-crash-safety-fix-for-operator).

Added `withdrawal_pipeline_halt_runbook.md` for the SMT-root-mismatch startup halt plus a one-line comment at the `insert_release_signature` site naming the consequence. Also added a test, `initialize_smt_state_halts_on_consumed_but_unrecorded_nonce` pinning the halt so a change that silently absorbs the divergence fails.

Notes:

1. Interim - obsolete once the pre-broadcast persistence follow-up removes the wedge.
2. Filed a separate [ticket](https://linear.app/solana-fndn/issue/EXO-112/document-the-poison-pill-malformed-withdrawal-halt-in-the-pipeline) for the malformed-withdrawal (poison-pill) halt steps; this runbook covers the SMT-mismatch path only.

commit(https://github.com/solana-foundation/solana-private-channels/pull/127/commits/603eaf0f4298b99bdc06a99237b4c5e1f4bc5601)

indexer/src/operator/sender/mint.rs
Huzaifa696: Documented the deliberate fail-closed stance at the lookup.

docs/runbooks/deposit_manual_review.md
Huzaifa696: Reworded Path E to drop the stale "live path fails open" claim - both the recovery and live paths now fail closed on RPC uncertainty.

indexer/src/operator/sender/mint.rs
Huzaifa696: This change was added later as result of this [comment](https://github.com/solana-foundation/solana-private-channels/pull/127#discussion_r3337190215) thats why its missing from the summary.

indexer/src/storage/postgres/db.rs
Huzaifa696: I've filed a separate [ticket](https://linear.app/solana-fndn/issue/8a6a211e-3d00-4317-9146-7821d7304fb5) to settle the referential strategy.

indexer/src/operator/recovery.rs
```diff
@@ -168,7 +171,20 @@ async fn decide_action(
             WithdrawalAction::LeaveProcessing { reason } => RecoveryAction::NoAction { reason },
             WithdrawalAction::Quarantine { reason } => RecoveryAction::Quarantine { reason },
         },
+    };
+    // Cap recovery requeue attempts. Rows that fail to make progress after
+    // MAX_RECOVERY_REQUEUE_ATTEMPTS are quarantined (and paged) rather than
+    // looping between Pending and Processing indefinitely.
+    if matches!(action, RecoveryAction::Demote)
+        && row.recovery_requeue_attempts + 1 >= MAX_RECOVERY_REQUEUE_ATTEMPTS
```
gitteri: i think this means max queue recovery attempts is actually 2, not 3 right? let's change this from `>=` to `>` so we get 3 max recovery attempts

indexer/src/operator/operator.rs
```diff
@@ -173,6 +174,29 @@ pub async fn run(
         tokio::spawn(async {})
     };
 
+    // Recovery worker: resolves rows stuck in Processing after a crash.
+    let recovery_handle = {
```
gitteri, line 178: Do we need to add this to the "Task supervision" set? (lines 228+)

indexer/src/operator/recovery.rs
Huzaifa696: Fixed, dropped the `+1` so it's now `recovery_requeue_attempts >= MAX`, giving a full 3 requeues before quarantine; bumped the cap-test seeds (unit + it13) from MAX-1 to MAX to match. 
commit (https://github.com/solana-foundation/solana-private-channels/pull/127/commits/3e2404948a901492c289b5745f9ed714e292d93a)

indexer/src/operator/operator.rs
Huzaifa696: Added `recovery_handle` to the supervision `select!` so an unexpected exit/panic now fires `critical_exit (OPERATOR_TASK_EXIT{task="recovery"} + restart`) like the other critical tasks. 
commit (https://github.com/solana-foundation/solana-private-channels/pull/127/commits/3e2404948a901492c289b5745f9ed714e292d93a)

indexer/src/storage/postgres/db.rs
```diff
@@ -889,7 +944,133 @@ impl PostgresDb {
...
+            transaction_cols::CREATED_AT,
+            transaction_cols::UPDATED_AT,
+            transaction_cols::PROCESSED_AT,
+            transaction_cols::COUNTERPART_SIGNATURE,
+            transaction_cols::REMINT_SIGNATURES,
+            transaction_cols::REMINT_LAST_VALID_BLOCK_HEIGHTS,
+            transaction_cols::PENDING_REMINT_DEADLINE_AT,
+            transaction_cols::FINALITY_CHECK_ATTEMPTS,
+            transaction_cols::RECOVERY_REQUEUE_ATTEMPTS,
+            // Filters
```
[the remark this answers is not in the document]
Huzaifa696: It's actually handled by the [`update_transactions_updated_at` BEFORE-UPDATE trigger](https://github.com/solana-foundation/solana-private-channels/blob/a6dc38bc7a9c4bfb6ba1daa926a602ce7edbc27a/indexer/src/storage/postgres/db.rs#L323-L342), which sets `updated_at = NOW()` on every `transactions` update.

gitteri (approved):

PR #133, opened 2026-05-22
fix(indexer): harden pending-remint gate against double payouts and restart resets   (SOLA3-19, SOLA3-1)

sebatustra:
Closes SOLA3-19. The pending-remint gate was reminting private-channel tokens while a prior low-fee `ReleaseFunds` broadcast could still finalize on Solana, producing a double payout. The 32s deadline ran shorter than blockhash validity (~60-90s) and the previous gate only suppressed reminting when a signature was already finalized.

The gate now classifies each stored signature by both its on-chain status and the blockhash's `last_valid_block_height`:
  - `finalized + success` → Completed
  - `finalized + error` → dead
  - `confirmed` / `processed` → still live (will finalize regardless of lvbh)
  - `null` + lvbh expired → dead
  - `null` + lvbh valid → still live

If any signature is still live, defer with a fresh deadline. Cap at `MAX_FINALITY_CHECK_ATTEMPTS`, then escalate to ManualReview.

 To support this, `last_valid_block_height` is now captured at broadcast time and persisted alongside `remint_signatures` in a new `BIGINT[]` column, plumbed through send → in-memory → DB → recovery.

Also closes SOLA3-1. `finality_check_attempts` was in-memory only, so a restart reset the safety budget to zero and let ambiguous PendingRemint rows survive indefinitely instead of escalating to ManualReview after the intended cap.

The counter is now persisted alongside `pending_remint_deadline_at` on every defer via a new `bump_pending_remint_finality_attempt` storage method, and restored on startup so the cap survives across restarts. Recovery defensively escalates corrupt (negative) counters to ManualReview, matching the existing pattern for the amount field.

Defer-time persistence is fail-closed: if the DB write fails, the entry escalates to ManualReview via `send_guaranteed` rather than continuing to defer with a counter we can't trust.
### Coverage Report

| Core | 9,620 | 11,100 | 86.7% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/26546659288) |
| Indexer | 16,163 | 18,903 | 85.5% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/26546659288) |
| Gateway | 994 | 1,164 | 85.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/26546659288) |
| Auth | 717 | 831 | 86.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/26546659288) |
| E2E Integration | 10,226 | 12,669 | 80.7% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/26546659288) |
| **Total** | **37,720** | **44,667** | **84.4%** | |

> Last updated: 2026-05-28 00:45:36 UTC by E2E Integration

### DIFF
--- a/indexer/src/operator/sender/remint.rs
+++ b/indexer/src/operator/sender/remint.rs
@@ -294,8 +294,10 @@ pub async fn process_pending_remints(
 
         // Walk the sigs to see if any could still land. Exit early on the
         // first one that isn't dead. Index-aligned with response.value
-        // (length equality enforced above).
-        let mut any_still_live = false;
+        // (length equality enforced above). Captures a reason describing
+        // why the broadcast could still land so the defer/ManualReview
+        // message can guide operator triage.
+        let mut live_reason: Option<String> = None;
         for (index, pending_sig) in entry.signatures.iter().enumerate() {
             let signature_status = &response.value[index];
 
@@ -306,9 +308,12 @@ pub async fn process_pending_remints(
                 if status.satisfies_commitment(CommitmentConfig::finalized()) {
                     continue;
                 }
-                // `confirmed` or `processed` — already included in a block,
+                // `confirmed` or `processed`: already included in a block,
                 // will finalize regardless of blockhash validity.
-                any_still_live = true;
+                live_reason = Some(
+                    "signature is on-chain (confirmed/processed) and awaiting finalization"
+                        .to_string(),
+                );
                 break;
             }
 
@@ -316,23 +321,16 @@ pub async fn process_pending_remints(
             if current_height > pending_sig.last_valid_block_height {
                 continue;
             }
-            any_still_live = true;
+            live_reason = Some(format!(
+                "signatures still within blockhash validity (current_height={})",
+                current_height
+            ));
             break;
         }
 
         // Case 2: at least one broadcast could still land, defer rather than remint.
-        if any_still_live {
-            defer_or_escalate(
-                &mut remaining,
-                entry,
-                &nonce_label,
-                &format!(
-                    "signatures still within blockhash validity (current_height={})",
-                    current_height
-                ),
-                storage_tx,
-            )
-            .await;
+        if let Some(reason) = live_reason {
+            defer_or_escalate(&mut remaining, entry, &nonce_label, &reason, storage_tx).await;
             continue;
         }
 
--- a/indexer/src/operator/sender/remint.rs
+++ b/indexer/src/operator/sender/remint.rs
@@ -237,6 +237,7 @@ pub async fn process_pending_remints(
                     entry,
                     &nonce_label,
                     &format!("signature status RPC failed: {}", e),
+                    &state.storage,
                     storage_tx,
                 )
                 .await;
@@ -257,6 +258,7 @@ pub async fn process_pending_remints(
                     response.value.len(),
                     sigs.len()
                 ),
+                &state.storage,
                 storage_tx,
             )
             .await;
@@ -285,6 +287,7 @@ pub async fn process_pending_remints(
                     entry,
                     &nonce_label,
                     &format!("block height RPC failed: {}", e),
+                    &state.storage,
                     storage_tx,
                 )
                 .await;
@@ -330,7 +333,15 @@ pub async fn process_pending_remints(
 
         // Case 2: at least one broadcast could still land, defer rather than remint.
         if let Some(reason) = live_reason {
-            defer_or_escalate(&mut remaining, entry, &nonce_label, &reason, storage_tx).await;
+            defer_or_escalate(
+                &mut remaining,
+                entry,
+                &nonce_label,
+                &reason,
+                &state.storage,
+                storage_tx,
+            )
+            .await;
             continue;
         }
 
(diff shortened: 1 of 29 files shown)


indexer/src/operator/sender/state.rs
```diff
@@ -262,22 +262,38 @@ impl SenderState {
...
-                &tx.trace_id,
-            )
-            .await
+                    .zip(&lvbhs)
+                    .map(|(sig_string, &lvbh)| {
+                        let signature = Signature::from_str(sig_string)
+                            .map_err(|e| format!("invalid withdrawal signature: {e}"))?;
+                        Ok(PendingSig {
+                            signature,
+                            last_valid_block_height: lvbh as u64,
```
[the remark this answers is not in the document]
sebatustra: Fixed. Replaced lvbh as u64 with u64::try_from(lvbh).map_err(...), mirroring the amount guard right above it.

indexer/src/operator/sender/remint.rs
```diff
@@ -198,118 +208,236 @@ pub async fn process_pending_remints(
...
+                ),
+                storage_tx,
+            )
+            .await;
+            continue;
+        }
+
+        // Case 3: every sig is finalized-failed or expired, safe to remint.
+        info!(
+            "All withdrawal signatures for nonce {} are finalized-failed or expired; attempting remint",
```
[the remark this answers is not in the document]
sebatustra: Fixed. Reason is now branch-specific

sebatustra:
@greptileai

indexer/src/operator/sender/remint.rs
```diff
@@ -177,17 +180,24 @@ pub async fn execute_deferred_remint(
...
+///   3. Every sig finalized-failed, or null-status with expired blockhash
+///      → remint.
+///
+/// RPC failures during classification fall through the same defer-or-escalate
+/// path as case 2.
 pub async fn process_pending_remints(
     state: &mut SenderState,
     storage_tx: &mpsc::Sender<TransactionStatusUpdate>,
 ) {
     let now = Utc::now();
```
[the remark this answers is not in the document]
sebatustra: gated the `get_block_height()` call on `response.value.iter().any(|s| s.is_none())` so it only fires when at least one null-status sig actually needs the lvbh check

indexer/src/operator/utils/rpc_util.rs
```diff
@@ -136,6 +136,28 @@ impl RpcClientWithRetry {
         .await
     }
 
+    /// Get recent blockhash + last_valid_block_height with retry.
+    pub async fn get_latest_blockhash_with_lvbh(
```
Taylor123: perhaps worth following the RPC naming convention and using `get_latest_blockhash_with_commitment` so that those familiar with the RPC methods will know what the intention is?

indexer/src/operator/sender/transaction.rs
```diff
@@ -1536,7 +1566,19 @@ mod tests {
...
+        state.pending_signatures.insert(
+            5,
+            vec![
+                PendingSig {
+                    signature: sig1,
+                    last_valid_block_height: 100,
+                },
+                PendingSig {
+                    signature: sig2,
+                    last_valid_block_height: 200,
```
Taylor123: nit: use variables that can also be used in the assertions below

indexer/src/operator/sender/remint.rs
```diff
@@ -198,123 +208,297 @@ pub async fn process_pending_remints(
...
+        );
+        execute_deferred_remint(state, &entry, storage_tx).await;
     }
 
+    // `remaining` = entries not yet due + entries `defer_or_escalate` re-queued.
     state.pending_remints = remaining;
 }
 
+/// Report a pending-remint entry as Completed because one of its withdrawal
+/// signatures landed on Solana.
```
Taylor123: nit: maybe more clear
```suggestion
/// signatures finalized on Solana.
```

indexer/src/operator/sender/remint.rs
```diff
@@ -198,123 +208,297 @@ pub async fn process_pending_remints(
...
+
+        // Case 1: if any sig finalized successfully, the withdrawal landed.
+        // Mark Completed and drop the entry.
+        let finalized_success_index = response.value.iter().position(|signature_status| {
+            signature_status.as_ref().is_some_and(|status| {
+                status.satisfies_commitment(CommitmentConfig::finalized()) && status.err.is_none()
+            })
+        });
+        if let Some(index) = finalized_success_index {
+            send_completed(storage_tx, &entry, &nonce_label, sigs[index]).await;
```
Taylor123, line 276: what are the scenarios where `entry.ctx.transaction_id` is None? If it's an impossibility within `send_completed`, should there be an error thrown instead of return?

sebatustra: transaction_id can't be None here (entries are only queued for withdrawals, which always have a DB id; the enqueue path guards against None at transaction.rs:738). Added an error! log on the unreachable branch instead of panicking, since a panic would take down the sender loop.

Huzaifa696 (approved):

PR #136, opened 2026-06-02
feat(indexer): ingest BlockMint and mirror current mint status

Huzaifa696:
## Summary

The indexer parsed `AllowMint` but ignored the on-chain `BlockMint` instruction. The deposit gate therefore kept treating a blocked mint as allowed, and deposits against it were still accepted.

- **Parse `BlockMint`:** the indexer now recognizes the block instruction and reads the mint it targets.
- **Record the transition:** a block is appended to mint status history as a point-in-time `"blocked"` transition, alongside the existing `"allowed"` transitions.
- **Mirror current state:** each mint's current status is exposed as a derived column, kept in sync from history after every finalized slot.
- **Migration:** adds the current-status column, defaulting existing mints to allowed.

## Design decision

- **History is the source of truth; current status is derived.** The deposit gate keys off point-in-time status history, so a block takes effect the moment its transition lands. The mirror is always recomputed from history, never written directly, so a replayed or out-of-order write can't regress current state.

## Testing

- **Parser (unit):** valid 8-account `BlockMint` parses to instance/mint; insufficient accounts errors.
- **Conversion (unit):** `BlockMint` yields a `"blocked"` transition with no mint row; instance-mismatch instructions are dropped.
- **Storage (unit):** allowed→blocked→allowed round-trip, stale-replay no-op, missing-row no-op; upsert-on-conflict preserves `status`.
- **Mirror semantics (unit):** mock `sync_mint_status` mirrors the Postgres `DISTINCT ON` / latest-slot behavior.
- **Postgres (integration):** migration + `sync_mint_status_internal` against real Postgres.
- **End-to-end (real Postgres):** `deposit_gate_rejects_after_block_against_real_postgres` — a deposit in the allowed window is accepted, and a deposit after the block slot is rejected.

## Scope

**Out of scope:** distinguishing intra-slot allow/block ordering; backfilling `mint_status_history` for blocks that landed before this indexer change.

### Coverage Report

| Core | 11,036 | 12,647 | 87.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27656673447) |
| Indexer | 20,623 | 23,492 | 87.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27656673447) |
| Gateway | 994 | 1,164 | 85.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27656673447) |
| Auth | 717 | 831 | 86.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27656673447) |
| Withdraw Program | 120 | 232 | 51.7% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27530747213) |
| Escrow Program | 1,186 | 1,973 | 60.1% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27530747213) |
| DvP Swap Program | 247 | 1,112 | 22.2% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27530747213) |
| E2E Integration | 11,637 | 14,472 | 80.4% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27656673447) |
| **Total** | **46,313** | **54,811** | **84.5%** | |

> Last updated: 2026-06-17 00:41:22 UTC by E2E Integration


indexer/src/indexer/transaction_processor.rs
```diff
@@ -182,6 +185,22 @@ impl TransactionProcessor {
             }
         }
 
+        // Derive the `mints.status` mirror for each touched mint from history.
+        // Gated on the writes above, so the mirror never leads the timeline.
+        if send_checkpoint && !mint_statuses.is_empty() {
+            let touched: Vec<String> = mint_statuses
+                .iter()
+                .map(|s| s.mint_address.clone())
+                .collect();
```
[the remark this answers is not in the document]
Huzaifa696: Done, added `sort_unstable()` + `dedup()` before the `sync_mint_status` call to make the intent explicit. 
No correctness change (`ANY($1)` + `DISTINCT ON` already collapse duplicates); a mint only repeats in the same-slot allow+block case, which is pinned as a known limitation.

indexer/src/indexer/datasource/common/parser/escrow.rs
```diff
@@ -673,6 +724,40 @@ mod tests {
...
+            // instance @ index 2, mint @ index 3 — read straight from accounts.
+            assert_eq!(accounts.instance, account_keys[2]);
+            assert_eq!(accounts.mint, account_keys[3]);
+        } else {
+            panic!("Expected BlockMint instruction");
+        }
+    }
+
+    #[test]
+    fn test_block_mint_insufficient_accounts() {
```
sebatustra, line 750: what is this test testing? if it's just checking this condition `if instruction.accounts.len() < 8` in `parse_block_mint` I think we can remove it

Huzaifa696: It pins the bounds-check: without the < 8 guard, parse_block_mint would index accounts[7] out of bounds and panic on a malformed/truncated instruction.  

It also matches the _valid/_insufficient pairing every other parser in this file has (deposit, release_funds, etc.).  e.g

```RUST
#[test]
    fn test_deposit_insufficient_accounts() {
        let borsh_data = create_deposit_borsh_data();
        let instruction = create_instruction_with_accounts(11, "dummy".to_string()); // Only 11 accounts (need 12)
        let account_keys = create_n_account_keys(11);

        let result = parse_deposit(&borsh_data, &instruction, &account_keys, &[]);

        assert!(result.is_err());
        let err = result.unwrap_err().to_string();
        assert!(err.contains("Insufficient accounts"), "Error: {}", err);
    }
```

sebatustra (dismissed):

sebatustra (approved):

PR #140, opened 2026-06-05
fix(devnet): keep auth opt-in

onspeedhp:
## Summary

Keep the devnet stack unauthenticated by default by leaving `JWT_SECRET` blank in `.env.devnet`, matching the gateway/auth docs and the fact that the `auth` service only starts when the `auth` profile is enabled.

Also adds a small auth-service config validation so running the auth service with an empty signing secret fails fast instead of issuing JWTs signed with an empty secret.

## Root cause

`.env.devnet` set `JWT_SECRET=change_me_in_devnet`, so the gateway enabled RBAC in the default devnet compose stack even though the `auth` service is profile-gated and not started by `docker compose ... up -d`. This can make devnet testers hit 401s for gated RPC methods without any auth service running to issue tokens.

## Changes

- Leave `JWT_SECRET` blank in `.env.devnet` for the default unauthenticated gateway.
- Document that devnet auth/RBAC requires setting `JWT_SECRET` and starting Compose with `--profile auth`.
- Reject empty `JWT_SECRET` when the auth service is actually started.
- Update auth README to describe `JWT_SECRET` as non-empty.

## Validation

- `cargo fmt --package auth`
- `cargo test -p auth config::tests`
- `cargo test -p auth --lib`
- `git diff --check`
- `docker compose -f docker-compose.devnet.yml --env-file versions.env --env-file .env.devnet config --format json | jq -r ' .services.gateway.environment.JWT_SECRET '` returns an empty value
- `docker compose -f docker-compose.devnet.yml --env-file versions.env --env-file .env.devnet config --format json | jq -r ' .services | keys[] '` includes `gateway` but not `auth`

Note: `cargo test -p auth` was also run locally, but the admin CLI tests that start Testcontainers failed because Docker Desktop was manually paused in this local environment (`DockerResponseServerError 503`). The auth library tests all pass.


Huzaifa696:
Changes look good, the failing "Initialize Coverage Table" check is just our workflow not handling fork PRs. 
I'll push a workflow fix so this PR can be merged.

onspeedhp:
Thanks for confirming. I checked the run logs as well: the Rust tests themselves passed, and the failures are from the coverage PR-body update path on the fork PR (`Resource not accessible by integration`, then the coverage table is missing). I will leave this PR as-is and wait for the workflow fix.

Huzaifa696:
There is no empty JWT_SECRET guard anywhere in auth/, gateway/ or core/ on main. But Config has grown to 14 fields and already has its own validate() -> Result<(), String> at auth/src/config.rs:110, so this adds a duplicate method and the test helper no longer compiles. Happy to take it as a few lines folded into the existing validate() if you want to rebase.

PR #141, opened 2026-06-05
Fix/cantina dvp audit

sebatustra:
# DvP Program: Cantina Audit Remediation

Addresses the Cantina security review of the DvP swap program. Each finding is listed in audit order with how it was resolved and the commit that did it. All 21 findings are resolved: fixed or documented in code, except 3.3.11, which is acknowledged with rationale (no code change).

## Medium Risk

- **3.1.1 Closed unfunded mint can block funded-leg recovery.** Store each leg's token program in `SwapDvp` at create, and validate terminal transfers against that stored value instead of the (possibly closed) mint's owner, so a closed mint can no longer block refunds. (`0d065bf`)
- **3.1.2 Memo-required ATAs can block DvP recovery.** Detect destinations with the `MemoTransfer` extension and emit a Memo CPI immediately before the transfer in settle/cancel/reject/reclaim. (`9d1d482`)
- **3.1.3 Post-expiry recovery is all-or-nothing.** Remove the expiry gate on `ReclaimDvp` so each party can recover their own leg after expiry, independent of a problem on the other leg. Safe because `Settle` is independently gated on `now <= expiry` and on both legs being funded. (`18f3113`)
- **3.1.4 Nonce-reusable PDA address after close.** Create a per-`(seeds, nonce)` tombstone PDA at create that no instruction closes; `CreateDvp` rejects when it already exists, so a closed trade's address cannot be re-instantiated with different terms. (`5cb26ba`)

## Low Risk

- **3.2.1 `NonTransferable` mints can brick funded DvPs.** Reject `NonTransferable` mints at `CreateDvp`, since a balance reaching such an escrow could never be transferred out. (`0de7f6d`)
- **3.2.2 Unsynced WSOL lamports bypass refunds.** `get_token_account_balance` now `SyncNative`s a native escrow before reading it, so lamports sent via a raw system transfer are counted and refunded to the depositor instead of leaking to the close recipient. (`93e3461`)
- **3.2.3 Dust surplus can block settlement when refund ATAs are absent.** Documented that all four user ATAs at settle, including the two surplus-refund ATAs, are required accounts and must be pre-initialized. (`efe6d83`)
- **3.2.4 Destination-dependent `TransferHook` extras can block surplus settlement.** Documented that a hook resolving accounts from the transfer destination or amount can revert the surplus refund, and that clients should fund hook-bearing escrows at exactly the agreed amount. (`9f90334`)
- **3.2.5 Permissionless `CreateDvp` permits slot squatting.** Documented that create is permissionless and a record is not proof of agreement; clients must use a random nonce and verify the stored terms before funding or settling. (`ea9f519`)
- **3.2.6 `Clock.unix_timestamp` drift.** Documented that all time gates evaluate cluster time, not wall-clock, and recommended a drift budget for integrators. (`83d6901`)

## Informational

- **3.3.1 Closeable mints can bypass the extension deny-list.** Documented as a trusted-authority risk. Not blacklisted, since that would contradict 3.1.1's deliberate support for closeable mints. (`f135174`)
- **3.3.2 Mint authorities can freeze or claw back escrows.** Documented freeze authority and `DefaultAccountState` as trusted-authority risks, and recommended surfacing them in settlement UIs. (`e566cad`)
- **3.3.3 Closed-account lamports go to terminal signers.** Documented the rent-recipient model: the create payer is not recorded as the rent recipient, so closed-account lamports go to the closing party. (`159c7bb`)
- **3.3.4 Confidential transfer fee denial is undocumented.** Added `ConfidentialTransferFeeConfig`, and the previously omitted `NonTransferable`, to the documented blocked-extension set. (`f20ccb3`)
- **3.3.5 TransferHook account cap can brick terminal instructions.** Documented the 32-accounts-per-leg cap and that a mutable hook authority can exceed it, as a trusted-authority risk. (`dbbb5ed`)
- **3.3.6 Transfer hooks can receive caller signer privileges.** Documented that hook extras are forwarded with the client-set signer/writable flags, so clients must never include a transaction signer or unrelated writable account as an extra. (`3b5b243`)
- **3.3.7 README omits Token-2022 funding requirements.** Documented that extension-aware mints must be funded with `TransferChecked`, including the per-extension account requirements. (`16f69c5`)
- **3.3.8 SBF build leaves deploy keypairs that don't match declared program IDs.** Added a `make verify-program-id` pre-deploy guard and documented that the current `declare_id!` is a temporary placeholder, with the real-deploy workflow. (`2032ae3`)
- **3.3.9 IDL does not describe TransferHook extra accounts.** Documented that the generated IDL lists only the fixed accounts, so clients and indexers must add the trailing hook accounts per the README. (`d64b714`)
- **3.3.10 Unbounded `expiry_timestamp`.** Cap expiry at one year past creation, expressed as a duration from now, bounding escrow rent lock-up. (`8b926b2`)
- **3.3.11 Settlement window may be configured too narrow.** Acknowledged, not fixed. There is no security impact, as funds are always recoverable via Reject; left as a client-side configuration concern rather than an on-chain check.

### Coverage Report

| Core | - | - | - | - |
| Indexer | - | - | - | - |
| Gateway | - | - | - | - |
| Auth | - | - | - | - |
| Withdraw Program | 120 | 232 | 51.7% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27047726799) |
| Escrow Program | 1,185 | 1,963 | 60.4% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27047726799) |
| DvP Swap Program | 247 | 1,112 | 22.2% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27047726799) |
| E2E Integration | 10,449 | 12,988 | 80.5% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27047726821) |
| **Total** | **11,754** | **15,183** | **77.4%** | |

> Last updated: 2026-06-06 01:12:39 UTC by E2E Integration


dvp-swap-program/program/src/processor/shared/token_utils.rs
```diff
@@ -161,6 +209,36 @@ pub fn validate_mint_extensions(mint_info: &AccountView) -> ProgramResult {
...
+        memo_program.address() == &MEMO_PROGRAM_ID,
+        ProgramError::IncorrectProgramId
+    );
+    let instruction = InstructionView {
+        program_id: &MEMO_PROGRAM_ID,
+        accounts: &[],
+        data: MEMO_TEXT,
+    };
+    invoke_signed_with_bounds::<0>(&instruction, &[], &[])
+}
```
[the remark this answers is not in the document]
sebatustra: The check uses `get_processed_sibling_instruction`, not the instructions sysvar — 
see `check_previous_sibling_instruction_is_memo`. That syscall is CPI-aware: it 
returns the sibling at the same nesting level. We emit the Memo CPI and the 
TransferChecked CPI as siblings under the same DvP instruction (memo first), so 
`get_processed_sibling_instruction`(0) returns the Memo and the check passes. The 
memo-required tests run the real Token-2022 BPF in LiteSVM and pass, which would
be impossible if the memo weren't satisfying the check. No change needed.

source: https://github.com/solana-program/token-2022/blob/program%40v10.0.0/program/src/extension/memo_transfer/mod.rs

dvp-swap-program/program/src/processor/shared/token_utils.rs
```diff
@@ -55,17 +64,46 @@ pub fn verify_canonical_ata(
...
+        }
+        SyncNative {
+            native_token: info,
+            token_program: &TOKEN_2022_PROGRAM_ID,
+        }
+        .invoke()?;
         let data = info.try_borrow()?;
-        let account = unsafe { Token2022Account::from_bytes_unchecked(&data) };
-        return Ok(account.amount());
+        return Ok(unsafe { Token2022Account::from_bytes_unchecked(&data) }.amount());
```
Taylor123: could this logic be consolidated? It looks more/less the same except the program ID being used and the Account type being deserialized

dvp-swap-program/program/src/processor/create_dvp.rs
```diff
@@ -107,6 +118,23 @@ pub fn process_create_dvp(
...
+    // this trade's seeds. It's created below and never closed, so a
+    // non-system owner here means the nonce was already used — reject
+    // before re-creating the (closed) SwapDvp at the same address.
+    let (expected_tombstone, tombstone_bump) = Address::find_program_address(
+        &[NONCE_TOMBSTONE_SEED, expected_swap_dvp.as_ref()],
+        program_id,
+    );
+    require!(
+        nonce_tombstone_info.address() == &expected_tombstone,
+        ProgramError::InvalidSeeds
```
Taylor123: nit: InvalidAccountData might be a better error since this error isn't a failure to find a PDA. Rather, it's a mismatch between supplied and expected addresses.

gitteri:
@greptileai

dvp-swap-program/program/src/processor/shared/token_utils.rs
```diff
@@ -87,28 +126,37 @@ pub fn get_mint_decimals(mint_info: &AccountView) -> Result<u8, ProgramError> {
...
+///   settle/refund/reclaim can drain it, stranding both legs.
 ///
 /// Everything else is allowed, including `Pausable`, `PermanentDelegate`,
-/// `NonTransferable`, `DefaultAccountState`, `TransferHook`, etc.
-/// These fail *loudly* (reverted CPI, atomic rollback) rather than
-/// silently — funds stay in escrow and recover once the blocking
-/// condition lifts. `PermanentDelegate` is a deliberate carve-out for
-/// regulated RWA tokens (issuer/transfer-agent clawback is an
-/// intrinsic property of the asset).
+/// `DefaultAccountState`, `TransferHook`, etc. These fail *loudly*
```
gitteri, line 141: what fails loudly?

dvp-swap-program/program/src/processor/create_dvp.rs
```diff
@@ -107,6 +118,23 @@ pub fn process_create_dvp(
         ProgramError::InvalidSeeds
     );
 
+    // Nonce tombstone, derived from the SwapDvp address so it's 1:1 with
+    // this trade's seeds. It's created below and never closed, so a
+    // non-system owner here means the nonce was already used — reject
+    // before re-creating the (closed) SwapDvp at the same address.
+    let (expected_tombstone, tombstone_bump) = Address::find_program_address(
```
gitteri, line 125: what's the full seed to create the pda? does it include the expiry timestamp or a ref id or something? if not then someone can't execute the same dvp swap even if they want to do it twice (amount a for amount b)

dvp-swap-program/program/src/processor/shared/token_utils.rs
sebatustra: the token transfer fails loudly, for example if a mint is Paused the Redeem ix will simply fail, until the mint is unpaused by the authority (nothing moves, funds stay in escrow). This is in contrast with the transfer fee extension, where the amounts would "silently" differ from the terms stored in the SwapDvp account

dvp-swap-program/program/src/processor/create_dvp.rs
sebatustra: the full seeds are = `[b"dvp", settlement_authority, user_a, user_b, mint_a, mint_b, nonce.to_le_bytes(), bump]` . it does include nonce (u64), like a ref_id, its job is to disambiguate identical DvPs

sebatustra: this tombstone pda was added in order to make the nonce non-reusable (finding 3.1.4)

gitteri: cool makes sense !

gitteri (approved):

PR #143, opened 2026-06-08
fix(SOLA3-16): harden Postgres credential and network exposure for the private channel

Huzaifa696:
## Summary

Addresses SOLA3-16. Tracked config templates (`.env.*`, `docker-compose*.yml`, indexer `*.toml`) shipped working/default secrets, and a blank credential was accepted silently: a blanked `${POSTGRES_PASSWORD}` interpolates into a passwordless URL (`postgres://user:@host/db`) that sqlx/Postgres will still connect with, so an operator who forgot to set the secret comes up on an unintended connection with no error.

This PR removes working secrets from tracked files and makes every credential path fail closed: the core node, the indexer, and the operator signer each refuse to start on a blank or missing credential instead of connecting, and a pre-launch check rejects an unset secret before any service boots.

## What's added

- **Blanked secrets in tracked config.** Every committed `.env.*`, compose file, and indexer TOML now ships empty secret placeholders only — no working credentials in the repo. Config interfaces tightened (privatized) alongside.
- **Fail-closed DB credential guard.** `database_url_has_password` (core `postgres.rs` and indexer `db.rs`) rejects an empty or missing password component before opening a connection; unparseable URLs are deferred to sqlx so the real connect error still surfaces. Covers the no-password (`user@host`), empty-password (`user:@host`), and no-userinfo forms.
- **Fail-closed signer guard.** `signer_util` rejects a set-but-empty `ADMIN_PRIVATE_KEY` — `env::var` returns `Ok("")` for a blank var, which an explicit emptiness check catches.
- **Pre-launch env check.** New `scripts/check-required-env.sh` sources the env-file chain and asserts `POSTGRES_PASSWORD` / `POSTGRES_REPLICATION_PASSWORD` are non-empty; wired into `run.sh`, the Makefiles, and the Dockerfile so the stack won't start with blank secrets. Docs (`README`, `DEVNET_QUICKSTART`, devnet README) point operators at the required vars.

## Tests

- **DB password guard (unit, core + indexer):** classifies empty-password, no-password, no-userinfo, real, and percent-encoded URLs; unparseable URLs pass through to sqlx.
- **`PostgresAccountsDB::new` fails closed (unit):** a blank password is rejected before any connection attempt.
- **Signer guard (unit):** `ADMIN_SIGNER=memory` with a set-but-empty/whitespace `ADMIN_PRIVATE_KEY` fails closed with a message flagging the empty var.

## Dockerfile build fix

Unrelated to the credential hardening, but folded into this branch because the image otherwise fails to build: the `Dockerfile` now builds the `dvp-swap-program` alongside the escrow/withdraw programs. It copies the program's Cargo manifests (preserving the dependency-cache layering), builds the SBF artifact with `cargo-build-sbf`, copies `dvp_swap_program.so` into `/out/deploy`, and replaces the `core/precompiles/dvp_swap_program.so` symlink with the real artifact. This is required because `core/src/accounts/precompiles.rs` embeds that `.so` via `include_bytes!` — without building it in the image, the core binary won't compile.

### Coverage Report

| Core | 10,754 | 12,361 | 87.0% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27255727562) |
| Indexer | 18,801 | 21,787 | 86.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27255727562) |
| Gateway | 994 | 1,164 | 85.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27255727562) |
| Auth | 717 | 831 | 86.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27255727562) |
| E2E Integration | 11,200 | 13,892 | 80.6% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27255727562) |
| **Total** | **42,466** | **50,035** | **84.9%** | |

> Last updated: 2026-06-10 06:13:37 UTC by E2E Integration

### DIFF
--- a/core/src/accounts/postgres.rs
+++ b/core/src/accounts/postgres.rs
@@ -32,13 +32,13 @@ pub struct PostgresAccountsDB {
     pub read_only: bool,
 }
 
-/// Returns true when the database URL carries a non-empty password component.
-fn database_url_has_password(database_url: &str) -> bool {
+/// Returns true when the URL parses and its password is absent or empty (a blanked secret).
+fn database_url_password_is_blank(database_url: &str) -> bool {
     match url::Url::parse(database_url) {
         // None (no password) and Some("") (blanked secret) are both missing credentials.
-        Ok(parsed) => parsed.password().is_some_and(|p| !p.is_empty()),
-        // Leave unparseable URLs for sqlx to reject on connect.
-        Err(_) => true,
+        Ok(parsed) => parsed.password().unwrap_or("").is_empty(),
+        // Leave unparseable URLs for sqlx to reject with the real connection error.
+        Err(_) => false,
     }
 }
 
@@ -48,7 +48,7 @@ impl PostgresAccountsDB {
         read_only: bool,
     ) -> Result<Self, Box<dyn std::error::Error>> {
         // Fail closed: reject a blank password before connecting (blanked env templates interpolate an empty ${POSTGRES_PASSWORD} into a passwordless URL).
-        if !database_url_has_password(database_url) {
+        if database_url_password_is_blank(database_url) {
             return Err(
                 "database_url password component is empty; set a non-empty POSTGRES_PASSWORD"
                     .into(),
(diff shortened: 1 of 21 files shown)


indexer/src/storage/postgres/db.rs
```diff
@@ -38,8 +38,38 @@ pub struct PostgresDb {
...
+    let userinfo = match after_scheme.split_once('@') {
+        Some((userinfo, _)) => userinfo,
+        None => return false,
+    };
+    // Password is the part after the first ':' in the userinfo; empty means missing.
+    match userinfo.split_once(':') {
+        Some((_, password)) => !password.is_empty(),
+        None => false,
+    }
+}
```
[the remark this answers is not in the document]
Huzaifa696: Fixed, `database_url_has_password` now uses `url::Url` (matching core)

core/src/accounts/postgres.rs
```diff
@@ -32,12 +32,30 @@ pub struct PostgresAccountsDB {
     pub read_only: bool,
 }
 
+/// Returns true when the database URL carries a non-empty password component.
+fn database_url_has_password(database_url: &str) -> bool {
+    match url::Url::parse(database_url) {
+        // None (no password) and Some("") (blanked secret) are both missing credentials.
+        Ok(parsed) => parsed.password().is_some_and(|p| !p.is_empty()),
+        // Leave unparseable URLs for sqlx to reject on connect.
+        Err(_) => true,
```
sebatustra: it feels awkward to return true when the url is unparseable. Is there a reason to not rely on sqlx to reject it and just return false?

indexer/src/storage/postgres/db.rs
```diff
@@ -38,8 +38,29 @@ pub struct PostgresDb {
...
 }
 
+// Returns true when the database URL carries a non-empty password component.
+// Kept in sync with the identical guard in core's accounts/postgres.rs.
+fn database_url_has_password(database_url: &str) -> bool {
+    match url::Url::parse(database_url) {
+        // None (no password) and Some("") (blanked secret) are both missing credentials.
+        Ok(parsed) => parsed.password().is_some_and(|p| !p.is_empty()),
+        // Leave unparseable URLs for sqlx to reject on connect.
+        Err(_) => true,
```
sebatustra: same comment as above

core/src/accounts/postgres.rs
Huzaifa696: Inverted it to `database_url_password_is_blank` (core + indexer)

Inverting makes `Err(_) => false` read naturally, unparseable isn't blank, so it defers to sqlx.

indexer/src/storage/postgres/db.rs
Huzaifa696: see [comment](https://github.com/solana-foundation/solana-private-channels/pull/143#discussion_r3385768804)

sebatustra (approved):

PR #144, opened 2026-06-09
SOLA3-8: durably record confirmed remint and fail closed on idempotency lookup

sebatustra:
### Problem
After a remint confirmed on-chain, persistence was async (channel →
`DbTransactionWriter`). If the operator died between confirmation and that write, the
row stayed `PendingRemint`. On restart it was re-hydrated and reminted again. The only
guard, `attempt_remint`'s idempotency lookup, was **fail-open**: on any lookup error it
logged a warning and minted anyway, producing duplicate channel assets from one failed
withdrawal.

### Fix
1. **Durable record before ack.** New `record_remint_result` does one guarded `UPDATE`
  (`status → FailedReminted`, `landed_remint_signature`, `processed_at`), called
  synchronously in `execute_deferred_remint` before the channel send. New
  `landed_remint_signature` column (idempotent migration).
2. **Recovery keys off the durable record.** The status flip means
  `get_pending_remint_transactions` (filters `status = 'pending_remint'`) no longer
  returns the row, so restart recovery never re-queues it.
3. **Fail closed on unavailable lookup.** `attempt_remint` now escalates to
  `ManualReview` when the idempotency lookup can't run, instead of minting blind
### Coverage Report

| Core | 10,361 | 11,940 | 86.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27178141777) |
| Indexer | 18,806 | 21,906 | 85.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27178141777) |
| Gateway | 994 | 1,164 | 85.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27178141777) |
| Auth | 717 | 831 | 86.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27178141777) |
| E2E Integration | 11,198 | 13,926 | 80.4% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27178141777) |
| **Total** | **42,076** | **49,767** | **84.5%** | |

> Last updated: 2026-06-09 01:57:02 UTC by E2E Integration

### DIFF
--- a/indexer/src/operator/sender/remint.rs
+++ b/indexer/src/operator/sender/remint.rs
@@ -70,11 +70,13 @@ async fn attempt_remint(
             return Ok(existing_signature);
         }
         Ok(None) => {}
+        // Fail closed: an unverifiable lookup escalates to ManualReview (via the
+        // Err arm of execute_deferred_remint) instead of risking a duplicate remint.
         Err(e) => {
-            warn!(
-                "Remint idempotency lookup failed for transaction {}: {}; proceeding with send",
+            return Err(format!(
+                "idempotency lookup unavailable for transaction {}: {}; refusing to remint",
                 info.transaction_id, e
-            );
+            ));
         }
     }
 
@@ -125,33 +127,63 @@ pub async fn execute_deferred_remint(
                 "Withdrawal failed but tokens reminted successfully: {}",
                 signature
             );
-            if let Some(transaction_id) = entry.ctx.transaction_id {
-                if let Err(e) = send_guaranteed(
-                    storage_tx,
-                    TransactionStatusUpdate {
-                        transaction_id,
-                        trace_id: entry.ctx.trace_id.clone(),
-                        status: TransactionStatus::FailedReminted,
-                        counterpart_signature: None,
-                        processed_at: Some(Utc::now()),
-                        error_message: Some(entry.original_error.clone()),
-                        remint_signature: Some(signature.to_string()),
-                        remint_attempted: true,
-                    },
-                    "transaction status update",
-                )
-                .await
-                {
-                    error!(
-                        "Failed to send FailedReminted status for txn {}: {}. \
-                         Remint sig {} confirmed on-chain but not recorded.",
-                        transaction_id, e, signature
-                    );
-                }
-            } else {
+
+            // Always Some for remint entries: they are only queued for a failed
+            // withdrawal, which has a DB row. With no id there is no row to record
+            // the landed remint against and no status message to key, so the only
+            // action is a loud log. The remint already confirmed on-chain.
+            let Some(transaction_id) = entry.ctx.transaction_id else {
                 error!(
-                    "Remint succeeded (sig: {}) but no transaction_id to record status",
+                    "Remint confirmed (sig: {}) but entry has no transaction_id; \
+                     cannot record FailedReminted",
                     signature
+                );
+                return;
+            };
+
+            // Durably record the landed remint before the async channel send.
+            // This flips status to FailedReminted now, so a crash in the window
+            // before the writer runs can no longer leave the row PendingRemint
+            // for restart recovery to pick up and remint a second time.
+            //
+            // If this write fails we do not abort: the channel send below still
+            // drives the row to FailedReminted (its UPDATE accepts pending_remint),
+            // which is enough to stop replay. Only landed_remint_signature is lost.
+            if let Err(persist_err) = state
+                .storage
+                .record_remint_result(transaction_id, signature.to_string())
+                .await
+            {
+                error!(
+                    "Remint sig {} confirmed but durable persist failed for txn {}: {}; \
+                     falling back to async status writer",
+                    signature, transaction_id, persist_err
+                );
+            }
+
+            // Drives the webhook alert, and is the fallback status write when the
+            // durable persist above errored. A no-op if the row is already
+            // FailedReminted (the UPDATE guard rejects non-pending_remint rows).
+            if let Err(e) = send_guaranteed(
+                storage_tx,
+                TransactionStatusUpdate {
+                    transaction_id,
+                    trace_id: entry.ctx.trace_id.clone(),
+                    status: TransactionStatus::FailedReminted,
+                    counterpart_signature: None,
+                    processed_at: Some(Utc::now()),
+                    error_message: Some(entry.original_error.clone()),
+                    remint_signature: Some(signature.to_string()),
+                    remint_attempted: true,
+                },
+                "transaction status update",
+            )
+            .await
+            {
+                error!(
+                    "Failed to send FailedReminted status for txn {}: {}. \
+                     Remint sig {} confirmed on-chain.",
+                    transaction_id, e, signature
                 );
             }
         }
@@ -559,6 +591,7 @@ mod tests {
                 pending_remint_deadline_at: Some(now),
                 finality_check_attempts: attempts,
                 recovery_requeue_attempts: 0,
+                landed_remint_signature: None,
             });
     }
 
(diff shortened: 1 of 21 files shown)


indexer/src/storage/postgres/db.rs
```diff
@@ -1143,6 +1165,40 @@ impl PostgresDb {
...
+        .bind(transaction_id)
+        .bind(TransactionStatus::FailedReminted)
+        .bind(remint_signature)
+        .execute(&self.pool)
+        .await?;
+
+        if result.rows_affected() == 0 {
+            return Err(sqlx::Error::RowNotFound);
+        }
+
```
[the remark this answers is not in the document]
sebatustra: Fixed. On the 0-row path it now does one SELECT status::text and logs the two cases  distinctly: warn! "not found" (bug) vs info! "not pending_remint (status X).

indexer/src/operator/sender/remint.rs
```diff
@@ -125,34 +127,64 @@ pub async fn execute_deferred_remint(
...
+            };
+
+            // Durably record the landed remint before the async channel send.
+            // This flips status to FailedReminted now, so a crash in the window
+            // before the writer runs can no longer leave the row PendingRemint
+            // for restart recovery to pick up and remint a second time.
+            //
+            // If this write fails we do not abort: the channel send below still
+            // drives the row to FailedReminted (its UPDATE accepts pending_remint),
+            // which is enough to stop replay. Only landed_remint_signature is lost.
```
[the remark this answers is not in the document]
sebatustra: Fixed, reworded

indexer/src/operator/sender/remint.rs
```diff
@@ -70,11 +70,13 @@ async fn attempt_remint(
             return Ok(existing_signature);
         }
```
Huzaifa696, line 71: Isn't the pagination risk still present in attempt_remint's find_existing_mint_signature_with_memo call? e.g

  - a remint is sent, then the operator crashes before record_remint_result persists (so the row stays pending_remint)
  - the remint lands on-chain during the crash
  - before recovery runs, the recipient ATA accrues >1000 signatures
  - on restart, the row re-hydrates and attempt_remint's lookup misses the landed remint -> it re-sends -> double-mint

indexer/src/operator/sender/remint.rs
```diff
@@ -125,34 +127,65 @@ pub async fn execute_deferred_remint(
...
-                    "Remint succeeded (sig: {}) but no transaction_id to record status",
+                    "Remint confirmed (sig: {}) but entry has no transaction_id; \
+                     cannot record FailedReminted",
                     signature
                 );
+                return;
+            };
+
+            // Durably record the landed remint before the async channel send.
+            // This flips status to FailedReminted now, so a crash in the window
```
Huzaifa696, line 145: Is `FailedReminted` a terminal state? if so then it should be handled like or converted to a `ManualReview` and a runbook should exist for this recovery.

indexer/src/operator/sender/remint.rs
```diff
@@ -125,34 +127,65 @@ pub async fn execute_deferred_remint(
...
+            // before the writer runs can no longer leave the row PendingRemint
+            // for restart recovery to pick up and remint a second time.
+            //
+            // If this write fails we do not abort: the channel send below still
+            // drives the row to FailedReminted (its UPDATE accepts pending_remint),
+            // which is enough to stop replay. Only landed_remint_signature is lost.
```
Huzaifa696, line 151: will the landed_remint_signature matter for the runbook recovery? I am referring to "Only landed_remint_signature is lost."

indexer/src/operator/sender/remint.rs
```diff
@@ -125,34 +127,65 @@ pub async fn execute_deferred_remint(
...
+                    processed_at: Some(Utc::now()),
+                    error_message: Some(entry.original_error.clone()),
+                    remint_signature: Some(signature.to_string()),
+                    remint_attempted: true,
+                },
+                "transaction status update",
+            )
+            .await
+            {
+                error!(
```
Huzaifa696, line 184: what happens if the fallback also fails, do we miss the status update? should we fail close it?

indexer/src/storage/postgres/db.rs
```diff
@@ -671,7 +688,7 @@ impl PostgresDb {
             r#"
             SELECT
                 {}, {}, {}, {}, {}, {}, {}, {}, {}, {},
-                {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}
+                {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}
```
Huzaifa696, line 691: SELECT column lists multiple cols duplicated across 5 query builders. do you think we should move to one shared column-list const, leave it for a cleanup follow-up or the refactor is not worth it?

Huzaifa696:
Every terminal state in the remint flow (including "idempotency lookup unavailable") should have a corresponding recovery runbook.

If it expands the scope then maybe it could be handled in a follow-up?

indexer/src/operator/sender/remint.rs
sebatustra: SOLA3-8's three recommendations are all implemented — durable transactional record before ack, recovery keys off the pending_remint status  filter (not the scan), and -32601/unavailable lookups fail closed to ManualReview.
The finding is related to the persistence being asynchronous after a remint confirms.

indexer/src/operator/sender/remint.rs
sebatustra: FailedReminted means the failed withdrawal was successfully reminted, why would it need a runbook?

indexer/src/operator/sender/remint.rs
sebatustra: The lost landed_remint_signature only ever happens on a FailedReminted row.
The row is in the success terminal state, no need for a runbook recovery. Could you please explain why do you associate FailedReminted with a runbook?

indexer/src/operator/sender/remint.rs
sebatustra: The remint already confirmed on-chain at this point, and the row  stays PendingRemint, which is the retryable state: restart recovery re-hydrates exactly those rows, the idempotency lookup short-circuits the confirmed remint, and it retries the durable write. So the update isn't lost, just deferred to recovery

indexer/src/storage/postgres/db.rs
sebatustra: I think this refactor is out of scope, this PR focuses on closing sola3-8

indexer/src/operator/sender/remint.rs
Huzaifa696: Oh, okay I see, then no need for ManualReview, I misunderstood as failed-to-remint

indexer/src/operator/sender/remint.rs
Huzaifa696: I think this relates to the above [confusion](https://github.com/solana-foundation/solana-private-channels/pull/144#discussion_r3397095028).

Huzaifa696 (approved):

indexer/src/operator/sender/remint.rs
Huzaifa696: so you think, its a separate issue?

sebatustra: partially yes. now that we synchronously record the successful remint, on restart `find_existing_mint_signature_with_memo` wont be called.

PR #145, opened 2026-06-09
fix(indexer): make checkpoint gap-safe under concurrent backfill (SOLA3-17)

Huzaifa696:
## Summary

Addresses SOLA3-17. With the shipped indexer config (`backfill.enabled = true`, `exit_after_backfill = false`), backfill and the live datasource feed one channel into a slot-agnostic processor, and the checkpoint writer persisted the **max** slot it saw. A live-tip `SlotComplete` arriving while backfill was still draining could advance `indexer_state` past the unfilled historical range; a second interruption then made startup backfill resume from the advanced checkpoint, **permanently skipping** the remaining slots, silently dropping deposits/withdrawals that landed during the outage.

This PR makes the persisted checkpoint **monotonic and gap-safe** under concurrent backfill, and makes slot finalization **per-slot atomic**. Concurrent live-during-backfill ingestion is preserved.

## What's changed

- **Per-slot finalize (`transaction_processor.rs`).** Instructions are buffered in a per-slot map instead of one shared `Vec`, so each `SlotComplete` finalizes only its own slot. A foreign (live-tip) marker can no longer flush another slot's in-flight rows, restoring slot-atomic finalize.
- **Gated-contiguous checkpoint (`checkpoint.rs`).** The writer persists a per-program-type `frontier` and advances it only across contiguous slots. While gated to the backfill range `(from_slot, target]`, live-tip slots above `target` are still written but can't move the checkpoint; once the frontier reaches `target` it hands off to plain max-logic. Ungated (no backfill / no gap) keeps the prior max behavior.
- **Single resolved range, armed before live (`backfill.rs`, `indexer.rs`).** `BackfillService` gains `resolve_range()` and `run_range(from, to)` (`run()` is now a thin wrapper). `run()` resolves the range once and gates the writer to the **same** bounds backfill fills, arming it before the processor starts so no live slot can move the checkpoint first. `exit_after_backfill` is gated too, so a withheld slot stalls rather than being leapfrogged.
- **Stall observability (`metrics.rs`, `checkpoint.rs`).** New `INDEXER_CHECKPOINT_FRONTIER_LAG` gauge (saturating) plus a single `warn!` after ~15s of a gated frontier not advancing, signalling a hole or blocked backfill. (`INDEXER_BACKFILL_SLOTS_REMAINING` tracks the fill cursor, not the durable frontier, so it misses a frontier frozen on a hole.)

## Design decisions

- **Contiguity is scoped to the backfill range, not all slot numbers.** Yellowstone emits `SlotComplete` only for slots that produced a block, so Solana-skipped slot numbers never arrive, and a global contiguous watermark would freeze at the first skip. Backfill's own stream *is* contiguous over its range, so the gate enforces contiguity there and hands off to max-logic above `target`, where skipped slots are vacuously covered as today.
- **Data is never gated, only the checkpoint.** Live rows are written immediately (idempotent); only checkpoint advancement waits for the gap to close, so memory stays bounded (≈`max_gap_slots` u64s) and live deposits are visible to the operator at once. On restart, backfill idempotently re-touches them.
- **Fail-closed.** A withheld slot (DB write failed, so no checkpoint update) stalls the frontier below the hole rather than letting a later slot leapfrog it; the next restart re-backfills it. Handoff keys on `frontier == target`, never on "backfill task returned Ok," so a partial/failed backfill cannot release the gate.

## Tests

- **Gated checkpoint (unit):** drops live-tip slots above `target`; folds the full range to `target` and then hands off to max; **stalls on a hole without leapfrogging**; folds out-of-order arrivals; seeds the frontier from `from_slot` (respects `start_slot`); ungated path is byte-for-byte the prior max behavior; lag gauge saturates post-handoff.
- **Per-slot atomicity (unit):** two interleaved slots finalize independently; a foreign `SlotComplete` between a same-slot AllowMint and Deposit does not split the finalize (the deposit is still withheld on mint-status failure).
- **Channel-level integration:** the auditor's exact scenario (`deposit@103` + live `SlotComplete{1_000_000}` interleaved with backfill `101..=105`) asserts the checkpoint ends at `105`, never the tip, and the deposit row survives; a mid-backfill "crash" persists the frontier (`102`), not the tip, so resume skips nothing.

### Coverage Report

| Core | 10,965 | 12,563 | 87.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27351062579) |
| Indexer | 19,697 | 22,644 | 87.0% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27351062579) |
| Gateway | 994 | 1,164 | 85.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27351062579) |
| Auth | 717 | 831 | 86.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27351062579) |
| E2E Integration | 11,421 | 14,180 | 80.5% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27351062579) |
| **Total** | **43,794** | **51,382** | **85.2%** | |

> Last updated: 2026-06-11 14:16:31 UTC by E2E Integration

### DIFF
--- a/indexer/src/indexer/backfill.rs
+++ b/indexer/src/indexer/backfill.rs
@@ -219,9 +219,15 @@ impl BackfillService {
         }
     }
 
-    /// Run the backfill process
-    /// Returns Ok(()) if no gap or backfill successful, Err if gap too large or backfill failed
-    pub async fn run(&self, instruction_tx: InstructionSender) -> Result<(), IndexerError> {
+    /// Work out which slots backfill needs to fill: `Some((from_slot, target))`, or
+    /// `None` if there's no gap. `from_slot` is exclusive (the last durable
+    /// checkpoint) and `target` is inclusive, so the range to fill is
+    /// `(from_slot, target]` — derived from the stored checkpoint, the configured
+    /// `start_slot`, the current chain tip, and the max gap size.
+    ///
+    /// The caller resolves the range once and uses it for two things — gating the
+    /// checkpoint writer and driving the fill — so both see the exact same bounds.
+    pub async fn resolve_range(&self) -> Result<Option<(u64, u64)>, IndexerError> {
         info!(
             "Checking for gaps in indexed data for {:?}...",
             self.program_type
@@ -270,20 +276,29 @@ impl BackfillService {
                     "No gap detected for {:?}. Current slot: {}, From slot: {}",
                     self.program_type, current_slot, from_slot
                 );
-                return Ok(());
+                Ok(None)
             }
             Some(gap) => {
                 info!(
                     "Gap detected for {:?}: {} slots (from {} to {}). Starting backfill...",
                     self.program_type, gap, from_slot, current_slot
                 );
+                Ok(Some((from_slot, current_slot)))
             }
         }
+    }
 
+    /// Fill the resolved range `(from_slot, to_slot]` over the instruction channel.
+    pub async fn run_range(
+        &self,
+        from_slot: u64,
+        to_slot: u64,
+        instruction_tx: InstructionSender,
+    ) -> Result<(), IndexerError> {
         fill_slot_range(
             &self.rpc_poller,
             from_slot,
-            current_slot,
+            to_slot,
             self.config.batch_size,
             self.program_type,
             self.escrow_instance_id,
@@ -293,6 +308,15 @@ impl BackfillService {
 
         info!("Backfill complete for {:?}", self.program_type);
         Ok(())
+    }
+
+    /// Run the backfill process
+    /// Returns Ok(()) if no gap or backfill successful, Err if gap too large or backfill failed
+    pub async fn run(&self, instruction_tx: InstructionSender) -> Result<(), IndexerError> {
+        match self.resolve_range().await? {
+            None => Ok(()),
+            Some((from_slot, to_slot)) => self.run_range(from_slot, to_slot, instruction_tx).await,
+        }
     }
 }
 
--- a/indexer/src/indexer/backfill.rs
+++ b/indexer/src/indexer/backfill.rs
@@ -262,11 +262,32 @@ impl BackfillService {
             last_checkpoint
         };
 
-        let current_slot = self
-            .rpc_poller
-            .get_latest_slot()
-            .await
-            .map_err(|e| BackfillError::SlotFetchFailed { slot: 0, source: e })?;
+        // Retry transient RPC failures with backoff (same policy as block fetches), so a
+        // single hiccup gating the checkpoint writer doesn't force an ungated fallback.
+        let mut retry_count = 0;
+        let current_slot = loop {
+            match self.rpc_poller.get_latest_slot().await {
+                Ok(slot) => break slot,
+                Err(e) => {
+                    retry_count += 1;
+                    if retry_count >= BACKFILL_MAX_RETRIES {
+                        error!(
+                            "Failed to fetch latest slot after {} retries: {}",
+                            BACKFILL_MAX_RETRIES, e
+                        );
+                        return Err(BackfillError::SlotFetchFailed { slot: 0, source: e }.into());
+                    }
+                    warn!(
+                        "Retry {}/{} fetching latest slot after error: {}",
+                        retry_count, BACKFILL_MAX_RETRIES, e
+                    );
+                    tokio::time::sleep(Duration::from_millis(
+                        BACKFILL_RETRY_DELAY_MS * retry_count as u64,
+                    ))
+                    .await;
+                }
+            }
+        };
 
         match validate_gap(current_slot, from_slot, self.config.max_gap_slots)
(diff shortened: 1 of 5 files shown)


indexer/src/indexer/checkpoint.rs
```diff
@@ -109,42 +218,75 @@ impl CheckpointWriter {
...
+        for (&program_type, state) in states.iter_mut() {
+            metrics::INDEXER_CHECKPOINT_FRONTIER_LAG
+                .with_label_values(&[program_type.as_label()])
+                .set(state.lag() as f64);
+            if state.gate.is_none() || state.lag() == 0 || state.dirty {
+                state.stalled_ticks = 0;
+                continue;
+            }
+            state.stalled_ticks += 1;
+            if state.stalled_ticks == STALL_WARN_TICKS {
```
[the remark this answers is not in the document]
Huzaifa696: changed `== STALL_WARN_TICKS` to `% STALL_WARN_TICKS == 0`, so it re-fires every ~15s for the duration of the stall instead of a single line.

indexer/src/indexer/checkpoint.rs
```diff
@@ -109,42 +218,75 @@ impl CheckpointWriter {
...
             }
         }
 
-        // Only remove successfully written checkpoints
-        for program_type in success {
-            pending.remove(&program_type);
-        }
-
         Ok(())
     }
```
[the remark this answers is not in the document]
Huzaifa696: changed the signature to `-> ()` and removed the dead `if let Err` handling at the three call sites.

indexer/src/indexer/transaction_processor.rs
```diff
@@ -30,11 +31,9 @@ use tracing::{debug, error, info, warn};
 pub struct TransactionProcessor {
     storage: Arc<Storage>,
     checkpoint_tx: mpsc::Sender<CheckpointUpdate>,
-    current_slot: Option<u64>,
-    current_program_type: Option<ProgramType>,
 
-    // Buffer all instructions from current slot for batch processing
-    current_slot_instructions: Vec<InstructionWithMetadata>,
+    // Per-slot instruction buffers, so a foreign SlotComplete finalizes only its own slot's rows.
+    slot_buffers: HashMap<u64, Vec<InstructionWithMetadata>>,
```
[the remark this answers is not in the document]
Huzaifa696: It self-heals via reconnect gap-fill in steady state, and safe eviction must not drop un-checkpointed rows

indexer/src/indexer/backfill.rs
```diff
@@ -219,9 +219,15 @@ impl BackfillService {
         }
     }
 
-    /// Run the backfill process
-    /// Returns Ok(()) if no gap or backfill successful, Err if gap too large or backfill failed
-    pub async fn run(&self, instruction_tx: InstructionSender) -> Result<(), IndexerError> {
+    /// Work out which slots backfill needs to fill: `Some((from_slot, target))`, or
+    /// `None` if there's no gap. `from_slot` is exclusive (the last durable
```
sebatustra, line 223: why is from_slot exclusive? is it not possible that the indexer crashed mid slot?

indexer/src/indexer/indexer.rs
```diff
@@ -138,19 +151,41 @@ pub async fn run(
...
-                // Run backfill concurrently with live indexing
-                let instruction_tx_clone = instruction_tx.clone();
-                tokio::spawn(async move {
-                    if let Err(e) = backfill_service.run(instruction_tx_clone).await {
-                        error!("Backfill failed: {}", e);
-                    } else {
-                        info!("Backfill completed successfully");
+                // Gate the writer to the exact range backfill will fill; on resolve
+                // failure fall back to an ungated writer (today's behavior) rather
+                // than aborting startup, since the live datasource still runs.
```
sebatustra: > fall back to an ungated writer (today's behavior)

should a transient resolve error retry with backoff instead of falling back to ungated-no-backfill? if for example `getSlot` fails in `resolve_range` we go back to the unprotected pre fix behaviour

indexer/src/indexer/backfill.rs
Huzaifa696: `from_slot = last_checkpoint`  and checkpoint N means slot N is fully committed; a mid-slot crash leaves checkpoint at N-1, so the partial slot is re-fetched and idempotently re-upserted

indexer/src/indexer/indexer.rs
Huzaifa696: I've added bounded backoff retries around `resolve_range` and changed the live branch to fail closed i.e propagate the error and refuse to start.

sebatustra (approved):

PR #146, opened 2026-06-09
fix(core): bound the write-pipeline hot-path queues and shed at public ingress (SOLA2-8)

Huzaifa696:
## Summary

Closes SOLA2-8. The public, un-authed `sendTransaction` path fed three **unbounded** in-memory channels in the write pipeline. Because the runtime is gasless, an attacker needs no prefunded accounts to keep producing valid, allowlisted transactions; a sustained stream that outpaces drain throughput grows the heap without limit until the node OOMs, losing all accepted-but-unsettled in-memory transactions.

This PR bounds all three attacker-reachable channels, turns the public ingress into a fail-fast load-shedder, and lets backpressure chain upstream from the slowest stage back to RPC admission. The two settler-to-upstream **feedback** channels stay unbounded on purpose (bounding them would close a deadlock cycle).

- **RPC -> dedup**: bounded; a full queue **sheds**. `sendTransaction` returns a new retryable `-32003` (`NODE_AT_CAPACITY_CODE`), logs `warn!` with the signature, and increments `rpc_ingress_shed_total`. Fail-fast (not `send().await`) so a memory-DoS can't become an RPC-connection-exhaustion DoS, and so an honest client can resubmit.
- **sigverify -> sequencer**: bounded; the worker's forward becomes `send().await` raced against the shutdown token (backpressure without wedging shutdown).
- **executor -> settler**: bounded; the executor's two result sends become `send().await` raced against shutdown. The sends carry owned values after `execute_batch` returns, with no `RwLock` guard held across the await.
- Three new `usize` config knobs (CLI flag + env var) sized so steady state never sheds.

## Design decisions

- **Shed, don't await, at ingress.** A full queue returns a distinct retryable `-32003` (not `INVALID_PARAMS`; the tx is valid), not `send().await`, so a memory-DoS can't turn into RPC-connection exhaustion and a returned signature never falsely means "accepted".
- **Bound + blocking backpressure.** sigverify and executor block on a full downstream instead of dumping into an unbounded queue, so the unboundedness can't relocate to `sequencer_rx` or the settler's `processing_results`. Both sends race the shutdown token so shutdown stays prompt.
- **Feedback edges stay unbounded (deliberate).** `settled_accounts_tx` and `settled_blockhashes_tx` must stay unbounded; bounding an edge a blocked upstream stage still consumes would deadlock.
- **Config + one metric.** Three `usize` capacities (CLI flag + `PRIVATE_CHANNEL_*` env); `run_node` rejects `0` (fail closed) so a misconfiguration can't panic `mpsc::channel(0)`. One counter `rpc_ingress_shed_total`; no gauges (`/health` heartbeats + RSS already show saturation).

## Why it's safe

- **No deadlock.** Backpressure is strictly upstream; the only downstream-to-upstream edges are the unbounded feedback channels, so the settler always drains `execution_results_rx` even while the executor is blocked.
- **Blockhash freshness.** The settler is tick-driven, not recv-driven, so it keeps producing blocks under backpressure and `getLatestBlockhash` stays fresh.
- **Feedback serviced.** dedup drains `settled_blockhashes_rx` while blocked on its output send; `settled_accounts_rx` drives only cache eviction (lazy, via `BOB::garbage_collect`), never execution correctness.
- **Health non-regression.** sigverify and executor record progress before the blocking send, so a backpressured-but-advancing stage reads healthy; only a genuine >5s stall (DB hang) trips `/health`.
- **Shed before dedup.** A shed tx never enters the dedup cache, so the client may resubmit the identical tx and it is accepted.

## Testing

- **Ingress (unit):** `ingress_sheds_when_full` (full queue returns `-32003`, receiver never exceeds capacity), `ingress_shed_increments_metric` (via the Prometheus registry), `shed_tx_can_be_resubmitted` (shed, drain, then the identical tx is accepted), accept-path tests migrated to the bounded helper.
- **sigverify (unit):** `sigverify_blocks_not_drops_when_sequencer_full` (conservation: all txs arrive, none dropped; queue never exceeds cap), `sigverify_worker_exits_on_shutdown_with_full_sequencer`.
- **executor (unit, real Postgres):** `executor_blocks_then_completes_on_full_results` (blocked send completes once the receiver drains, no panic/lock-poison), `executor_exits_on_shutdown_with_full_results`.
- **Cross-cutting (unit):** `health_ok_under_sustained_backpressure` (heartbeat-level: backpressure-with-progress stays healthy).
- **Config guard (unit):** `test_run_node_rejects_zero_queue_capacity` (a `0` capacity fails closed with a clear error instead of panicking `mpsc::channel(0)`).
- **Real-node backpressure (integration, CI, real Postgres):** `health_ok_under_sustained_backpressure` (`/health` stays 200 past the 5s heartbeat margin) and `no_deadlock_under_slow_settler` (slots keep advancing under a 2s blocktime). ~15s combined; wired into the integration Makefile private-channel group (plus its coverage target).
- **Staging load (`#[ignore]`):** `bounded_memory_under_burst` (RSS plateaus, shed counter climbs); the RSS-plateau assertion is resource-sensitive, so it runs in staging, not CI (`-- --ignored`).
- **Migration-only:** all remaining `mpsc::unbounded_channel()` test constructors for the now-bounded channels migrated to `mpsc::channel(const)`, bound to the same named config consts (no drifting literals). Feedback-channel test constructors stay unbounded.
- Full `private-channel-core` suite: 385 passed, 1 `#[ignore]`d.

## Scope

**Out of scope:** gateway per-IP rate limiting / `sendTransaction` auth; improving the heartbeat to distinguish "backpressured" from "wedged".

### Coverage Report

| Core | 10,926 | 12,524 | 87.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27255363699) |
| Indexer | 18,753 | 21,734 | 86.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27255363699) |
| Gateway | 994 | 1,164 | 85.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27255363699) |
| Auth | 717 | 831 | 86.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27255363699) |
| Withdraw Program | 120 | 232 | 51.7% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27255363643) |
| Escrow Program | 1,185 | 1,963 | 60.4% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27255363643) |
| DvP Swap Program | 247 | 1,112 | 22.2% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27255363643) |
| E2E Integration | 11,240 | 13,912 | 80.8% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27255363699) |
| **Total** | **43,935** | **52,360** | **83.9%** | |

> Last updated: 2026-06-10 06:01:59 UTC by E2E Integration

### DIFF
--- a/core/src/nodes/node.rs
+++ b/core/src/nodes/node.rs
@@ -27,6 +27,13 @@ use {
     tracing::{error, info, warn},
 };
 
+/// RPC→dedup ingress queue capacity. Sized so steady state never sheds.
+pub const DEFAULT_INGRESS_QUEUE_CAPACITY: usize = 10_000;
+/// sigverify→sequencer queue capacity (mirrors the sigverify queue size).
+pub const DEFAULT_SEQUENCER_QUEUE_CAPACITY: usize = 1000;
+/// executor→settler results queue capacity.
+pub const DEFAULT_EXECUTION_RESULTS_CAPACITY: usize = 1000;
+
 #[derive(Debug, Clone, PartialEq, clap::ValueEnum)]
 pub enum NodeMode {
     /// Read-only node - serves read RPCs only
@@ -47,6 +54,12 @@ pub struct NodeConfig {
     pub max_tx_per_batch: usize,
     pub batch_deadline_ms: u64,
     pub batch_channel_capacity: usize,
+    /// RPC→dedup ingress queue capacity; a full queue sheds load at admission.
+    pub ingress_queue_capacity: usize,
+    /// sigverify→sequencer queue capacity; full applies upstream backpressure.
+    pub sequencer_queue_capacity: usize,
+    /// executor→settler results queue capacity; full applies upstream backpressure.
+    pub execution_results_capacity: usize,
     /// Max parallel SVM worker threads per batch (including the calling thread).
     /// Set to 1 to disable intra-batch parallelism entirely. Effective only for
     /// batches ≥ `MIN_PARALLEL_BATCH_SIZE`; smaller batches always run sequentially.
@@ -78,6 +91,9 @@ impl Default for NodeConfig {
             max_tx_per_batch: 64,
             batch_deadline_ms: 10,
             batch_channel_capacity: 16,
+            ingress_queue_capacity: DEFAULT_INGRESS_QUEUE_CAPACITY,
+            sequencer_queue_capacity: DEFAULT_SEQUENCER_QUEUE_CAPACITY,
+            execution_results_capacity: DEFAULT_EXECUTION_RESULTS_CAPACITY,
             max_svm_workers: 8,
             accountsdb_connection_url: "postgresql://user:password@localhost:5432/private_channel"
                 .to_string(),
@@ -120,6 +136,21 @@ pub async fn run_node(config: NodeConfig) -> Result<NodeHandles, Box<dyn std::er
             "transaction_expiration_ms must be >= blocktime_ms (max_blockhashes would be 0)".into(),
         );
     }
+    // Zero capacity would panic the bounded-channel constructors below; fail closed instead.
+    if matches!(config.mode, NodeMode::Write | NodeMode::Aio) {
+        for (name, cap) in [
+            ("ingress_queue_capacity", config.ingress_queue_capacity),
+            ("sequencer_queue_capacity", config.sequencer_queue_capacity),
+            (
+                "execution_results_capacity",
+                config.execution_results_capacity,
+            ),
+        ] {
+            if cap == 0 {
+                return Err(format!("{name} must be greater than 0").into());
+            }
+        }
+    }
 
     // Create a single shutdown token for all services
     let shutdown_token = CancellationToken::new();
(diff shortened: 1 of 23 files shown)


core/src/rpc/send_transaction_impl.rs
```diff
@@ -104,41 +105,149 @@ pub async fn send_transaction_impl(
...
+
+        let err = send_transaction_impl(&deps, encode_tx(&spl_tx()), None)
+            .await
+            .expect_err("a full ingress queue must shed");
+        assert_eq!(err.code(), NODE_AT_CAPACITY_CODE);
+        assert_eq!(
+            rx.max_capacity(),
+            TEST_INGRESS_CAP,
+            "receiver must never exceed the bounded capacity"
+        );
```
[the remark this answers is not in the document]
Huzaifa696: Replaced it with `assert_eq!(rx.len(), TEST_INGRESS_CAP)`, which actually proves the queue buffered every accepted tx and is full at the shed point.

core/src/stages/sigverify.rs
```diff
@@ -700,24 +714,31 @@ mod tests {
...
+        // Let the pipeline saturate against the paused consumer, then assert the
+        // sequencer side holds no more than cap + workers (each worker can park
+        // at most one send beyond the queue).
+        tokio::time::sleep(std::time::Duration::from_millis(200)).await;
+        assert!(
+            sequencer_rx.len() <= sequencer_cap,
+            "sequencer queue must never exceed its capacity ({} > {})",
+            sequencer_rx.len(),
+            sequencer_cap
+        );
```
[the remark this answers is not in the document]
Huzaifa696: Removed the len check and rely on the conservation assertion (the real coverage).

integration/tests/private-channel/test_ingress_backpressure.rs
```diff
@@ -0,0 +1,241 @@
...
+fn rss_kb() -> u64 {
+    // statm field 2 (resident pages) × page size, in KiB. Linux-only.
+    let statm = std::fs::read_to_string("/proc/self/statm").unwrap_or_default();
+    let resident_pages: u64 = statm
+        .split_whitespace()
+        .nth(1)
+        .and_then(|s| s.parse().ok())
+        .unwrap_or(0);
+    let page_kb = 4; // 4 KiB pages on x86-64
+    resident_pages * page_kb
```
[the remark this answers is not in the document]
Huzaifa696: Fixed, `rss_kb()` now reads `VmRSS` from `/proc/self/status` (already in kB), so it's page-size agnostic.

core/src/nodes/node.rs
```diff
@@ -27,6 +27,13 @@ use {
     tracing::{error, info, warn},
 };
 
+/// RPC→dedup ingress queue capacity. Sized so steady state never sheds.
```
sebatustra, line 30: how were these default values picked? is it possible to provide recommended values depending on the machine specs?

core/src/stages/sigverify.rs
```diff
@@ -700,24 +714,25 @@ mod tests {
...
-        let drained_clone = Arc::clone(&drained);
-        let drainer = tokio::spawn(async move {
-            while sequencer_rx.recv().await.is_some() {
-                drained_clone.fetch_add(1, Ordering::Relaxed);
-                if drained_clone.load(Ordering::Relaxed) >= total_txs {
-                    break;
-                }
+        // Let the workers actually park on the full queue so the backpressure
+        // path is exercised before we drain; the conservation check below is the
+        // real regression guard (a drop-instead-of-park bug fails it).
```
sebatustra, line 719: This may be a dumb question: how and where are txs parked? and how does this not eventually lead to OOM?

core/src/nodes/node.rs
Huzaifa696: They're sized so steady state never sheds while keeping a hard, tiny memory cap - each queued item is one `SanitizedTransaction` (<=1232 B wire, a few KB in memory), so even 10k ingress is only tens of MB; memory isn't the binding constraint, so there's no meaningful per-machine formula. 

The real deciding factor is burst-absorption vs. latency (traffic-shaped), so we exposed them as env vars to tune empirically against the metric `rpc_ingress_shed_total` - added a "Tuning" note to `docs/CONFIG.md`.

core/src/stages/sigverify.rs
Huzaifa696: A worker that hits a full sequencer queue suspends at `tx.send(transaction).await`, holding that one tx in its task - so "parked" = at most one tx per worker (default 4), not a buffer. 

Parking is what *prevents* OOM: a blocked worker stops pulling input -> the bounded sigverify input fills -> dedup blocks -> the bounded RPC->dedup queue fills -> ingress `try_send` sheds (`-32003`). Total in-memory txs <= sum of the bounded capacities + one parked send per stage; the old unbounded queue is what let it grow without limit.

sebatustra (approved):

PR #147, opened 2026-06-09
fix(escrow): make ResetSmtRoot idempotent (SOLA3-4)

sebatustra:
fix(escrow): make ResetSmtRoot idempotent (SOLA3-4)

Summary

ResetSmtRoot was retried as idempotent on confirmation timeout, but the on-chain instruction was
not: every success unconditionally bumped current_tree_index. An ambiguously-confirmed boundary
rotation could land twice, advancing the escrow two generations while the operator advanced one —
stranding every withdrawal in the skipped generation. (SOLA3-4)

Changes

On-chain (escrow program)
- ResetSmtRoot now carries expected_current_tree_index: u64 and rejects with a new
UnexpectedTreeIndex error when it no longer matches the instance. A replayed reset can no longer
advance the tree twice.

Operator
- Stamps the reset with its local tree index when building the instruction.
- Self-heals on UnexpectedTreeIndex: re-fetches the authoritative on-chain index and syncs local SMT
(fail-closed if the fetch fails), so the boundary withdrawal proceeds instead of stalling.

Testing

- New escrow integration test: a replay with a stale expected index is rejected and
current_tree_index stays put.
- New operator unit tests: self-heal happy path (re-sync to chain), fail-closed path (left unchanged), and Custom(13) decode.
### Coverage Report

| Core | 10,965 | 12,563 | 87.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27361398682) |
| Indexer | 19,605 | 22,657 | 86.5% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27361398682) |
| Gateway | 994 | 1,164 | 85.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27361398682) |
| Auth | 717 | 831 | 86.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27361398682) |
| Withdraw Program | 120 | 232 | 51.7% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27361398699) |
| Escrow Program | 1,186 | 1,973 | 60.1% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27361398699) |
| DvP Swap Program | 247 | 1,112 | 22.2% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27361398699) |
| E2E Integration | 11,385 | 14,186 | 80.3% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27361398682) |
| **Total** | **44,972** | **53,606** | **83.9%** | |

> Last updated: 2026-06-11 16:45:02 UTC by DvP Swap Program

### DIFF
--- a/indexer/src/operator/sender/transaction.rs
+++ b/indexer/src/operator/sender/transaction.rs
@@ -110,12 +110,19 @@ impl SenderState {
                     compute_budget,
                 })
             }
-            TransactionBuilder::ResetSmtRoot(ref builder) => {
-                let in_flight_count = self
+            TransactionBuilder::ResetSmtRoot(mut builder) => {
+                // Bind the reset to our local tree index so the on-chain program
+                // rejects a replay. Initialize SMT state first in case a reset is
+                // the first thing we process after a restart.
+                if self.smt_state.is_none() {
+                    self.initialize_smt_state().await?;
+                }
+                let smt = self
                     .smt_state
                     .as_ref()
-                    .map(|s| s.nonce_to_builder.len())
-                    .unwrap_or(0);
+                    .ok_or(ProgramError::SmtNotInitialized)?;
+                let in_flight_count = smt.nonce_to_builder.len();
+                let expected_current_tree_index = smt.smt_state.tree_index();
 
                 if in_flight_count > 0 {
                     info!(
@@ -123,14 +130,15 @@ impl SenderState {
                         in_flight_count
                     );
 
-                    self.pending_rotation = Some(builder.clone());
+                    self.pending_rotation = Some(builder);
 
                     return Err(ProgramError::RotationPending { in_flight_count }.into());
                 }
 
                 // No in-flight transactions - process immediately
+                builder.expected_current_tree_index(expected_current_tree_index);
                 Ok(InstructionWithSigners {
-                    instructions: tx_builder.instructions()?,
+                    instructions: vec![builder.instruction()],
                     fee_payer,
                     signers,
                     compute_budget,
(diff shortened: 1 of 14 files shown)


indexer/src/operator/sender/transaction.rs
```diff
@@ -539,6 +547,27 @@ pub(super) fn handle_confirmation_result<'a>(
...
+                // Rejected because a reset already advanced the tree on-chain. Sync
+                // local SMT to the authoritative index. On fetch failure, leave it
+                // unchanged rather than guess; a restart re-syncs from chain.
+                match state.fetch_onchain_tree_index().await {
+                    Ok(idx) => {
+                        if let Some(ref mut smt_state) = state.smt_state {
+                            smt_state.smt_state.reset(idx);
+                        }
+                        warn!("ResetSmtRoot rejected - synced local SMT to on-chain tree_index {idx}");
+                    }
```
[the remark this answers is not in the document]
sebatustra: fixed

indexer/src/operator/sender/transaction.rs
```diff
@@ -539,6 +547,29 @@ pub(super) fn handle_confirmation_result<'a>(
...
                 }
             },
+            Ok(ConfirmationResult::Failed(Some(
+                PrivateChannelEscrowProgramError::UnexpectedTreeIndex,
+            ))) => {
+                metrics::OPERATOR_TRANSACTION_ERRORS
+                    .with_label_values(&[pt, "reset_tree_already_advanced"])
+                    .inc();
+                // Rejected because a reset already advanced the tree on-chain. Sync
+                // local SMT to the authoritative index. On fetch failure, leave it
```
Huzaifa696, line 620: do you mean that if there is any RPC blip then after a retry of same transaction, this branch should succeed?

and is there a case where this could happen in steady state?
"warn!("ResetSmtRoot rejected - on-chain tree_index is {idx} but smt_state is None, sync skipped");"
i think this is only a possibility at startup?

Huzaifa696:
should we add the new arg in the admin-ui?
```ts
const instruction = await getResetSmtRootInstructionAsync({
  payer: transactionSigner,
  operator: transactionSigner,
  instance: address(instancePubkey),
  // expectedCurrentTreeIndex missing
});
```

and in the fuzz testing harness?
```rust
let ix = ResetSmtRootBuilder::new()
    .payer(self.trident.payer().pubkey())
    .operator(operator)
    .instance(instance)
    .operator_pda(operator_pda)
    .event_authority(event_authority)
    .private_channel_escrow_program(PRIVATE_CHANNEL_ESCROW_PROGRAM_ID)
    .instruction();                       // expected_current_tree_index never set
```

indexer/src/operator/sender/transaction.rs
sebatustra: > do you mean that if there is any RPC blip then after a retry of same transaction, this branch should succeed?

No, the retry is intentionally rejected with UnexpectedTreeIndex; letting it land is the double-rotation bug. The handler reconciles by fast-forwarding local SMT to the authoritative on-chain index.

> and is there a case where this could happen in steady state?

you are right, good catch, that branch is unreachable. removed.

Huzaifa696 (approved):

PR #148, opened 2026-06-10
fix(indexer): key transaction identity on (signature, instruction_index) (SOLA2-13)

Huzaifa696:
## Summary

Addresses SOLA2-13. The indexer's durable accounting model assumed one economic event per transaction
signature: the decoders copy the outer transaction signature onto every parsed deposit/withdraw
instruction, the `transactions` table made `signature` globally unique, and the insert loop reused the
first row's id for every later same-signature instruction, so every economic event after the first was **silently dropped**.

This PR makes the durable identity of an indexed event the pair **`(signature, instruction_index)`** (the
instruction's absolute position in its transaction) instead of the signature alone, so every instruction in
a multi-instruction transaction persists, is minted/settled, and is counted in reconciliation.

## What's changed

- **Position carried end to end.** `InstructionWithMetadata` gains `instruction_index`; both decoders set it
  from `enumerate()` over the full top-level instruction list, taken **before** the program-id filter (so it's
  the true, replay-stable position). The processor threads it into deposit and withdrawal rows.
- **Composite durable identity.** `DbTransaction`/`transactions` gain an `instruction_index` column
  (`INTEGER NOT NULL DEFAULT 0`), and the unique key becomes `(signature, instruction_index)`.
- **Idempotent in-place migration (`db.rs`).** Add the column, build the composite unique index
  `idx_transactions_signature_ix` **while old single-signature uniqueness still holds**, then drop the old
  constraint and index. Fresh databases reach the same end state.

## Design decisions

- **Composite key — not a synthetic signature or a reject-multi-instruction guard.** Keeps the real on-chain
  signature for provenance and persists every legitimate event (multi-instruction txs are supported), rather
  than mangling identity or dropping events.
- **Write-path only; downstream is already keyed by row `id`.** Status updates, fetcher locking, mint
  idempotency (`transaction_id`), and the per-row nonce INSERT trigger all key on `id`, so two events from one
  tx settle independently with no downstream change.
- **Migration is build-before-drop.** `signature` is never left unprotected and the column exists before any
  index references it, so an in-place upgrade can't fail at startup.

## Tests

- **Decoder (unit):** `test_multiple_instructions_same_tx` now asserts indices `0/1/2`; a new
  `test_index_is_absolute_position_across_filtered_instructions` proves a filtered-out middle instruction
  still yields indices `0` and `2` (absolute, not relative). Yellowstone has an equivalent two-instruction
  test (indices `0/1`, equal signature).
- **Processor (unit):** a single transaction yielding two in-instance deposits produces two rows, each
  carrying its index.
- **Storage (integration, testcontainers):** two rows with the same signature and indices `0/1` (a deposit
  and a withdrawal) persist as two distinct ids; `COUNT = 2`; indices `[0, 1]`; the withdrawal receives a
  non-null nonce; re-insert is idempotent on the pair. The existing same-signature race test stays valid and
  now also covers the same-index collision.
- **End-to-end pipeline (integration, testcontainers):** `multi_instruction_pipeline.rs` drives two
  same-signature deposits through the real `TransactionProcessor` channel API into Postgres and asserts both
  events survive with distinct indices and that instruction 1's amount/recipient are not overwritten by
  instruction 0; a second test replays the same slot and asserts the composite key stays idempotent (no
  duplicates, no collapse).
- **Migration (integration):** idempotency across double-init asserts the composite index exists and the old
  index/constraint are gone; a new `test_init_schema_upgrades_legacy_signature_only_schema` rewinds the table
  to the pre-change shape and asserts the in-place upgrade succeeds and converges to the composite identity.
### Coverage Report

| Core | 10,965 | 12,563 | 87.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27364331644) |
| Indexer | 19,936 | 22,865 | 87.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27364331644) |
| Gateway | 994 | 1,164 | 85.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27364331644) |
| Auth | 717 | 831 | 86.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27364331644) |
| Withdraw Program | 120 | 232 | 51.7% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27364331535) |
| Escrow Program | 1,186 | 1,973 | 60.1% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27364331535) |
| DvP Swap Program | 247 | 1,112 | 22.2% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27364331535) |
| E2E Integration | 11,437 | 14,241 | 80.3% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27364331644) |
| **Total** | **45,355** | **53,869** | **84.2%** | |

> Last updated: 2026-06-11 17:35:35 UTC by E2E Integration


integration/tests/indexer/multi_instruction_pipeline.rs
```diff
@@ -0,0 +1,260 @@
...
+    handle
+        .await
+        .expect("processor task join")
+        .expect("processor ok");
+}
+
+/// Two `Deposit` instructions sharing one signature must both persist as
+/// distinct rows with their absolute instruction indices — not collapse to one.
+#[tokio::test(flavor = "multi_thread")]
+async fn two_same_signature_deposits_persist_as_distinct_rows() {
```
sebatustra, line 170: can we add a a test with withdrawals as well?

sebatustra (commented):
what would happen in the case of nested instructions (some program CPIs into escrow for example)?

integration/tests/indexer/multi_instruction_pipeline.rs
Huzaifa696: I've added `two_same_signature_withdrawals_persist_as_distinct_rows` driving two `WithdrawFunds` through the processor and asserting two rows (indices 0/1)

Huzaifa696:
> what would happen in the case of nested instructions (some program CPIs into escrow for example)?


Both decoders only enumerate top-level instructions; inner/CPI instructions never make a row. So a foreign program CPI-ing into escrow isn't indexed at all. This pre-existing behavior is unchanged here.

I have created this follow-up [ticket](https://linear.app/solana-fndn/issue/EXO-121/indexer-misses-escrowwithdraw-instructions-invoked-via-cpi) for this?

sebatustra (approved):

PR #153, opened 2026-06-12
fix(indexer): full-range u64 amounts and on-chain-driven startup reconciliation, plus gitignored env secrets (SOLA2-1,    SOLA2-17, SOLA3-7, SOLA3-12)

Huzaifa696:
## Summary

SOLA2-1, SOLA2-17 (= SOLA3-2): every bridge amount was persisted as a signed `i64` / Postgres `BIGINT`
while on-chain amounts are `u64`, so a valid amount above `i64::MAX` wrapped negative in storage. On the
deposit side the operator cast the negative value back to `u64` (the user was credited) but reconciliation
read the poisoned sum and fail-closed every later restart. On the withdrawal side the burn succeeded, the
operator rejected the negative row as a poison pill, and the processor halted the entire withdrawal
pipeline. One valid large transaction was enough to trigger either outage.

SOLA3-7: startup reconciliation derived its mint universe from the `mints` table alone and returned
success when that table was empty. Only `AllowMint` populates `mints`, so a fresh or partially restored DB
against a funded escrow passed the boot-time safety gate without a single on-chain check, and the bridge
resumed blind.

SOLA3-12: `make build-localnet` / `build-devnet` wrote the generated admin private key into `.env.local` /
`.env.devnet`, which are git-tracked. A routine commit after a routine build will leak a live signing key

## What's changed

- Amounts (SOLA2-1, SOLA2-17). The `transactions.amount` column is `NUMERIC(20,0)` and a new
  `TokenAmount(u64)` newtype is the single conversion seam: it encodes through `BigDecimal` and decodes
  with a checked conversion that rejects negative, fractional, and over-`u64` values loudly. The
  `amount as i64` narrowing in the model builder is gone, and all four operator-side
  `u64::try_from` / `as u64` guard sites now read the checked `u64` directly. Reconciliation totals are
  `BigDecimal` (a gross sum of near-`u64::MAX` deposits can exceed `u64`), and the shared
  `net_to_u64` / `NetBalance` helper converts nets without wrapping in both reconciliation paths. The
  streamer reads the column as `amount::text`; its wire format (string amounts) is unchanged.
- Startup reconciliation (SOLA3-7). The operator's `get_token_accounts_by_owner` sweep moved into a shared `escrow_sweep` module, and startup reconciliation now compares the union of DB mints and on-chain escrow balances. A mint present on only one side compares against 0 on the other, so a funded escrow with an empty DB fails closed instead of passing blind; the empty-state pass remains only when both sides are genuinely empty. Recovery for a legitimately fresh DB is `indexer resync`.
- Secrets (SOLA3-12). `update-admin-env.sh` writes the public admin key to the tracked template and the
  private key only to a gitignored runtime env file (default `.env`). The compose env-file chain loads
  `.env` last so runtime secrets override the blanked templates, and the chain drops the flag when `.env`
  does not exist so a fresh clone still starts. The fail-closed secrets check validates the same chain
  compose resolves. README and DEVNET_QUICKSTART now direct secrets to `.env`.

## Tests

- `TokenAmount` unit tests: decode accepts the full `u64` range and rejects negative, fractional, and
  over-`u64` values; serde stays a bare integer. Net conversion tests cover clamp-to-zero and the
  `u64::MAX` mismatch sentinel.
- Postgres (testcontainers): a deposit and a withdrawal of `i64::MAX + 1` round-trip bit-for-bit through
  the raw column and the `FromRow` path; reconciliation sums two above-`i64::MAX` deposits correctly
  (`postgres_db_test` 44, `reconciliation_db_test` 7, `reconciliation_e2e_test` 9, `remint_e2e_test` 6).
- Reconciliation unit tests: union set construction (DB-only, chain-only, merged, disjoint), and
  startup-contract tests against a mocked RPC sweep, including the previously missing case: an empty DB
  with a funded escrow must block startup.
- Secrets: a shell test generates a keypair, runs `update-admin-env.sh`, and asserts the tracked template
  carries only the public key while the runtime file carries the private key.

### Coverage Report

| Core | 10,965 | 12,563 | 87.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27612171290) |
| Indexer | 19,674 | 22,639 | 86.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27612171290) |
| Gateway | 994 | 1,164 | 85.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27612171290) |
| Auth | 717 | 831 | 86.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27612171290) |
| Withdraw Program | 120 | 232 | 51.7% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27612171254) |
| Escrow Program | 1,186 | 1,973 | 60.1% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27612171254) |
| DvP Swap Program | 247 | 1,112 | 22.2% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27612171254) |
| E2E Integration | 11,407 | 14,179 | 80.4% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/27612171290) |
| **Total** | **45,063** | **53,581** | **84.1%** | |

> Last updated: 2026-06-16 11:20:14 UTC by E2E Integration

### DIFF
--- a/indexer/src/indexer/reconciliation.rs
+++ b/indexer/src/indexer/reconciliation.rs
@@ -12,24 +12,25 @@
 //! Only completed withdrawals (`release_funds`) reduce the ATA balance.
 //!
 //! Flow:
-//! 1. Query the DB for per-mint aggregate balances (all deposits − completed withdrawals).
-//! 2. Derive the escrow ATA for each mint (instance PDA + mint + token program).
-//! 3. Fetch the live ATA balance via RPC.
+//! 1. Sweep the escrow instance's on-chain token accounts, summed per mint.
+//! 2. Query the DB for per-mint aggregate balances (all deposits − completed withdrawals).
+//! 3. Compare the union of both mint sets; a mint on only one side compares against 0.
 //! 4. If any |on_chain - db_expected| > threshold → log error, emit alert, abort startup.
 //! 5. If any mismatch ≤ threshold (but > 0) → log warning, continue.
-//! 6. If all balanced → log info, continue.
+//! 6. If all balanced (or both sides empty) → log info, continue.
 
 use crate::{
     config::{ProgramType, ReconciliationConfig},
     error::{IndexerError, ReconciliationError},
-    operator::{rpc_util::RpcClientWithRetry, RetryConfig, RetryPolicy},
+    operator::{
+        escrow_sweep::fetch_escrow_balances_by_mint, rpc_util::RpcClientWithRetry, RetryConfig,
+    },
     storage::common::amount::{net_to_u64, NetBalance},
+    storage::common::models::MintDbBalance,
     storage::Storage,
 };
-use private_channel_core::rpc::error::INVALID_PARAMS_CODE;
 use solana_sdk::{commitment_config::CommitmentConfig, pubkey::Pubkey};
-use spl_associated_token_account::get_associated_token_address_with_program_id;
-use std::str::FromStr;
+use std::collections::{BTreeMap, HashMap};
 use tracing::{error, info, warn};
 
 /// Per-mint result produced during reconciliation.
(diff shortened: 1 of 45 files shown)


scripts/tests/update-admin-env.test.sh
```diff
@@ -0,0 +1,41 @@
...
+  || fail "tracked file missing PRIVATE_CHANNEL_ADMIN_KEYS"
+if grep -qE '^ADMIN_PRIVATE_KEY=.+' "$tracked_env"; then
+  fail "tracked file leaked the admin private key"
+fi
+
+# 2. The runtime file must carry the private key, non-empty.
+admin_priv="$(grep '^ADMIN_PRIVATE_KEY=' "$runtime_env" | tail -n1 | cut -d= -f2-)"
+[[ -n "$admin_priv" ]] || fail "ADMIN_PRIVATE_KEY is empty in runtime file"
+
+echo "PASS: the admin private key stays out of the tracked template"
```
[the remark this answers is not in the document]
Huzaifa696: Not required: the guarantee is structural, not test-enforced (`.env` is gitignored and the tracked templates ship a blank `ADMIN_PRIVATE_KEY`), so a missing CI hook can't reopen the finding.
Also no shell script runs from CI in this repo and this one needs `solana-keygen` on PATH, so it stays a developer-run check (`./scripts/tests/update-admin-env.test.sh`).

indexer/src/indexer/reconciliation.rs
```diff
@@ -171,82 +181,37 @@ async fn log_orphan_deposit_rows_at_startup(storage: &Storage) {
...
+                "Expected escrow balance exceeds u64::MAX; flagging as a reconciliation mismatch"
+            );
+            u64::MAX
         }
-        let msg = message.to_lowercase();
-        msg.contains("could not find account") || msg.contains("account not found")
-    } else {
-        false
     }
 }
```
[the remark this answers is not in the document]
Huzaifa696: Fixed. The overflow case now returns `Err(ReconciliationError::DbBalanceOverflow)`

indexer/src/indexer/reconciliation.rs
```diff
@@ -12,31 +12,36 @@
...
 use tracing::{error, info, warn};
 
 /// Per-mint result produced during reconciliation.
 #[derive(Debug, Clone)]
 pub struct MintReconciliation {
     pub mint: String,
     /// Expected balance according to DB: all indexed deposits − completed withdrawals.
-    pub db_expected: i64,
+    /// Unsigned because it mirrors the escrow ATA balance, itself a u64; a negative
+    /// net is clamped to 0 at the call site so this value stays lossless across the
```
sebatustra, line 42: under which scenario can there be a negative net?

indexer/src/indexer/reconciliation.rs
```diff
@@ -93,57 +98,62 @@ pub async fn run_startup_reconciliation(
...
+    db_balances: &[MintDbBalance],
+    on_chain_balances: &HashMap<Pubkey, u64>,
+) -> Result<Vec<MintReconciliation>, ReconciliationError> {
+    // Keyed by mint string so the DB side (String addresses) and on-chain side
+    // (Pubkey) merge into one universe; BTreeMap keeps the order deterministic.
+    let mut by_mint: BTreeMap<String, (u64, u64)> = BTreeMap::new();
 
-        let db_expected = balance.total_deposits - balance.total_withdrawals;
+    for balance in db_balances {
+        let net = &balance.total_deposits - &balance.total_withdrawals;
```
sebatustra, line 142: should we change to `checked_sub`? or no risk of underflow?

indexer/src/indexer/reconciliation.rs
```diff
@@ -171,82 +181,36 @@ async fn log_orphan_deposit_rows_at_startup(storage: &Storage) {
...
-    use solana_rpc_client_api::{client_error::ErrorKind, request::RpcError};
-
-    if let ErrorKind::RpcError(RpcError::RpcResponseError { code, message, .. }) = &e.kind {
-        if *code == INVALID_PARAMS_CODE as i64 {
-            return true;
+) -> Result<u64, ReconciliationError> {
+    match net_to_u64(net) {
+        NetBalance::Exact(v) => Ok(v),
+        NetBalance::Negative => {
+            warn!(
```
sebatustra, line 196: i wonder is this `warn!` is enough. Isn't it a serious issue to have a negative balance?

indexer/src/operator/escrow_sweep.rs
```diff
@@ -0,0 +1,305 @@
...
+            } else {
+                warn!(
+                    token_program = %token_program_id,
+                    "Skipping escrow token account with unrecognised data encoding"
+                );
+                continue;
+            };
+
+            // Multiple token accounts can hold the same mint; sum them. Saturating so a
+            // pathological set of balances reports u64::MAX instead of wrapping to a small value.
```
sebatustra: what does "pathological set of balances" mean?

indexer/src/indexer/reconciliation.rs
Huzaifa696: A negative net means completed withdrawals exceed all indexed deposits. In a healthy system that can't happen. 

It can only happen when the DB is missing deposit history it should have e.g a partially-restored DB indexed from a slot after some deposits, a backfill gap where a withdrawal is indexed before its deposit etc.
I'll tighten the doc comment to name that.

indexer/src/indexer/reconciliation.rs
Huzaifa696: No underflow risk here. `total_deposits` / `total_withdrawals` are `BigDecimal` (signed), so the subtraction can't wrap or panic. 

The negative case is handled explicitly downstream in `net_db_expected` (clamp to 0 + warn).

indexer/src/indexer/reconciliation.rs
Huzaifa696: It does fail closed when it matters. With the default threshold of 0, a negative net clamps `db_expected` to 0, so `compute_mismatch(0, on_chain)` = on-chain balance, which trips the gate (`classify_and_report`) and aborts startup. 

The `warn!`-only path is reached solely when on-chain is also ~0, i.e. a DB history gap with no funds at risk. So Negative case is reachable in legitimate degraded states and hard-failing it would block recovery.

indexer/src/operator/escrow_sweep.rs
Huzaifa696: It means the per-mint sum across the escrow's own token accounts exceeding `u64::MAX`. 

It can't actually happen because a mint's total supply is itself u64-capped, so it's purely defensive against corrupt RPC data. I've reworded the comment to say that plainly.

sebatustra (approved):

PR #168, opened 2026-06-23
fix: Unblock devnet deployment

gitteri:
Three bugs surfaced while standing up a devnet deployment off main, and each one blocks a clean deploy. This branch is scoped to just those fixes so the next person doesn't hit them. Each fix is its own commit so they can be reviewed or reverted independently.

1. Image build: `core` depends on `dvp-swap-program-client`, but the Dockerfile only generated the escrow and withdraw clients (the dvp program only ran `cargo-build-sbf`). The final release build then failed compiling the dvp client with a missing `generated` module. Now it generates the dvp client too, matching the other two programs.

2. Read node crash loop: read mode nodes connect to a read only Postgres replica, but `repair_address_signatures` writes (it seeds and advances the watermark). On a standby that fails with "cannot execute INSERT in a read-only transaction" as soon as any block exists, so the read node never stays healthy. The write node already owns repair on the primary, and the index reaches the replica via streaming replication, so the read path now skips repair.

3. Indexer config not found: the devnet compose mounts `scripts/devnet/config` into the indexer and operator containers (`program_type` and other structural config live there), but the Ansible render phase only copied `indexer/config`. The bind mount resolved to an empty directory and the services crashed with `MissingField(program_type)`. The render now ships `scripts/devnet/config` to the target host.

### Coverage Report

| Core | 11,149 | 12,761 | 87.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/28089092787) |
| Indexer | 22,892 | 25,988 | 88.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/28089092787) |
| Gateway | 1,035 | 1,195 | 86.6% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/28089092787) |
| Auth | 717 | 831 | 86.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/28089092787) |
| E2E Integration | 12,126 | 15,116 | 80.2% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/28089092787) |
| **Total** | **47,919** | **55,891** | **85.7%** | |

> Last updated: 2026-06-24 10:03:42 UTC by E2E Integration


core/src/nodes/node.rs
gitteri, line 336: @Huzaifa696 why did you add this?

Huzaifa696: I added that read-path `repair_address_signatures` call in #132. That was a mistake for read nodes which connects to the read-only replica.

It's the same read-node crash I later **fixed** in #163 (already on main).  Removing it on the read path here is the right call, the cleaner version of the same fix.

Huzaifa696:
@gitteri  **Update on fix no. 3**: merged main in, the devnet-config copy already existed there from #163 with a `when: network != 'localnet'` guard. 
The review bot's network-scoping suggestion is covered by the version on main.

Huzaifa696 (approved):

sebatustra (approved):

PR #170, opened 2026-06-24
feat: auth-admin set-role command (and actually ship the auth admin binary)

gitteri:
Replaces the manual SQL UPDATE for granting operator role with a proper CLI command, and fixes a latent bug that kept the auth admin binary out of the image.

**The shadowing bug:** core/src/bin/admin.rs and auth/src/bin/admin.rs both compile to a binary named admin, so the release build collides on target/release/admin and the Dockerfile only ever copied core's. The auth admin tool (its attach-wallet command) was never actually present in the deployed image. Renamed auth's binary to auth-admin and ship it.

**The feature:** add a set-role subcommand that sets a user's role by username, reading the DB URL from the environment like attach-wallet does (so the password never lands in argv). Promoting an account is now:

    docker exec private-channel-auth auth-admin set-role --username <name> --role operator

Verified on the live devnet container: the command promotes the account, is idempotent, errors cleanly on an unknown user, and the gateway then accepts the operator token. Includes tests for the success and unknown-user paths.
### Coverage Report

| Core | 11,149 | 12,761 | 87.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/28147710677) |
| Indexer | 22,898 | 25,994 | 88.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/28147710677) |
| Gateway | 1,035 | 1,195 | 86.6% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/28147710677) |
| Auth | 785 | 903 | 86.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/28147710677) |
| **Total** | **35,867** | **40,853** | **87.8%** | |

> Last updated: 2026-06-25 05:01:49 UTC by Auth

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -97,7 +97,7 @@ RUN mkdir -p private-channel-escrow-program/program/src private-channel-escrow-p
     private-channel-escrow-program/clients/rust/src private-channel-withdraw-program/clients/rust/src \
     dvp-swap-program/program/src dvp-swap-program/tests/integration-tests/src \
     dvp-swap-program/tests/transfer-hook-fixture/src dvp-swap-program/clients/rust/src \
-    core/src metrics/src auth/src bench-tps/src
+    core/src metrics/src auth/src auth/src/bin bench-tps/src
 RUN touch private-channel-escrow-program/program/src/lib.rs private-channel-escrow-program/tests/integration-tests/src/lib.rs \
     private-channel-escrow-program/clients/rust/src/lib.rs private-channel-withdraw-program/program/src/lib.rs \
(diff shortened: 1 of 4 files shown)


gitteri:
Addressed the `set_user_role` feedback in 271029e: it now takes the typed `Role` enum instead of a `&str`, so the variant is compiler-enforced and the lowercase mapping lives in the db layer. The CLI passes `RoleArg::to_model()`.

sebatustra (approved):

PR #182, opened 2026-06-30
fix(indexer): retry failed slot writes, crash-to-restart instead of skipping

Huzaifa696:
## Summary

Closes issue 4. In live/ungated operation the durable checkpoint tracked the highest slot seen, not the highest contiguous slot committed: when slot `N`'s writes failed its checkpoint was withheld but the processor moved on, and `N+1`'s success persisted `max(frontier, N+1)`, leapfrogging `N` so on restart a missed deposit was never minted and a missed withdrawal never released. 

The fix retries a slot's writes with bounded backoff and sends the checkpoint only after they commit; since the processor is a single sequential loop, `N+1` is never processed until `N` is persisted. On exhaustion the processor returns a fatal error that exits `run()` so the supervisor restarts and replays `N`.

Tested: transient blips self-heal, permanent failures exhaust to `Err`, deposit-withhold semantics hold, and the key proof (`start` returns `Err`, no checkpoint sent, `N+1` never persisted) plus the biased `supervise()` race are covered 

### Coverage Report

| Core | 11,486 | 13,120 | 87.5% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/28443932886) |
| Indexer | 23,200 | 26,282 | 88.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/28443932886) |
| Gateway | 1,035 | 1,195 | 86.6% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/28443932886) |
| Auth | 785 | 903 | 86.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/28443932886) |
| E2E Integration | 12,174 | 15,167 | 80.3% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/28443932886) |
| **Total** | **48,680** | **56,667** | **85.9%** | |

> Last updated: 2026-06-30 12:48:42 UTC by E2E Integration

### DIFF
--- a/indexer/src/indexer/transaction_processor.rs
+++ b/indexer/src/indexer/transaction_processor.rs
@@ -198,6 +198,11 @@ impl TransactionProcessor {
                             "Slot {} writes failed after {} attempt(s); giving up: {}",
                             slot, attempt, e
                         );
+                        // Count the slot once, only when the retry budget is spent;
+                        // a transient failure ridden out by the retry is not an error.
+                        metrics::INDEXER_SLOT_SAVE_ERRORS
+                            .with_label_values(&[program_type.as_label()])
+                            .inc();
                         return Err(e.into());
                     }
                     let backoff = self.retry.backoff(attempt);
@@ -270,9 +275,6 @@ impl TransactionProcessor {
                 }
                 Err(e) => {
                     error!("Failed to save mints from slot {}: {}", slot, e);
-                    metrics::INDEXER_SLOT_SAVE_ERRORS
-                        .with_label_values(&[program_type.as_label()])
-                        .inc();
                     return Err(e);
                 }
             }
@@ -292,9 +294,6 @@ impl TransactionProcessor {
                         "Failed to save mint status history from slot {}: {}",
                         slot, e
                     );
-                    metrics::INDEXER_SLOT_SAVE_ERRORS
-                        .with_label_values(&[program_type.as_label()])
-                        .inc();
                     return Err(e);
                 }
             }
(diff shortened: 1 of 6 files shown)


indexer/src/indexer/transaction_processor.rs
```diff
@@ -21,8 +21,45 @@ use private_channel_metrics::{HealthState, MetricLabel};
...
 use std::sync::Arc;
+use std::time::Duration;
 use tokio::sync::mpsc;
-use tracing::{debug, error, info, warn};
+use tracing::{debug, error, info};
+
+/// Production defaults for the per-slot DB-write retry. Sized to ride out a
+/// routine Postgres restart or failover (about 15s of cumulative backoff)
+/// before the processor gives up and exits so a restart can replay the slot.
+const DEFAULT_WRITE_MAX_ATTEMPTS: usize = 6;
```
sebatustra, line 31: seems like production always builds the processor with this default policy (`.with_write_retry ()`is only called in tests). Was it considered to make these config-driven? In case an operator wants to extend the window without a recompile

Huzaifa696: That's intentional. The retry window is meant as a defensive timeout rather than an operator-tunable setting. The default (~15s) is enough to ride out routine Postgres restarts, and if the outage lasts longer, exiting is safe since the indexer replays idempotently from the last durable checkpoint on restart.

So the timeout mainly trades off waiting longer versus restarting sooner, not correctness, which is why I left it hardcoded. `with_write_retry()` exists primarily as a testing seam.

PR #192, opened 2026-07-08
fix(indexer): make a finality Dead verdict mean coverage-proven, corroborated non-inclusion

Huzaifa696:
## Summary

Closes issue 15. Builds on PR #187 by making a `Dead` verdict mean **coverage-proven, corroborated non-inclusion**. Previously, an absent `getSignatureStatuses` response could be incorrectly treated as proof that a transaction never landed, even if the RPC node had pruned history or was behind.

The fix adds a ledger-floor (`getFirstAvailableBlock`) check before trusting any absence-based `Dead` verdict. If the endpoint cannot prove it still covers the transaction window, the result becomes `Uncertain` and the system refuses to remint. The withdraw operator now requires an independent fallback RPC, validated at startup for reachability, distinct URL, and same-cluster genesis hash.

### Coverage Report

| Core | 11,486 | 13,120 | 87.5% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29085887826) |
| Indexer | 25,892 | 29,175 | 88.7% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29085887826) |
| Gateway | 1,325 | 1,515 | 87.5% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29085887826) |
| Auth | 785 | 903 | 86.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29085887826) |
| E2E Integration | 12,418 | 15,666 | 79.3% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29085887826) |
| **Total** | **51,906** | **60,379** | **86.0%** | |

> Last updated: 2026-07-10 11:22:01 UTC by E2E Integration

### DIFF
--- a/indexer/src/operator/operator.rs
+++ b/indexer/src/operator/operator.rs
@@ -45,19 +45,19 @@ pub async fn run(
 
     // Optional destination fallback for recovery and the boot pre-flight to
     // re-check a Dead verdict. Empty means unset (env renders "") and maps to None.
-    let fallback_rpc_client = common_config
+    let normalized_fallback_url = common_config
         .fallback_rpc_url
         .as_deref()
-        .filter(|s| !s.is_empty())
-        .map(|url| {
-            Arc::new(RpcClientWithRetry::with_retry_config(
-                url.to_string(),
-                RetryConfig::default(),
-                CommitmentConfig {
-                    commitment: config.rpc_commitment,
-                },
-            ))
-        });
+        .filter(|s| !s.is_empty());
+    let fallback_rpc_client = normalized_fallback_url.map(|url| {
+        Arc::new(RpcClientWithRetry::with_retry_config(
+            url.to_string(),
+            RetryConfig::default(),
+            CommitmentConfig {
+                commitment: config.rpc_commitment,
+            },
+        ))
+    });
 
     // The withdraw operator's compensating remint MintTo must broadcast on the source
     // chain (PrivateChannel), where the burn happened. Without source_rpc_url the sender
@@ -74,11 +74,7 @@ pub async fn run(
     }
 
     // A lone prunable Solana RPC's absent status is not proof of non-inclusion, so require an
-    // independent, same-cluster, reachable fallback (empty env string treated as unset) before starting.
-    let normalized_fallback_url = common_config
-        .fallback_rpc_url
-        .as_deref()
-        .filter(|s| !s.is_empty());
+    // independent, same-cluster, reachable fallback before starting.
     validate_withdraw_fallback(
         common_config.program_type,
         &rpc_client,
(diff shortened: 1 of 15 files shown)


indexer/src/operator/sender/remint.rs
```diff
@@ -354,71 +342,169 @@ impl<'a> FinalityRpc<'a> {
     }
 }
 
-/// Classify `sigs`. Only a `Dead` verdict is re-checked against the fallback; the
-/// safe verdicts return from `primary` first, so the fallback is queried rarely.
+/// A blockhash is valid for this many blocks (`MAX_PROCESSING_AGE`), so a landed
+/// transaction's block height is at most this far below its `last_valid_block_height`.
+const BLOCKHASH_VALIDITY_BLOCKS: u64 = 150;
```
sebatustra: can we just use `clock::MAX_PROCESSING_AGE`

indexer/src/operator/operator.rs
```diff
@@ -73,6 +73,21 @@ pub async fn run(
         ));
     }
 
+    // A lone prunable Solana RPC's absent status is not proof of non-inclusion, so require an
+    // independent, same-cluster, reachable fallback (empty env string treated as unset) before starting.
+    let normalized_fallback_url = common_config
```
sebatustra: the `.as_deref().filter(|s| !s.is_empty())` (for `fallback_rpc_client`)is also present in line 48 in this file, is it possible to normalize it once?

indexer/src/operator/sender/remint.rs
```diff
@@ -354,71 +342,169 @@ impl<'a> FinalityRpc<'a> {
...
+    DeadByAbsence {
+        min_lvbh: u64,
+    },
+}
+
+/// Ledger-floor coverage decision for an absence-based `Dead`. `covered` is the
+/// verdict; `floor` and `bound` are surfaced so an uncovered `Uncertain` names the
+/// concrete slots an operator must retain to resolve it.
+struct AbsenceCoverage {
+    covered: bool,
```
sebatustra: if `AbsenceCoverage.covered` is always set to `covered: floor <= bound`, cant we just inline the floor <= bound comparison in `coverage_verdict` and drop the `covered` field?

indexer/src/operator/sender/remint.rs
Huzaifa696: Done

indexer/src/operator/operator.rs
Huzaifa696: normalized once and reused for both.

indexer/src/operator/sender/remint.rs
Huzaifa696: inlined and dropped the struct entirely.

sebatustra (approved):

PR #197, opened 2026-07-10
fix(indexer): ownership-checked write-ahead persist so a requeued deposit cannot double-mint

Huzaifa696:
## Summary

Closes issue 113.  A deposit transaction could be requeued while a sender still held an in-memory `Mint` awaiting broadcast. Since the transaction signature had not yet been persisted, the recovery worker interpreted the `Processing` row as never broadcast, demoted it back to `Pending`, and allowed it to be fetched again. The original sender could still broadcast its transaction, causing the re-fetched deposit to be minted a second time and creating unbacked channel token supply.

This change makes persisting the deposit transaction signature an atomic ownership claim. Before broadcasting, the sender now performs a compare-and-swap (CAS) against the deposit row's fetch-time `updated_at` and only persists the signature if it still owns that exact `Processing` incarnation. If ownership has already been lost, the builder is discarded without broadcasting, preventing the recovery worker and a live sender from racing to mint the same deposit. The recovery flow itself is unchanged; it simply can no longer reclaim work owned by an active sender.

Unit and integration tests cover ownership claiming, race conditions, and end-to-end scenarios to verify that a deposit can be requeued safely without ever being double-minted.
### Coverage Report

| Core | 11,486 | 13,120 | 87.5% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29245886678) |
| Indexer | 26,354 | 29,700 | 88.7% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29245886678) |
| Gateway | 1,325 | 1,515 | 87.5% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29245886678) |
| Auth | 785 | 903 | 86.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29245886678) |
| E2E Integration | 12,557 | 15,819 | 79.4% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29245886678) |
| **Total** | **52,507** | **61,057** | **86.0%** | |

> Last updated: 2026-07-13 12:22:59 UTC by E2E Integration

### DIFF
--- a/indexer/src/storage/postgres/db.rs
+++ b/indexer/src/storage/postgres/db.rs
@@ -1123,7 +1123,7 @@ impl PostgresDb {
         // Use a transaction to ensure atomicity
         let mut tx = self.pool.begin().await?;
 
-        let transactions = sqlx::query_as::<_, DbTransaction>(&format!(
+        let mut transactions = sqlx::query_as::<_, DbTransaction>(&format!(
             r#"
             SELECT
                 {}, {}, {}, {}, {}, {}, {}, {}, {}, {},
@@ -1175,7 +1175,6 @@ impl PostgresDb {
         // Update status to Processing and RETURNING the trigger-bumped
         // `updated_at`, so the fetched row carries its true post-lock token (the
         // deposit sender CASes on it at broadcast, not the stale Pending value).
-        let mut transactions = transactions;
         if !transactions.is_empty() {
             let ids: Vec<i64> = transactions.iter().map(|txn| txn.id).collect();
             let bumped: Vec<chrono::DateTime<chrono::Utc>> = sqlx::query_scalar(&format!(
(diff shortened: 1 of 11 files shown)


indexer/src/storage/postgres/db.rs
```diff
@@ -1172,18 +1172,31 @@ impl PostgresDb {
         .fetch_all(&mut *tx)
         .await?;
 
-        // Update status to Processing in a single query
+        // Update status to Processing and RETURNING the trigger-bumped
+        // `updated_at`, so the fetched row carries its true post-lock token (the
+        // deposit sender CASes on it at broadcast, not the stale Pending value).
+        let mut transactions = transactions;
```
sebatustra: nit: pointless rebinding (declare original binding `mut` instead)

sebatustra (approved):

indexer/src/storage/postgres/db.rs
Huzaifa696: fixed

PR #208, opened 2026-07-21
fix(indexer): close the reconnect residual tail gap with an in-band checkpoint gate (issue-251)

Huzaifa696:
## Summary

Closses issue 251. On a Yellowstone reconnect, the datasource previously backfilled up to a snapshot of the current tip and only then resubscribed. Any blocks produced between the snapshot and the new subscription were processed by neither path. Because the checkpoint advanced normally once live streaming resumed, it could permanently skip this "residual tail" window, silently missing deposits or withdrawals that occurred during it. The longer the outage, the larger this unindexed window could become.

This PR closes that gap by subscribing first, observing the actual resume slot, and then gating checkpoint advancement until the residual range has been backfilled. On the first live block after reconnect, the datasource emits an in-band `Regate` message ahead of the block on the existing processing pipeline and concurrently backfills the missing slots. Because the gate and the protected block travel through the same FIFO pipeline, the checkpoint cannot advance past the reconnect boundary before the backfill completes, while live streaming continues uninterrupted.

The existing backfill logic is unchanged; only the reconnect flow is updated. After successfully reconnecting, the checkpoint is temporarily held at its last durable value while a single backfill catches up to the new resume point. Once the missing slots are filled, the checkpoint continues advancing normally.
### Coverage Report

| Core | 11,846 | 13,489 | 87.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29995472226) |
| Indexer | 29,813 | 33,150 | 89.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29995472226) |
| Gateway | 1,325 | 1,515 | 87.5% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29995472226) |
| Auth | 785 | 903 | 86.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29995472226) |
| E2E Integration | 13,223 | 16,700 | 79.2% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29995472226) |
| **Total** | **56,992** | **65,757** | **86.7%** | |

> Last updated: 2026-07-23 10:21:40 UTC by E2E Integration

### DIFF
--- a/indexer/src/indexer/datasource/yellowstone/source.rs
+++ b/indexer/src/indexer/datasource/yellowstone/source.rs
@@ -257,6 +257,10 @@ struct ReconnectGapCtx {
 /// the fold at the durable checkpoint, then spawn a bounded backfill (one in-flight; any
 /// prior is aborted). Sent before the first live block so the gate wins the FIFO race.
 #[cfg(feature = "datasource-rpc")]
+/// Returns whether it is safe to forward the block that triggered this arm. `false`
+/// means cancellation interrupted the checkpoint read before the gate was armed, so the
+/// caller must NOT forward the block: its SlotComplete would advance the checkpoint over
+/// the still-unfilled gap. `true` means the gate is armed, or a fresh system needs none.
 async fn arm_reconnect_gap(
     t_sub: u64,
     ctx: &ReconnectGapCtx,
@@ -264,10 +268,10 @@ async fn arm_reconnect_gap(
     cancel: &CancellationToken,
     backoff: Duration,
     prev: &mut Option<tokio::task::JoinHandle<()>>,
-) -> Result<(), DataSourceRpcError> {
+) -> Result<bool, DataSourceRpcError> {
     // Resolve the anchor before arming: the gate must seed the frontier to the durable
-    // checkpoint, so a wrong/zero `from` would freeze it. Retry a failed read (fail
-    // closed: the first block is not forwarded yet) until cancellation.
+    // checkpoint, so a wrong/zero `from` would freeze it. Retry a failed read until it
+    // succeeds or cancellation ends it (returning false so the block is not forwarded).
     let checkpoint = loop {
         match get_last_checkpoint(&ctx.storage, ctx.program_type).await {
             Ok(slot) => break slot,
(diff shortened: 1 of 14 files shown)


indexer/src/indexer/transaction_processor.rs
```diff
@@ -1723,6 +1828,115 @@ mod tests {
         assert!(committed >= DEPOSIT_SLOT);
     }
 
+    /// The exact issue #251 reproduction, end-to-end through the real pipeline: the
```
sebatustra: nit: lets avoid referencing audit tickets on the codebase, this wont make sense to future readers

indexer/src/indexer/datasource/yellowstone/source.rs
```diff
@@ -251,6 +241,99 @@ where
...
+    // checkpoint, so a wrong/zero `from` would freeze it. Retry a failed read (fail
+    // closed: the first block is not forwarded yet) until cancellation.
+    let checkpoint = loop {
+        match get_last_checkpoint(&ctx.storage, ctx.program_type).await {
+            Ok(slot) => break slot,
+            Err(e) => {
+                warn!("Reconnect gap: checkpoint read failed, retrying: {}", e);
+                record_gap_fill_error(ctx.program_type);
+                if backoff_cancelled(cancel, backoff).await {
+                    return Ok(());
```
sebatustra: If it is shutting down while arm_reconnect_gap is retrying, it returns Ok(()) just like success. Can't that skip the unfilled slots the fix is supposed to protect?

indexer/src/indexer/transaction_processor.rs
Huzaifa696: done

indexer/src/indexer/datasource/yellowstone/source.rs
Huzaifa696: `arm_reconnect_gap` now returns false on cancel-during-read so the block is withheld (not forwarded), and a restart replays it. Test added.

PR #217, opened 2026-07-24
fix: prevent deposit misattribution from short getSignatureStatuses responses (issue-280)

Huzaifa696:
## Summary

Closes issue 280. `getSignatureStatuses` returns results by position, but both confirmation pollers concatenated chunked RPC responses without verifying that each chunk returned the expected number of entries. A malformed response (too few or too many statuses) could shift later results, causing confirmations to be applied to the wrong transactions and potentially marking an unconfirmed deposit as `Completed`, preventing recovery from retrying it.

This PR introduces a shared `fetch_statuses_checked` helper that validates every chunk returns exactly one status per requested signature before combining the results. If any chunk has an incorrect length or the RPC call fails, the entire batch is requeued without incrementing `poll_attempts`, matching the existing transient-error behavior. This prevents positional misalignment while leaving the polling flow otherwise unchanged, adding only an O(1) length check per chunk.

### Additional changes

**Build fix:** Updated `processor.rs` to use the newer `RequeueOutcome` API after rebasing onto `cantina-audit`, fixing a compile error caused by an API mismatch between earlier PRs.

### Coverage Report

| Core | 11,847 | 13,489 | 87.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30255234054) |
| Indexer | 30,831 | 34,191 | 90.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30255234054) |
| Gateway | 1,325 | 1,515 | 87.5% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30255234054) |
| Auth | 785 | 903 | 86.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30255234054) |
| E2E Integration | 13,249 | 16,780 | 79.0% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30255234054) |
| **Total** | **58,037** | **66,878** | **86.8%** | |

> Last updated: 2026-07-27 10:26:47 UTC by E2E Integration

### DIFF
--- a/indexer/src/operator/sender/transaction.rs
+++ b/indexer/src/operator/sender/transaction.rs
@@ -1596,13 +1596,32 @@ pub(super) async fn route_poll_results(
     }
 }
 
+/// Why a chunked status fetch was rejected. Both variants make the caller reinsert
+/// the batch and retry; the split only picks the metric reason label.
+#[derive(Debug)]
+enum StatusFetchError {
+    /// A chunk response length did not equal the request (short or oversized).
+    MalformedLength,
+    /// The RPC call itself failed after retries.
+    Rpc,
+}
+
+impl StatusFetchError {
+    fn reason(&self) -> &'static str {
+        match self {
+            StatusFetchError::MalformedLength => "malformed_status_response",
+            StatusFetchError::Rpc => "status_poll_rpc_error",
+        }
+    }
+}
+
 /// Fetch statuses in `MAX_SIGS_PER_CALL` chunks. `getSignatureStatuses` is positional, so a
 /// chunk whose length differs from the request would misalign every later status; reject it.
 /// Returns `Err` on any RPC error or length mismatch so the caller reinserts the batch and retries.
 async fn fetch_statuses_checked(
     rpc_client: &RpcClientWithRetry,
     signatures: &[Signature],
-) -> Result<Vec<Option<solana_transaction_status::TransactionStatus>>, ()> {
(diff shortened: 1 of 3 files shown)


indexer/src/operator/sender/transaction.rs
```diff
@@ -1596,6 +1596,40 @@ pub(super) async fn route_poll_results(
...
+/// Returns `Err` on any RPC error or length mismatch so the caller reinserts the batch and retries.
+async fn fetch_statuses_checked(
+    rpc_client: &RpcClientWithRetry,
+    signatures: &[Signature],
+) -> Result<Vec<Option<solana_transaction_status::TransactionStatus>>, ()> {
+    let mut statuses = Vec::with_capacity(signatures.len());
+    for chunk in signatures.chunks(MAX_SIGS_PER_CALL) {
+        match rpc_client.get_signature_statuses(chunk).await {
+            Ok(resp) if resp.value.len() == chunk.len() => statuses.extend(resp.value),
+            Ok(resp) => {
```
sebatustra, line 1629: should  we add metrics in the reject path?

Huzaifa696: Added metrics to distinguish malformed status responses (`malformed_status_response`) from transient RPC failures (`status_poll_rpc_error`) in `OPERATOR_TRANSACTION_ERRORS`.

PR #221, opened 2026-07-27
fix: Retarget escrow program id to 9tgHa1, fix deposit.rs Token-2022 detection

gitteri:
## Summary
- Main was still pointing at the old placeholder escrow program (GokvZqD2yP696rzNBNbQvcZ4VsLW7jNvFXU1kW9m7k83), but devnet has actually been running on our own program (9tgHa1DcnaSSUtmMsst8ovKTe1Gfxzezn27KnH9xXYeU) for weeks. Updates streamer.rs, the indexer's escrow parser, the IDL, and the program's own declare_id to match.
- deposit.rs hardcoded spl_token::ID for the ATA derivation and token_program account, which breaks for Token-2022 mints. Fixes it to detect the mint's actual owning token program, same as allow_mint.rs got in #174 (that PR missed this sibling script).
### Coverage Report

| Component | Lines Hit | Lines Total | Coverage | Artifact |
|-----------|-----------|-------------|----------|----------|
| Core | 11,407 | 13,001 | 87.7% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30300664689) |
| Indexer | 23,018 | 26,109 | 88.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30300664689) |
| Gateway | 1,035 | 1,195 | 86.6% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30300664689) |
| Auth | 785 | 903 | 86.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30300664689) |
| Withdraw Program | 129 | 241 | 53.5% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30300664801) |
| Escrow Program | 1,195 | 1,982 | 60.3% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30300664801) |
| DvP Swap Program | 270 | 1,179 | 22.9% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30300664801) |
| E2E Integration | 12,144 | 15,174 | 80.0% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30300664689) |
| **Total** | **49,713** | **58,605** | **84.8%** | |

[...]

PR #222, opened 2026-07-27
chore(deploy): Support locally-built images and add audit/sdp inventory groups

gitteri:
## Summary
- skip_pull lets a stack run a locally-built image with no registry, by skipping the GHCR creds assertion, docker login, and compose pull. Needed for stacks that build the retargeted 9tgHa1 escrow program locally instead of pulling from GHCR.
- Adds [audit] and [sdp] inventory groups for the two isolated devnet stacks: the external security auditor and the SDP dev shop integration.
- .gitignore now excludes any *secrets.yml so per-stack secret overrides (audit-secrets.yml, sdp-secrets.yml) never get committed.
### Coverage Report

| Core | - | - | - | - |
| Indexer | - | - | - | - |
| Gateway | - | - | - | - |
| Auth | - | - | - | - |

### DIFF
--- a/private-channel-deploy/.gitignore
+++ b/private-channel-deploy/.gitignore
@@ -10,6 +10,9 @@ ansible.log
 # improvements" for the SOPS upgrade path that lets this go in git encrypted.
 secrets.yml
 
+# Per-stack secret overrides (e.g. the auditor stack). Same rule as secrets.yml.
+*secrets.yml
+
 # Reserved for the SOPS upgrade. Keys must never live in the repo.
 age.key
 keys.txt
--- a/private-channel-deploy/.gitignore
+++ b/private-channel-deploy/.gitignore
@@ -10,7 +10,7 @@ ansible.log
 # improvements" for the SOPS upgrade path that lets this go in git encrypted.
 secrets.yml
 
-# Per-stack secret overrides (e.g. the auditor stack). Same rule as secrets.yml.
+# Per-stack secret overrides (e.g. sdp-secrets.yml). Same rule as secrets.yml.
 *secrets.yml
 
 # Reserved for the SOPS upgrade. Keys must never live in the repo.
--- a/private-channel-deploy/deploy.yml
+++ b/private-channel-deploy/deploy.yml
@@ -335,6 +335,8 @@
           Fix: edit secrets.yml and add `jwt_secret: <random-32+-byte-string>`  (e.g. `openssl rand -hex 32`).
       when: gateway_enforce_jwt | default(true) | bool
 
+    # When skip_pull is set the stack runs a locally-built image (no registry),
+    # so GHCR creds aren't needed.
     - name: Assert GHCR creds present
       ansible.builtin.assert:
         that:
@@ -343,6 +345,7 @@
         fail_msg: |
           `ghcr_user`/`ghcr_token` empty in secrets.yml. Deploy needs them to docker-login to GHCR before pulling images.
           Fix: edit secrets.yml — `ghcr_user: <github-username>`, `ghcr_token: <PAT with read:packages>`.
+      when: not (skip_pull | default(false) | bool)
 
     # Yellowstone gRPC required for indexer-solana on public clusters
     # (localnet uses the bundled validator's Yellowstone, no external endpoint).
@@ -429,6 +432,7 @@
         executable: /bin/bash
       changed_when: true
       no_log: true
+      when: not (skip_pull | default(false) | bool)
 
     # Drop the marker last so a partial bootstrap re-runs cleanly.
     - name: Drop bootstrap marker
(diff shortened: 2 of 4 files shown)


gitteri:
Fixed: moved the GHCR credential write out of the one-time bootstrap block so it runs on every deploy (gated only by `skip_pull`), instead of being tied to whether bootstrap already ran on that host.

gitteri:
Fixed: tagged the relocated GHCR-auth task `bootstrap, deploy` (was `bootstrap` only). Verified with `--list-tasks --tags deploy` that it now runs ahead of `Pull images from registry` in a tag-scoped deploy-only invocation.

gitteri:
Fixed: added `pull_policy: never` to every app + validator service in the compose override when `skip_pull` is set. Previously the override only set `image: local/private-channel-app:<tag>` with no pull policy, so compose's default "missing" policy could still reach for the registry if that exact tag wasn't found locally for any reason. Verified by rendering the template with skip_pull true/false/unset and confirming valid YAML output with `pull_policy: never` present only in the skip_pull=true case.

private-channel-deploy/inventory.ini
```diff
@@ -8,3 +8,8 @@
 # Pre-trust with: `ssh-keyscan -H <host> >> ~/.ssh/known_hosts`.
 [dev]
 private-channel ansible_host=<IP_ADDRESS> ansible_user=<USER> ansible_ssh_private_key_file=~/.ssh/id_ed25519 ansible_host_key_checking=False
+
+# SDP integration stack (spc-sdp-vm). Reuses the deployed programs; own instance.
+# Gateway open (JWT enforcement off). Vars: vars/sdp.yml.
```
Huzaifa696, line 13: deploy.yml derives the vars file from the host's group, so -l sdp aborts at load because vars/sdp.yml isn't in the repo, and skip_pull is assigned nowhere so the feature ships inert.

Fix: commit vars/sdp.yml with skip_pull: true + image_registry/image_tag/gateway_enforce_jwt: false, or drop [sdp] from the shipped inventory.

private-channel-deploy/inventory.ini
```diff
@@ -8,3 +8,8 @@
...
+[sdp]
+private-channel-sdp ansible_host=34.71.147.163 ansible_user=spcadmin ansible_ssh_private_key_file=~/.ssh/spc_gce ansible_host_key_checking=False
```
Huzaifa696, line 15: inventory.ini:15 commits a real public IP, SSH user and key path, annotated "Gateway open (JWT enforcement off)", while [dev] one line above uses <IP_ADDRESS>/<USER> placeholders.

Fix: placeholders in git, real host in a gitignored inventory.local.ini passed via -i.

private-channel-deploy/.gitignore
```diff
@@ -10,6 +10,9 @@ ansible.log
 # improvements" for the SOPS upgrade path that lets this go in git encrypted.
 secrets.yml
 
+# Per-stack secret overrides (e.g. sdp-secrets.yml). Same rule as secrets.yml.
+*secrets.yml
```
Huzaifa696, line 14: .gitignore now ignores *secrets.yml "e.g. sdp-secrets.yml", but pre_tasks (deploy.yml:17) only includes secrets.yml, so that file would be silently unused and the stack deploys with dev secrets.

Fix: load <env>-secrets.yml when it exists (stat + conditional include_vars), or reword the gitignore comment to stop promising it.

private-channel-deploy/deploy.yml
```diff
@@ -409,34 +412,42 @@
...
+      cfg = json.loads(p.read_text()) if p.exists() else {}
+      cfg.setdefault('auths', {})['ghcr.io'] = {'auth': open('/tmp/.ghcr_b64').read()}
+      p.write_text(json.dumps(cfg, indent=2))
+      "
+      rm -f /tmp/.ghcr_b64
+    args:
+      executable: /bin/bash
+    changed_when: true
+    no_log: true
+    tags: [ bootstrap, deploy ]
```
Huzaifa696, line 448: on --tags deploy the login task runs but its preflight-tagged creds assert doesn't, so blank secrets overwrite the host's working ~/.docker/config.json with base64(":") and the pull fails later with a bare 401.
Fix: tag the assert [ preflight, bootstrap, deploy ], optionally plus when: ghcr_user | default('') | length > 0 on the writer.

gitteri:
Closing in favor of https://github.com/solana-foundation/solana-private-channels/pull/270. Pivoting away from skip_pull (locally-built images, no provenance) to a real image-publish pipeline: CI builds and pushes a commit-sha-tagged app image to GHCR, and stacks pin that. This is unblocked now that #221 committed the 9tgHa1 program id to main, so the canonical image is correct.

huzaifa's points still stand and carry into follow-ups: the audit/sdp inventory groups + conditional <env>-secrets.yml loading are genuinely needed (done right this time), and the real host IP / open-gateway annotation should live in a gitignored inventory.local.ini, not git. Those are separable from the image pipeline and will land as their own change.

PR #228, opened 2026-07-30
fix(core): replace the writable-lamport cap with fabricated fee-payer conservation (issue-115-131)

Huzaifa696:
## Summary

Closes issues 115 and 131. To support gasless transactions, the callback fabricates 10 lamports for unknown fee payers so they pass SVM validation. Previously, `cap_lamports` reclaimed those fabricated lamports after execution by rewriting every writable account based solely on its shape, regardless of whether the transaction actually created it. This unintentionally modified legitimate accounts: it deleted the DvP nonce tombstone that enforces single-use nonces, allowing a closed trade to be recreated with different terms, and it drained any writable wrapped SOL (wSOL) escrow account because native token balances are stored as lamports.

This PR fixes the issue by enforcing **lamport conservation** instead of rewriting accounts after execution. The fabricated lamports are treated as a temporary loan that must be fully accounted for before the transaction commits. After execution, `enforce_lamport_conservation` checks only newly created writable accounts, allowing one lamport per created account as required by the SVM while ensuring no fabricated lamports escape into existing state. If the loan does not balance, the transaction fails with `UnbalancedTransaction` rather than mutating account balances. Once balanced, the temporary fee payer accounts are removed, leaving all legitimate accounts untouched. This preserves normal balance transfers while preventing fabricated funds from becoming durable.
### Coverage Report

| Core | 12,963 | 14,609 | 88.7% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30807745371) |
| Indexer | 32,527 | 35,947 | 90.5% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30807745371) |
| Gateway | 1,330 | 1,521 | 87.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30807745371) |
| Auth | 785 | 903 | 86.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30807745371) |
| E2E Integration | 13,649 | 17,382 | 78.5% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30807745371) |
| **Total** | **61,254** | **70,362** | **87.1%** | |

> Last updated: 2026-08-03 11:56:14 UTC by E2E Integration

### DIFF
--- a/core/src/stages/execution.rs
+++ b/core/src/stages/execution.rs
@@ -418,17 +418,23 @@ fn enforce_lamport_conservation(
         }
 
         fabricated.clear();
-        // How much of the invented float went missing, and how many brand new
-        // accounts exist to justify it.
+        // How much of the invented float went missing, how many brand new
+        // accounts exist to justify it, and whether any account that already
+        // existed ended up richer than it started.
         let mut shortfall = 0u64;
         let mut created = 0u64;
+        let mut credited = false;
         for (index, (pubkey, acct)) in executed.loaded_transaction.accounts.iter().enumerate() {
-            // Read-only accounts are never persisted, and anything BOB already
-            // knew is real money this check must not touch or count.
-            if !tx.is_writable(index) || bob.knows_account(pubkey) {
+            // Read-only accounts are never persisted, so they are never touched
+            // or counted.
+            if !tx.is_writable(index) {
                 continue;
             }
-            if fee_payers.contains(pubkey) {
+            if let Some(before) = bob.account_lamports(pubkey) {
+                // Already existed, so its balance is real money this check must
+                // never rewrite. Only note whether it grew.
+                credited |= acct.lamports() > before;
+            } else if fee_payers.contains(pubkey) {
                 // An invented payer: whatever it no longer holds has escaped.
                 // The set is batch-wide, so draining one payer into another
                 // still registers as a shortfall rather than as repayment.
@@ -443,7 +449,12 @@ fn enforce_lamport_conservation(
         // Lamports against a count, because the rate is exactly one: the SVM
         // deletes a 0-lamport account, and with rent at zero every creation
         // path funds a single lamport. So `created` is also the allowance.
-        if shortfall > created {
+        //
+        // The second clause closes an attribution gap. Counting creations does
+        // not prove the float paid for them, so a creation funded by real money
+        // could licence a float lamport that actually landed somewhere durable.
+        // While any float is missing, no pre-existing balance may grow.
+        if shortfall > created || (shortfall > 0 && credited) {
             // Reject rather than claw back, which would mean rewriting accounts
             // the transaction legitimately owns. Downstream skips failed txs.
             executed.execution_details.status = Err(TransactionError::UnbalancedTransaction);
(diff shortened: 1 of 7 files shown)


core/src/stages/execution.rs
```diff
@@ -415,16 +416,45 @@ fn cap_lamports(
...
-                acct.set_lamports(cap);
+            if fee_payers.contains(pubkey) {
+                // An invented payer: whatever it no longer holds has escaped.
+                // The set is batch-wide, so draining one payer into another
+                // still registers as a shortfall rather than as repayment.
+                fabricated.push(index);
+                shortfall += DEFAULT_FEE_PAYER_LAMPORTS.saturating_sub(acct.lamports());
+            } else if acct.lamports() > 0 {
+                // Unknown to BOB but funded, so this transaction created it.
+                created += 1;
```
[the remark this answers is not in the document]
Huzaifa696: fixed. While any float is missing no pre-existing balance may grow, so a real-funded creation can no longer licence it. See credited_pre_existing_account_rejects_when_float_is_missing.

core/src/stages/execution.rs
```diff
@@ -415,16 +416,45 @@ fn cap_lamports(
...
+            // Reject rather than claw back, which would mean rewriting accounts
+            // the transaction legitimately owns. Downstream skips failed txs.
+            executed.execution_details.status = Err(TransactionError::UnbalancedTransaction);
+            continue;
+        }
+        // Loan settled. Erase the invented payers to the empty zero-lamport
+        // shape BOB and the settler already treat as deleted, and touch nothing
+        // else: an address that never existed must not become durable state.
+        for index in &fabricated {
+            executed.loaded_transaction.accounts[*index].1 = AccountSharedData::default();
```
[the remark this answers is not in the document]
Huzaifa696: Intentional. Persisting the payer graduates an invented address and refunding it rewrites the sender (the #131 bug), and rejecting would break CancelDvp, where the fee payer receives the closed escrows' rent.

core/src/stages/execution.rs
```diff
@@ -415,16 +416,59 @@ fn cap_lamports(
...
+        let mut credited = false;
+        for (index, (pubkey, acct)) in executed.loaded_transaction.accounts.iter().enumerate() {
+            // Read-only accounts are never saved, so never touch or count them.
             if !tx.is_writable(index) {
                 continue;
             }
-            let cap = if acct.data().is_empty() { 0 } else { 1 };
-            if acct.lamports() > cap {
-                acct.set_lamports(cap);
+            if let Some(before) = bob.account_lamports(pubkey) {
```
sebatustra, line 432: could a None from `account_lamports` be caused by a DB query error? currently Postgres returns a `vec![None; n]` in `get_accounts_postgres`

core/src/stages/execution.rs
```diff
@@ -415,16 +416,59 @@ fn cap_lamports(
...
+        // making an account gives it exactly one lamport.
+        //
+        // Counting new accounts does not prove the made-up money paid for them,
+        // so a new account paid for with real money could excuse a made-up
+        // lamport that went elsewhere. Hence no older account may grow while any
+        // made-up money is missing.
+        if shortfall > created || (shortfall > 0 && credited) {
+            // Fail the tx instead of taking the money back, which would mean
+            // rewriting accounts it owns. Later stages skip failed txs.
+            executed.execution_details.status = Err(TransactionError::UnbalancedTransaction);
```
sebatustra, line 468: should we add a metric or a warn! for this rejection?

Huzaifa696: Added both, a warn! with the signature and the shortfall/created/credited numbers, plus a private_channel_executor_conservation_rejected_total counter.

core/src/stages/execution.rs
Huzaifa696: Yes, a Postgres error or a corrupt row all collapse to None today. Already fixed on the ottersec audit branch in #216, which returns Result<_, AccountLoadError> and aborts the batch. It will be incorporated at the merge time.

sebatustra (approved):

PR #229, opened 2026-07-30
fix(indexer): retry the PendingRemint handoff and read back its state before escalating (finding-274)

sebatustra:
## Problem

  A permanently-failed withdrawal could lose its compensating remint two ways:

  1. **The handoff was not durable.** `handle_permanent_failure` discarded the remint info and
  release signatures, then called `set_pending_remint`. Any error wrote `ManualReview` and
  returned. If the write had actually committed and only the ack was lost, that overwrote a
  valid recovery state — the status writer accepts `pending_remint` as a source. And no sweep
  selects `manual_review`, so the row left every automatic path while blocking the dequeue
  frontier behind its nonce. A statement timeout was enough.

  2. **The finality gate read the wrong source.** Signatures are journaled *before* each send;
  the in-memory stash is only written after a successful one. So the attempt whose send errored
  ambiguously — the one most likely still in flight — was the one the gate never saw. It
  classified the earlier finalized-failed attempt, called the release dead, and reminted. The
  SMT gate doesn't catch it: `max_lvbh` comes from the same truncated set, so the non-inclusion
  proof runs while the missing signature can still land.

  ## Changes

  - `set_pending_remint` is idempotent for an identical payload, so a retry after a lost ack
  succeeds. A different payload is still rejected.
  - The write retries with `with_storage_backoff`; the compensation material stays in its
  locals.
  - On exhausted retries the row is read back and its status decides ownership: `PendingRemint`
  → this sender keeps driving it; `Processing` → restore the caches and leave it to recovery;
  anything else → another writer owns it; unreadable → escalate.
  - That escalation goes through a new `try_escalate_manual_review`, guarded on `status =
  'processing'`. A guard miss means the write did commit, so the row is left for the startup
  reload instead of being clobbered.
  - The gate now loads signatures with `load_pending_sigs`, the same journal read recovery
  uses, so the in-process and post-restart gates classify the same set. An unreadable journal
  restores the caches and leaves the row `Processing`.
  - Added `Storage::get_transaction_status`; `record_remint_result_internal` uses it instead of
  its own inline query.
  - Mock storage enforces the same guards as the SQL, which exposed several tests transitioning
  rows that never existed
### Coverage Report

| Core | 12,498 | 14,129 | 88.5% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30836960965) |
| Indexer | 32,942 | 36,441 | 90.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30836960965) |
| Gateway | 1,330 | 1,521 | 87.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30836960965) |
| Auth | 785 | 903 | 86.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30836960965) |
| E2E Integration | 13,714 | 17,436 | 78.7% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/30836960965) |
| **Total** | **61,269** | **70,430** | **87.0%** | |

> Last updated: 2026-08-03 18:34:45 UTC by E2E Integration

### DIFF
--- a/indexer/src/operator/recovery.rs
+++ b/indexer/src/operator/recovery.rs
@@ -419,7 +419,10 @@ async fn check_withdrawal(
 /// finality classifier. Shared by deposit and withdrawal recovery. A read error or a
 /// malformed stored signature returns a quarantine reason (uncertainty, never "dead"),
 /// so callers never demote a row whose signatures could not be read or parsed.
-async fn load_pending_sigs(storage: &Storage, id: i64) -> Result<Vec<PendingSig>, String> {
+pub(crate) async fn load_pending_sigs(
+    storage: &Storage,
+    id: i64,
+) -> Result<Vec<PendingSig>, String> {
     // Retry a transient DB blip before quarantining; matches the deposit gate so
     // a momentary read failure never pages ops for a healthy withdrawal.
     let stored = with_storage_backoff("journal read", id, || storage.get_release_signatures(id))
--- a/indexer/src/operator/sender/transaction.rs
+++ b/indexer/src/operator/sender/transaction.rs
@@ -8,6 +8,7 @@ use crate::operator::tree_constants::MAX_TREE_LEAVES;
 use crate::operator::utils::instruction_util::{
     mint_extra_error_checks_policy, TransactionBuilder,
 };
+use crate::operator::utils::storage_util::with_storage_backoff;
 use crate::operator::utils::transaction_util::parse_program_error;
 use crate::operator::utils::transaction_util::{
     build_and_sign, check_transaction_status, send_signed, ConfirmationResult,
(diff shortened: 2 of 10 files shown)


Huzaifa696:
When the PendingRemint write commits but the sender can't confirm it, handle_permanent_failure returns without queueing the in-memory entry, leaving the row pending_remint with no owner (startup-only rehydration, processing-only sweep) and no alert, so the user's remint stalls until a restart.

Fix: In the Ok(false) and Err arms at transaction.rs:1246/:1250, fall through and push to state.pending_remints instead of returning, matching the Ok(Some(PendingRemint)) adopt branch.

Huzaifa696 (approved):

PR #231, opened 2026-08-03
fix(core): bind settlement acknowledgements to executor account generations (issue-277)

Huzaifa696:
## Summary

Closes the stale-settlement issue. The settler acknowledged committed accounts back to BOB using only `{pubkey, account, deleted}`, without identifying which execution the acknowledgement belonged to. For updates, BOB could compare account bytes, but for deletions every closed account becomes the same tombstone, so stale acknowledgements were indistinguishable from current ones. As a result, an old delete acknowledgement could remove a newer tombstone from the cache, allowing `preload_accounts` to reload the deleted account from Postgres. This could resurrect a closed ATA and make its funds withdrawable a second time.

The fix adds a monotonic generation number to every account update. BOB stamps each write with a generation, the executor propagates it, and the settler acknowledges the highest durable generation for each tick. An entry is reconciled only if its generation is less than or equal to the acknowledged generation, making stale acknowledgements harmless for both updates and deletions. Byte equality is retained as a fail-closed consistency check: mismatches leave the entry dirty instead of reconciling it. Since generations are process-local, the change requires no schema, wire format, or deployment changes.

The implementation is intentionally narrow. Settlement APIs remain unchanged, `ExecutionResult` gains two generation fields, and the existing settlement pipeline is preserved. Tests cover the stale-acknowledgement regression, cache resurrection path, and the new reconciliation guards.
### Coverage Report

| Core | 13,574 | 15,249 | 89.0% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31002721747) |
| Indexer | 32,951 | 36,441 | 90.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31002721747) |
| Gateway | 1,330 | 1,521 | 87.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31002721747) |
| Auth | 785 | 903 | 86.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31002721747) |
| E2E Integration | 13,774 | 17,513 | 78.7% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31002721747) |
| **Total** | **62,414** | **71,627** | **87.1%** | |

> Last updated: 2026-08-05 12:40:40 UTC by E2E Integration

### DIFF
--- a/core/src/accounts/bob.rs
+++ b/core/src/accounts/bob.rs
@@ -28,8 +28,8 @@
 /// have their `synced_since` updated frequently, so this is a good heuristic to
 /// evict less frequently accessed accounts.
 ///
-/// What decides whether `synced_since` may be set at all is the second field,
-/// `generation`. Every account BOB writes is stamped with the ordinal of that
+/// What decides whether `synced_since` may be set at all is `generation`. Every
+/// account BOB writes is stamped with the ordinal of that
 /// write, and settlement feedback carries a high-water generation meaning
 /// "everything up to here is durable". An entry is only ever marked
 /// synchronized, or dropped if it is a tombstone, when its own generation is
@@ -81,6 +81,12 @@ pub struct BobCacheStats {
     pub bytes: usize,
     /// Entries evicted (age + cap) since the previous read; drained on read.
     pub evicted: usize,
+    /// Settlement acknowledgements whose generation was covered but whose bytes
+    /// disagreed with the in-memory entry, since the previous read; drained on
+    /// read. Always zero unless BOB's and the settler's independent derivations
+    /// of account state have diverged, so any non-zero value is a bug worth
+    /// alerting on rather than a condition to tune.
+    pub settlement_divergences: usize,
 }
 
 struct AccountWithMeta {
@@ -123,6 +129,8 @@ pub struct BOB {
     dirty_entries: usize,
     /// Running sum of resident account-data bytes; maintained like dirty_entries.
     cache_bytes: usize,
+    /// Byte divergences seen since the last cache_stats read; drained on read.
+    settlement_divergences: usize,
     /// Ordinal of the next executor write. Needs no synchronization: the
     /// executor is a single task owning BOB through `&mut self`, so the
     /// counter is strictly monotonic.
@@ -147,6 +155,7 @@ impl BOB {
             evicted_delta: 0,
             dirty_entries: 0,
             cache_bytes: 0,
+            settlement_divergences: 0,
             next_generation: 0,
         }
     }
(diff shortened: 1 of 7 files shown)


core/src/accounts/bob.rs
```diff
@@ -27,10 +27,20 @@
 /// older than `OLDEST_SYNCED_ACCOUNT_AGE` seconds. Generally, hot accounts will
 /// have their `synced_since` updated frequently, so this is a good heuristic to
 /// evict less frequently accessed accounts.
+///
+/// What decides whether `synced_since` may be set at all is the second field,
```
sebatustra: nit: generation is the fourth field in `AccountWithMeta`

core/src/accounts/bob.rs
```diff
@@ -653,6 +735,39 @@ mod tests {
             .as_secs()
     }
 
+    /// Single-transaction Executed output carrying `accounts` at the loaded
+    /// transaction's account positions, for driving `update_accounts` directly.
+    fn executed_output(
+        accounts: Vec<(Pubkey, AccountSharedData)>,
+    ) -> LoadAndExecuteSanitizedTransactionsOutput {
+        use solana_svm::{
```
sebatustra: nit: should we avoid use statements inside function bodies?

core/src/accounts/bob.rs
```diff
@@ -271,39 +307,73 @@ impl BOB {
...
+                            if was_dirty {
+                                self.dirty_entries = self.dirty_entries.saturating_sub(1);
+                            }
+                        } else {
+                            // A covered generation must imply matching bytes, so a
+                            // mismatch is a bug: BOB's and the settler's independent
+                            // derivations of account state have diverged. Leave the
+                            // entry dirty, which cannot be evicted or reloaded.
+                            let entry_generation = account.generation;
+                            warn!(
```
sebatustra, line 395: should this divergence have a metric? or is a `warn!` enough?

core/src/accounts/bob.rs
Huzaifa696: fixed, dropped the ordinal

core/src/accounts/bob.rs
Huzaifa696: Removed both, the test mod already imports them.

core/src/accounts/bob.rs
Huzaifa696: Added a counter drained via cache_stats + a test.

PR #233, opened 2026-08-05
fix(deploy): separate the escrow instance admin from the channel admin key

sebatustra:
## What

Splits the two admins that the shipped deployment collapsed into one keypair.

  - Escrow `Instance.admin` is now its own keypair (`keypairs/escrow-admin.json`). It signs `CreateInstance`, `AllowMint`, `AddOperator` and `SetNewAdmin`, and no service ever loads it.
  - The channel admin (`ADMIN_PRIVATE_KEY` / `PRIVATE_CHANNEL_ADMIN_KEYS`) stays the hot service identity: receipt-mint authority, operator fee payer, and the registered escrow Operator.
  - `devnet-test.sh` fails closed if the two pubkeys match. Env files, ansible vars, compose and the CLI help text now name which key is which.

## Why this instead of the proposed changes

The finding recommends `pending_admin`, an admin epoch, and binding releases on-chain to a registered mint authority. None of that is needed to close it. The vulnerability is entirely that the key `SetNewAdmin` rotates was also a service key. Once the escrow admin is separate, it never becomes an SPL mint authority and never gets an `AddOperator` registration, so both drain paths (atomic mint-and-burn, and direct `ReleaseFunds` via a surviving Operator PDA) are gone. The indexer never signs an escrow admin instruction, so this is config only, no program change.

The two-phase transition and epoch binding remain reasonable hardening. Not in this PR.

## Docs

`SetNewAdmin` no longer claims the old admin loses all privileges. It rewrites one field: Operator PDAs survive until `RemoveOperator`, and a shared key keeps mint and freeze authority. The reconciliation halt runbook now requires SPL `SetAuthority` migration before a channel admin rotation.
### Coverage Report

| Core | - | - | - | - |
| Indexer | - | - | - | - |
| Gateway | - | - | - | - |
| Auth | - | - | - | - |

### DIFF
--- a/.env.devnet
+++ b/.env.devnet
@@ -3,6 +3,7 @@
 
 # Instance
 ESCROW_INSTANCE_ID=
+# Channel admin service key, not the escrow instance admin. See .env.example.
 # Keep blank here: `make build-devnet` writes the live key to the gitignored `.env`.
 ADMIN_PRIVATE_KEY=
 
--- a/.env.example
+++ b/.env.example
@@ -7,6 +7,15 @@ ESCROW_INSTANCE_ID=instance_pubkey
 # Created post-deploy via `create_instance`; compose interpolation needs it set,
 # so default to the all-1s placeholder until the real PDA is written back.
 COMMON_ESCROW_INSTANCE_ID=11111111111111111111111111111111
+# The channel admin service key: SPL mint+freeze authority on every receipt mint,
+# and the fee payer for the operator's own transactions.
+#
+# This is NOT the escrow instance admin. That is a separate offline keypair
+# (keypairs/escrow-admin.json) which no service ever loads, so it has no env var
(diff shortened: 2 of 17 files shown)


Huzaifa696 (commented):
devnet-test.sh's new same-key guard resolves the channel admin from $ADMIN_KEYPAIR (default ./keypairs/admin.json, a file no make target creates), but the containers load ADMIN_PRIVATE_KEY from .env (written from keypairs/operator-keypair.json), so the check passes even when .env holds the escrow admin key, and step 2 registers an Operator pubkey the containers never sign with.

Fix: Derive the channel admin pubkey from the already-sourced $ADMIN_PRIVATE_KEY and use that one value for both the collision check and the AddOperator registration.

Huzaifa696 (approved):

PR #238, opened 2026-08-11
fix(indexer): bound the withdrawal halt sweep to the poison nonce (issue-15)

Huzaifa696:
## Summary

The withdrawal halt previously moved every active withdrawal to `manual_review`, including rows below the poison nonce that may already be signed or broadcast. Sweeping one of these rows could prevent its completion from updating the database, causing the nonce to disappear from the local tree and leading to `SmtRootMismatch` on the next boot. `quarantine_active_withdrawals` now takes a `min_nonce` and only quarantines rows at or above the poison nonce; older rows remain available to the existing recovery workers.

The halt's previous justification was incorrect: the program has no cross-nonce check and the tree is sparse. The actual concern is tree rotation, once the tree moves past a quarantined row's generation, that nonce is permanently rejected. Comments and the runbook now reflect this. A skipped `Completed` status now logs at `error!` and increments a counter; other statuses remain INFO because they can occur during normal recovery.

Seven unit tests and `drill_18` cover the nonce floor against real Postgres. A NULL poison nonce preserves the existing unbounded behavior. 
### Coverage Report

| Core | 12,065 | 13,690 | 88.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31713153586) |
| Indexer | 24,220 | 27,395 | 88.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31713153586) |
| Gateway | 1,055 | 1,230 | 85.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31713153586) |
| Auth | 830 | 949 | 87.5% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31713153586) |
| E2E Integration | 12,581 | 15,663 | 80.3% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31713153586) |
| **Total** | **50,751** | **58,927** | **86.1%** | |

> Last updated: 2026-08-13 15:34:05 UTC by E2E Integration


indexer/src/operator/processor.rs
```diff
@@ -225,15 +223,22 @@ async fn halt_withdrawal_pipeline(
...
-    // Sweep the rest of the pipeline: any row still `Pending` (never
-    // fetched) or `Processing` (locked but unsent, e.g. a sibling was mid-
-    // flight in another instance) is flipped to `ManualReview`.
-    match storage.quarantine_all_active_withdrawals(poison_id).await {
+    // Sweep the rest of the pipeline: any row at or above the poison's nonce
+    // still `Pending` (never fetched), `Processing` (locked but unsent) or
+    // `Parked` is flipped to `ManualReview`. A poison row with no nonce
+    // yields no floor, so the sweep stays unbounded.
+    let poison_id = poison.map(|txn| txn.id);
+    let min_nonce = poison.and_then(|txn| txn.withdrawal_nonce);
```
sebatustra, line 231: could min_nonce be too low? 
the tree only gets swapped on nonces that are multiples of 65,536. In that case the only rows we need to block are the ones at or above the next multiple of 65,536, not every row above the bad one

indexer/src/storage/postgres/db.rs
```diff
@@ -1637,9 +1644,11 @@ impl PostgresDb {
             WHERE transaction_type = 'withdrawal'
               AND status IN ('pending', 'processing', 'parked')
```
sebatustra, line 1776: could rows that are processing already been sent to the chain? what is the impact of killing rows that are mid-flight?

indexer/src/operator/processor.rs
Huzaifa696: I kept the floor at the poison nonce because of the nonce frontier landing separately (#185). Once that's in, a manual_review row blocks every higher nonce from being dequeued. So those gap rows would be blocked anyway, just without a webhook: stuck in pending with nobody paged. I'd rather over-quarantine loudly than stall silently.

indexer/src/storage/postgres/db.rs
Huzaifa696: Yes, a processing row we sweep can already be on-chain.

The impact, in order:

1. The sweep flips the row to `manual_review`.
2. The release lands and the sender emits `Completed`.
3. That write only accepts `processing`/`pending_remint`, so it matches nothing and is silently dropped.
4. Recovery only scans `processing`, so it never revisits the row.
5. On the next boot, the local tree is rebuilt from `completed` rows, the leaf is missing, and startup fails with a root mismatch.

The nonce floor fixes this for rows below the poison nonce, which is where in-flight rows normally are. It is not completely airtight: a whole insert batch can share the same `created_at`, and the dequeue has no nonce tiebreaker, so an in-flight row can occasionally sit above the poison nonce and still be swept. Sorting the queue by nonce would close this gap and belongs with the frontier work #185 .  #239  also reconciles `manual_review` rows that still have a release signature, making step 5 recoverable in those cases.

sebatustra (approved):

PR #242, opened 2026-08-12
fix(indexer): gate the signatureless withdrawal quarantine behind an on-chain proof (issue-262)

Huzaifa696:
## Summary

Closes issue 262, the remaining finding after #209. A withdrawal burns on the PrivateChannel before the indexer creates its release, leaving the row in `processing`. #209 only requeues a stranded row when the original DB failure recovers quickly; a longer outage leaves it stuck, after which recovery can quarantine it for having no signature and block all higher nonces.

Recovery now re-arms these rows only after checking the on-chain SMT. If the withdrawal is proven absent, the row returns to `pending`; if it is proven present, it is quarantined as an invariant violation; if the proof cannot be read, the row is left untouched until it becomes stale. To make this safe against a live sender, the release path now uses the same atomic claim-and-persist logic as deposits. If the sender loses the claim, it rolls back its local SMT and exits quietly; recovery uses the same `updated_at` CAS, so only one side can win.

The change is limited to the release pre-broadcast path and one `check_withdrawal` branch. Tests cover the claim race, sender outcomes, lease refresh, and all recovery verdicts, including Postgres-backed race tests.
### Coverage Report

| Core | 13,783 | 15,483 | 89.0% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31796884355) |
| Indexer | 34,890 | 38,439 | 90.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31796884355) |
| Gateway | 1,476 | 1,698 | 86.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31796884355) |
| Auth | 908 | 1,059 | 85.7% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31796884355) |
| E2E Integration | 14,146 | 17,994 | 78.6% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31796884355) |
| **Total** | **65,203** | **74,673** | **87.3%** | |

> Last updated: 2026-08-14 12:41:19 UTC by E2E Integration


indexer/src/operator/recovery.rs
sebatustra, line 373: I think a DB read failure here still quarantines instantly, should we use the same escalation window used for `verify_release_landed`?

Huzaifa696: Fixed by splitting that one error into two cases:

- The read failed:, so we have no idea what's in the journal. That now waits and retries each sweep, escalating only after 10 minutes, the same way we already handle an unreadable on-chain proof.
- The read worked: but a stored signature is garbage. That still escalates immediately: retrying only returns the same bad bytes, and the fact that a signature exists at all means the release may already have gone out.

sebatustra (approved):

PR #243, opened 2026-08-12
fix(indexer): bind the channel absence verdict to one status response (issue-15)

Huzaifa696:
### Problem

`AccountsDB::get_transaction` and `get_block` treated Postgres, bincode, and Redis errors as `None`. This made internal failures look like missing data: `getSignatureStatuses` returned `null`, `getBlock` could silently drop unreadable transactions, and an unreadable block looked the same as an empty slot. Malformed signatures also returned `null` instead of an invalid-parameter error. The operator additionally combined the status response with a separate `getBlockHeight` call, allowing different backends to decide whether a MintTo could be replayed.

### Fix

`get_transaction`, `get_block`, and `get_block_time` now return `Result<Option<_>>`, so storage and decode failures surface as `-32000` while `null` means the data is genuinely absent. Malformed signatures now return `-32602`, matching Agave. `FinalityRpc` also uses an explicit `HeightSource`: the channel derives the height from the same response's `context.slot`, while Solana keeps `getBlockHeight`. The retention window is now derived from the channel node's configured `max_blockhashes` at startup and takes the maximum across both channel endpoints, failing closed if it cannot be read.

Tests cover error-vs-absence handling, batch semantics, height sourcing, the derived retention window, and the startup gate.
### Coverage Report

| Core | 14,008 | 15,713 | 89.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31820899977) |
| Indexer | 35,494 | 39,083 | 90.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31820899977) |
| Gateway | 1,476 | 1,698 | 86.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31820899977) |
| Auth | 908 | 1,059 | 85.7% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31820899977) |
| E2E Integration | 14,287 | 18,217 | 78.4% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31820899977) |
| **Total** | **66,173** | **75,770** | **87.3%** | |

> Last updated: 2026-08-14 17:28:35 UTC by E2E Integration


indexer/src/operator/operator.rs
```diff
@@ -491,6 +517,56 @@ async fn run_withdraw_preflight(
...
+) -> Result<u64, OperatorError> {
+    let (context_slot, last_valid_block_height) = channel_rpc
+        .get_latest_blockhash_with_context(CommitmentConfig::finalized())
+        .await
+        .map_err(|e| {
+            OperatorError::RpcError(format!(
+                "could not read the channel endpoint's blockhash window at startup: {e}"
+            ))
+        })?;
+    Ok(last_valid_block_height.saturating_sub(context_slot))
```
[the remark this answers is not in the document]
Huzaifa696: Fixed: the window is re-derived per verdict and maxed with the startup seed; a failed read is Uncertain.

indexer/src/operator/sender/remint.rs
```diff
@@ -425,24 +477,68 @@ enum EndpointVerdict {
...
+    match finality.height_source {
+        HeightSource::BlockHeightRpc => Ok(finality.blockhash_window),
+        HeightSource::ContextSlot => {
+            let (context_slot, last_valid_block_height) = rpc
+                .get_latest_blockhash_with_context(CommitmentConfig::finalized())
+                .await
+                .map_err(|e| e.to_string())?;
+            Ok(last_valid_block_height
+                .saturating_sub(context_slot)
+                .max(finality.blockhash_window))
```
sebatustra: I believe this .max() only guard the direction where the window grew. what happens if max_blockhashes is reduced and the operator restarts while a PendingRemint row from before the change is still stored?

Huzaifa696: Fixed by giving each attempt its own bound instead of patching the window: build_and_sign now records the slot its blockhash was read at, journaled next to the signature as blockhash_slot. A transaction cannot land in a block older than its own blockhash, so that slot is the exact earliest block the signature could be in, fixed at broadcast. max_blockhashes can now move either way without affecting work already in flight.

sebatustra (approved):

PR #245, opened 2026-08-12
fix: Recover the Yellowstone stream from silent stalls

gitteri:
The escrow indexer wedged on devnet for ~12h. The Yellowstone gRPC stream went silent with no FIN, no error, and no server ping, so stream.next() blocked forever and nothing tripped the existing reconnect path. The health endpoint correctly reported degraded/stalled, but Docker does not auto-restart on an unhealthy check, so it just sat there until a manual restart (which reconnects and gap-fills the backlog on its own).

cantina-audit already had the improved detection (require_continuous_progress makes /health report stalled even at pending=0) but no recovery. This adds the recovery.

Two independent backstops in connect_and_stream:

HTTP/2 keepalive on the gRPC client (15s interval, 10s timeout, while-idle) so a dead transport surfaces as a stream error and takes the existing reconnect path.

An application-level watchdog: a select arm that forces a reconnect if no message of any kind (a block or a server ping) arrives within a window. Fresh timer each loop iteration so any inbound message resets it. It fires only on a fully silent stream and returns Err into the existing reconnect + gap-fill path, which replays whatever slots were missed while wedged. Default is 60s, comfortably above the server ping cadence so a healthy idle stream is never torn down (escrow blocks are sparse since we subscribe to blocks filtered by account_include, not every-slot blocks_meta).

The stall window is injectable via with_stall_timeout so the regression test drives the reconnect without a 60s wait. New test stall_watchdog_forces_reconnect_on_silent_stream reproduces the exact failure: the mock holds the stream open and silent, and we assert a resubscribe happens. New metric label INDEXER_RPC_ERRORS{error_type="stall"}.
### Coverage Report

| Core | 14,008 | 15,713 | 89.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32067696247) |
| Indexer | 35,493 | 39,103 | 90.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32067696247) |
| Gateway | 1,476 | 1,698 | 86.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32067696247) |
| Auth | 908 | 1,059 | 85.7% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32067696247) |
| E2E Integration | 14,334 | 18,237 | 78.6% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32067696247) |
| **Total** | **66,219** | **75,810** | **87.3%** | |

> Last updated: 2026-08-17 21:41:07 UTC by E2E Integration

### DIFF
--- a/indexer/src/indexer/datasource/yellowstone/source.rs
+++ b/indexer/src/indexer/datasource/yellowstone/source.rs
@@ -54,6 +54,9 @@ pub struct YellowstoneSource {
     #[cfg(feature = "datasource-rpc")]
     storage: Option<Arc<Storage>>,
     health: Option<Arc<private_channel_metrics::HealthState>>,
+    /// Silent-stream watchdog window. Defaults to STREAM_STALL_TIMEOUT; overridable
+    /// (mainly so tests can drive the reconnect path without a 60s wait).
+    stall_timeout: std::time::Duration,
 }
 
 impl YellowstoneSource {
@@ -79,11 +82,18 @@ impl YellowstoneSource {
             #[cfg(feature = "datasource-rpc")]
             storage: None,
             health: None,
+            stall_timeout: STREAM_STALL_TIMEOUT,
         }
     }
 
     pub fn with_health(mut self, health: Arc<private_channel_metrics::HealthState>) -> Self {
         self.health = Some(health);
+        self
+    }
+
+    /// Override the silent-stream watchdog window (mainly for tests).
+    pub fn with_stall_timeout(mut self, stall_timeout: std::time::Duration) -> Self {
+        self.stall_timeout = stall_timeout;
         self
     }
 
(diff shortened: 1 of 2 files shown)


indexer/src/indexer/datasource/yellowstone/source.rs
```diff
@@ -54,6 +54,9 @@ pub struct YellowstoneSource {
     #[cfg(feature = "datasource-rpc")]
     storage: Option<Arc<Storage>>,
     health: Option<Arc<private_channel_metrics::HealthState>>,
+    /// Silent-stream watchdog window. Defaults to STREAM_STALL_TIMEOUT; overridable
+    /// (mainly so tests can drive the reconnect path without a 60s wait).
```
sebatustra, line 58: STREAM_STALL_TIMEOUT is set to 120s, not 60s

sebatustra (approved):
LGTM

indexer/src/indexer/datasource/yellowstone/source.rs
gitteri: Good catch, fixed in a0ca6f1: comment now says 120s to match STREAM_STALL_TIMEOUT.

PR #248, opened 2026-08-12
feat(deploy): Optional Doppler secrets backend, opt-in per operator

gitteri:
Lets an individual operator pull deploy secrets from Doppler instead of the plaintext secrets.yml, without changing anything for everyone else on the repo.

How it works: a secrets_backend toggle in the Ansible deploy. Default is file (the gitignored secrets.yml), so existing users are byte-for-byte unchanged. When DOPPLER_TOKEN is set in your shell, the playbook auto-selects the doppler backend, runs doppler secrets download at deploy time, and maps the keys onto the same var names the file backend provides. Everything downstream (env.j2 render, compose) is identical. Force either way with -e secrets_backend=file or -e secrets_backend=doppler.

Kept it opt-in and per-operator on purpose: the toggle lives in your environment (the token), not in a committed vars file, so nobody deploying with -l dev inherits it. The service token pins the Doppler project and config, so the playbook needs no extra flags. secrets.yml stays as a fallback: unset the token and the deploy reads the file again.

Scope note: this is deploy-time only (control node pulls, renders .env as today). Runtime injection on the VM (no .env on disk) is a separate future step. Also left the RPC token riding in the URL path as-is, worth moving to header auth with the provider separately.

README documents the setup and a one-liner to import an existing secrets.yml into Doppler with the keys uppercased.
### Coverage Report

| Core | - | - | - | - |
| Indexer | - | - | - | - |
| Gateway | - | - | - | - |
| Auth | - | - | - | - |

### DIFF
--- a/private-channel-deploy/README.md
+++ b/private-channel-deploy/README.md
@@ -51,7 +51,7 @@ The deploy reads secrets from one of two backends, chosen automatically:
 
 Doppler setup (one time):
 
-1. Create a Doppler project and config, and add the `secrets.yml` keys as secrets. Doppler names are the UPPERCASE of the file keys: `POSTGRES_PASSWORD`, `POSTGRES_REPLICATION_PASSWORD`, `JWT_SECRET`, `GRAFANA_ADMIN_PASSWORD`, `SOLANA_KEYPAIR`, `GHCR_USER`, `GHCR_TOKEN`, `YELLOWSTONE_TOKEN`, `RPC_URL`. Two use their rendered `.env` names instead: `DEVNET_YELLOWSTONE_ENDPOINT` (optional; overrides `yellowstone_endpoint` from `vars/<env>.yml`) and `ALERT_WEBHOOK_URL` (the webhook; `NOTIFY_WEBHOOK` is also accepted).
+1. Create a Doppler project and config, and add the `secrets.yml` keys as secrets. Doppler names are the UPPERCASE of the file keys: `POSTGRES_PASSWORD`, `POSTGRES_REPLICATION_PASSWORD`, `JWT_SECRET`, `GRAFANA_ADMIN_PASSWORD`, `GHCR_USER`, `GHCR_TOKEN`, `YELLOWSTONE_TOKEN`, `RPC_URL`. Two use their rendered `.env` names instead: `DEVNET_YELLOWSTONE_ENDPOINT` (optional; overrides `yellowstone_endpoint` from `vars/<env>.yml`) and `ALERT_WEBHOOK_URL` (the webhook; `NOTIFY_WEBHOOK` is also accepted).
 2. Import your existing file in one shot (uppercases the keys, keeps values):
    ```bash
    # from private-channel-deploy/, with the doppler CLI installed + `doppler setup` done.
--- a/private-channel-deploy/README.md
+++ b/private-channel-deploy/README.md
@@ -54,9 +54,13 @@ Doppler setup (one time):
 1. Create a Doppler project and config, and add the `secrets.yml` keys as secrets. Doppler names are the UPPERCASE of the file keys: `POSTGRES_PASSWORD`, `POSTGRES_REPLICATION_PASSWORD`, `JWT_SECRET`, `GRAFANA_ADMIN_PASSWORD`, `SOLANA_KEYPAIR`, `GHCR_USER`, `GHCR_TOKEN`, `YELLOWSTONE_TOKEN`, `RPC_URL`, `NOTIFY_WEBHOOK`.
 2. Import your existing file in one shot (uppercases the keys, keeps values):
    ```bash
-   # from private-channel-deploy/, with the doppler CLI installed + `doppler setup` done
-   uv run --with pyyaml python3 -c "import yaml,json,sys; d=yaml.safe_load(open('secrets.yml')); json.dump({k.upper():('' if v is None else str(v)) for k,v in d.items()}, sys.stdout)" > /tmp/spc-secrets.json
-   doppler secrets upload /tmp/spc-secrets.json && rm -f /tmp/spc-secrets.json
+   # from private-channel-deploy/, with the doppler CLI installed + `doppler setup` done.
+   # Subshell keeps the plaintext in a private 0700 tempdir and wipes it on any exit
+   # (including failure or Ctrl-C), so credentials never linger in a world-readable /tmp.
+   ( set -euo pipefail
+     dir="$(mktemp -d)"; trap 'rm -rf "$dir"' EXIT
+     uv run --with pyyaml python3 -c "import yaml,json,sys; d=yaml.safe_load(open('secrets.yml')); json.dump({k.upper():('' if v is None else str(v)) for k,v in d.items()}, sys.stdout)" > "$dir/secrets.json"
+     doppler secrets upload "$dir/secrets.json" )
    ```
 3. Mint a **read-only** service token scoped to that config, then run deploys with it in the environment:
    ```bash
--- a/private-channel-deploy/README.md
+++ b/private-channel-deploy/README.md
@@ -32,13 +32,13 @@ Three one-time edits per environment, done before the first deploy. They tell th
 
 1. **Point at a host** ([`inventory.ini`](./inventory.ini)): IP, SSH user, key path.
 2. **Choose what to deploy** ([`vars/dev.yml`](./vars/dev.yml)): image registry/tag, target Solana network, on-host paths. Each option has an inline comment explaining what to pick.
-3. **Provide credentials**: copy the template, then fill in real values:
+3. **Provide credentials**, via one of the two backends below. For the default file backend, copy the template and fill in real values:
 
    ```bash
    cp secrets.yml.example secrets.yml
    ```
 
-   `secrets.yml` is plain text and gitignored. See the file header for required vs optional and how to generate each value.
+   `secrets.yml` is plain text and gitignored. See the file header for required vs optional and how to generate each value. Doppler-only operators skip this file entirely — preflight does not require it when the doppler backend is selected.
 
 Everything else is wired and shouldn't need routine changes.
 
--- a/private-channel-deploy/README.md
+++ b/private-channel-deploy/README.md
@@ -51,7 +51,7 @@ The deploy reads secrets from one of two backends, chosen automatically:
 
 Doppler setup (one time):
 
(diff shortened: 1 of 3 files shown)


private-channel-deploy/README.md
```diff
@@ -40,7 +40,31 @@ Three one-time edits per environment, done before the first deploy. They tell th
...
+- **doppler** (opt-in, per-operator): pulls the same keys from Doppler at deploy time. Auto-selected when `DOPPLER_TOKEN` is set in your shell; otherwise the file backend is used. Force either with `-e secrets_backend=file|doppler`.
+
+Doppler setup (one time):
+
+1. Create a Doppler project and config, and add the `secrets.yml` keys as secrets. Doppler names are the UPPERCASE of the file keys: `POSTGRES_PASSWORD`, `POSTGRES_REPLICATION_PASSWORD`, `JWT_SECRET`, `GRAFANA_ADMIN_PASSWORD`, `SOLANA_KEYPAIR`, `GHCR_USER`, `GHCR_TOKEN`, `YELLOWSTONE_TOKEN`, `RPC_URL`, `NOTIFY_WEBHOOK`.
+2. Import your existing file in one shot (uppercases the keys, keeps values):
+   ```bash
+   # from private-channel-deploy/, with the doppler CLI installed + `doppler setup` done
+   uv run --with pyyaml python3 -c "import yaml,json,sys; d=yaml.safe_load(open('secrets.yml')); json.dump({k.upper():('' if v is None else str(v)) for k,v in d.items()}, sys.stdout)" > /tmp/spc-secrets.json
+   doppler secrets upload /tmp/spc-secrets.json && rm -f /tmp/spc-secrets.json
```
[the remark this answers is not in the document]
gitteri: Fixed. Wrapped the import in a subshell that writes to a private `mktemp -d` tempdir (0700) with `trap 'rm -rf' EXIT`, so the plaintext is unpredictable-path, owner-only, and wiped on any exit including a failed upload or Ctrl-C. Went with `mktemp -d` instead of a suffixed template since BSD/macOS mktemp does not accept chars after the X's, and the operators run this on macOS.

private-channel-deploy/deploy.yml
```diff
@@ -13,13 +13,78 @@
...
+      changed_when: false
+      no_log: true
+
+    - name: Map Doppler secrets onto deploy vars
+      ansible.builtin.set_fact:
+        postgres_password: "{{ _d.POSTGRES_PASSWORD | default('') }}"
+        postgres_replication_password: "{{ _d.POSTGRES_REPLICATION_PASSWORD | default('') }}"
+        jwt_secret: "{{ _d.JWT_SECRET | default('') }}"
+        grafana_admin_password: "{{ _d.GRAFANA_ADMIN_PASSWORD | default('') }}"
+        solana_keypair: "{{ _d.SOLANA_KEYPAIR | default('') }}"
```
Huzaifa696: SOLANA_KEYPAIR maps solana_keypair and README.md tells operators to put a funded Solana private key in Doppler but nothing reads that var. The admin key still comes from admin_keypair_path on disk (vars/dev.yml:79), derived at deploy.yml:106-165 into ADMIN_PRIVATE_KEY via env.j2:58. So a missing local file still hard-fails with "run solana-keygen." It's also absent from secrets.yml.example, so the backends aren't the mirror the PR claims.

Likely intent: centralizing the keypair was meant, but never wired up.

Fix

Delete deploy.yml:75 and drop SOLANA_KEYPAIR from the README list.

If centralizing it was the goal, do it deliberately in a follow-up: write the fetched bytes to admin_keypair_path (mode: 0600, no_log: true) before the stat at deploy.yml:106, define precedence when both sources have a value, and name Doppler in the assert's fail_msg.

gitteri: Good catch, fixed. Removed the solana_keypair map line and dropped SOLANA_KEYPAIR from the README list. You're right that nothing consumes it, the admin key comes from admin_keypair_path on disk, derived into ADMIN_PRIVATE_KEY, so mapping it was misleading (and it was never in secrets.yml.example either). I mirrored it from the secrets.yml key set without checking it was actually read; it's a pre-existing dead key there too, but no reason to propagate it into the Doppler backend. If we do want to centralize the keypair, I'll do it deliberately in a follow-up per your outline (write the fetched bytes to admin_keypair_path 0600/no_log before the stat, define precedence when both sources are set, and name Doppler in the assert fail_msg).

private-channel-deploy/README.md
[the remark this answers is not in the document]
gitteri: Addressed in 208ebaa4c: the import snippet now runs in a subshell with set -euo pipefail, writes the plaintext into a private 0700 mktemp -d dir, and wipes it via trap ... EXIT (covers failure and Ctrl-C). Using a 0700 tempdir rather than umask 077 means the file is never world-readable even briefly.

private-channel-deploy/deploy.yml
gitteri: Good catch, fixed in 8b7a3e271: dropped the dead SOLANA_KEYPAIR -> solana_keypair mapping from deploy.yml and removed it from the README key list. You're right that nothing read it (the admin key still comes from admin_keypair_path on disk), so it was a misleading no-op rather than a real mirror. Centralizing the keypair into Doppler is left as a deliberate follow-up rather than a half-wired var.

private-channel-deploy/deploy.yml
```diff
@@ -13,13 +13,77 @@
...
+      msg: "Secrets backend: {{ secrets_backend }}"
+    tags: [ always ]
+    run_once: true
+    delegate_to: localhost
+
+  # File backend: gitignored plain-text secrets.yml (see .gitignore).
+  - name: Load plain-text secrets (file backend)
     ansible.builtin.include_vars:
       file: "{{ playbook_dir }}/secrets.yml"
+    when: secrets_backend == 'file'
```
Huzaifa696, line 40: Preflight unconditionally requires `secrets.yml`, so Doppler-only deployments fail despite being the documented flow.

**Fix.** Gate `secrets.yml` on `secrets_backend` and remove the unconditional `cp secrets.yml.example secrets.yml` from README §First run.

private-channel-deploy/README.md
```diff
@@ -40,7 +40,35 @@ Three one-time edits per environment, done before the first deploy. They tell th
...
+### Secrets backends
+
+The deploy reads secrets from one of two backends, chosen automatically:
+
+- **file** (default): the gitignored `secrets.yml` above. No change for existing users.
+- **doppler** (opt-in, per-operator): pulls the same keys from Doppler at deploy time. Auto-selected when `DOPPLER_TOKEN` is set in your shell; otherwise the file backend is used. Force either with `-e secrets_backend=file|doppler`.
+
+Doppler setup (one time):
+
+1. Create a Doppler project and config, and add the `secrets.yml` keys as secrets. Doppler names are the UPPERCASE of the file keys: `POSTGRES_PASSWORD`, `POSTGRES_REPLICATION_PASSWORD`, `JWT_SECRET`, `GRAFANA_ADMIN_PASSWORD`, `GHCR_USER`, `GHCR_TOKEN`, `YELLOWSTONE_TOKEN`, `RPC_URL`. Two use their rendered `.env` names instead: `DEVNET_YELLOWSTONE_ENDPOINT` (optional; overrides `yellowstone_endpoint` from `vars/<env>.yml`) and `ALERT_WEBHOOK_URL` (the webhook; `NOTIFY_WEBHOOK` is also accepted).
```
Huzaifa696, line 54: `RPC_URL` is now a managed secret but is still printed in cleartext by `_report`, including to stdout, `./ansible.log`, and `notify_webhook`. Since paid RPC URLs may contain API keys, every deploy can leak credentials off-box.

**Fix.** Redact `RPC_URL` to scheme + host, mark report tasks `no_log`, or remove `rpc_url` from the mapping.

private-channel-deploy/deploy.yml
```diff
@@ -13,13 +13,77 @@
...
+        postgres_password: "{{ _d.POSTGRES_PASSWORD | default('') }}"
+        postgres_replication_password: "{{ _d.POSTGRES_REPLICATION_PASSWORD | default('') }}"
+        jwt_secret: "{{ _d.JWT_SECRET | default('') }}"
+        grafana_admin_password: "{{ _d.GRAFANA_ADMIN_PASSWORD | default('') }}"
+        ghcr_user: "{{ _d.GHCR_USER | default('') }}"
+        ghcr_token: "{{ _d.GHCR_TOKEN | default('') }}"
+        yellowstone_token: "{{ _d.YELLOWSTONE_TOKEN | default('') }}"
+        # yellowstone_endpoint normally comes from vars/<env>.yml; only override it
+        # when the Doppler config carries DEVNET_YELLOWSTONE_ENDPOINT (its .env name).
+        yellowstone_endpoint: "{{ _d.DEVNET_YELLOWSTONE_ENDPOINT | default(yellowstone_endpoint | default('')) }}"
```
Huzaifa696: **Nit** `| default(x)` only applies to undefined values, so an empty Doppler value can overwrite a valid `rpc_url` / `yellowstone_endpoint` from `vars/<env>.yml`.

Use the boolean form: `| default(rpc_url | default(''), true)`.

private-channel-deploy/deploy.yml
gitteri: Fixed in 6ab55e311: preflight's required-files loop now includes secrets.yml only when secrets_backend == file, so Doppler-only deploys pass (a short Doppler config still fails on the per-secret asserts with the exact missing key). The README first run step now says the cp applies to the file backend and Doppler operators skip the file entirely.

private-channel-deploy/README.md
gitteri: Fixed in 6ab55e311: the report now renders rpc_url as scheme plus host (port kept, path/query redacted), so an api-key bearing URL never reaches stdout, ansible.log or the notify webhook. Went with redaction over no_log so the report stays printable, and over dropping the mapping since centralizing RPC_URL in Doppler is the point of the backend.

private-channel-deploy/deploy.yml
gitteri: Fixed in 6ab55e311: both overrides use the boolean form now (default(x, true)), and I applied the same to the notify_webhook fallback chain so an empty ALERT_WEBHOOK_URL falls through to NOTIFY_WEBHOOK. Verified locally that an empty Doppler value falls back to the vars file and a set one still wins.

Huzaifa696 (approved):

PR #250, opened 2026-08-13
refactor(indexer): port the withdrawal operator off the SMT to the nonce bitmap

Huzaifa696:
## Summary

The escrow program replaced its withdrawal Sparse Merkle Tree with an 8 KB on-chain nonce bitmap in #244. This PR removes the corresponding off-chain tree: the local tree, proof generation, rollback and rebuild paths are deleted, and 544 bytes of proof data leave every release. The chain is now the only replay authority.

## Behavioral changes

Same stages, same failure routing, same recovery machinery. What changed is where the truth comes from.

- **`NonceAlreadyUsed` is terminal, where `InvalidSmtProof` was retryable.** The old error meant either a stale local tree or a genuine second release, and the operator assumed the first and retried with a new proof. `NonceAlreadyUsed` can only mean the release already landed, so the withdrawal is settled from its recorded signatures instead of retried.
- **A cached generation gates the build, asymmetrically.** The operator remembers the current generation and checks it before building. It believes that memory only when it says go: anything that would hold a release back is confirmed against the chain first, because a release wrongly held would never be sent, never rejected, and so nothing would ever correct the memory that held it.
- **The refund path reads the nonce bit before crediting.** Before crediting a user back, the operator reads that nonce's bit. If the read fails, or the bitmap has already moved past that nonce, the refund waits instead of going ahead on the signature verdict alone.
- **Per-nonce bits decouple withdrawals that a shared root entangled.** A Merkle root is a single value computed from every nonce, so one unresolved nonce made the whole tree unknowable and stalled every other withdrawal in it. Bits are independent, so nothing is shared to get stuck on, and a delayed release can be resent unchanged.
- **Boot reconciliation is a directional bit diff, not a root equality check.** If the chain released a nonce the database missed, the operator repairs the row and carries on. If the database claims a release the chain never made, it still refuses to start. Either way it names the exact nonces.
- **Rotation blocks on deferred refunds in the generation being cleared.** Wiping the bits would erase the only record of whether a user had already been paid, so a rotation is held until every refund depending on those bits has settled.
- **Queued releases and escalated rows are durable, not in-memory.** A release waiting for the next rotation is written to the database as `Parked` rather than held only in memory, so a crash rebuilds it instead of paging someone. Escalating an uncertain withdrawal also no longer deletes its broadcast signatures, which are exactly what anyone would need to settle it later.
### Coverage Report

| Core | 12,065 | 13,690 | 88.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31842363712) |
| Indexer | 25,765 | 29,038 | 88.7% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31842363712) |
| Gateway | 1,055 | 1,230 | 85.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31842363712) |
| Auth | 830 | 949 | 87.5% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31842363712) |
| E2E Integration | 12,521 | 16,058 | 78.0% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/31842363712) |
| **Total** | **52,236** | **60,965** | **85.7%** | |

> Last updated: 2026-08-14 22:11:16 UTC by Auth


admin-ui/src/components/OperatorFunctions.tsx
```diff
@@ -151,25 +136,36 @@ function OperatorFunctionsContent({ instancePubkey, account, network }: Operator
...
 
-      // Get the reset SMT root instruction
-      const instruction = await getResetSmtRootInstructionAsync({
+      const instruction = await getRotateBitmapInstructionAsync({
         payer: transactionSigner,
         operator: transactionSigner,
         instance: address(instancePubkey),
-        expectedCurrentTreeIndex: instanceAccount.data.currentTreeIndex,
+        expectedGeneration: bitmapHeader.generation,
       });
```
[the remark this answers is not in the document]
Huzaifa696: `expectedGeneration` only rejects a *replayed* rotation, not a premature one, and this handler bypasses both operator-side barriers. It now counts unset bits and blocks on a confirmation naming the exact count.

Not a hard block: a permanently-failed nonce leaves a hole that never fills, so requiring a full bitmap would wedge rotation forever. The page also cannot see refunds waiting inside the operator, so the confirmation says so.

indexer/src/indexer/datasource/common/parser/escrow.rs
```diff
@@ -19,7 +19,19 @@ const BLOCK_MINT: u8 = 2;
...
+// the legacy layouts of these two instructions: both kept their
+// discriminator, so nothing in the instruction data can select between them.
+// A resync backfills from the genesis slot, so legacy releases are still
+// replayed and both layouts must keep decoding correctly.
+// The legacy counts are what the escrow program required before the bitmap
+// account was added at index 3.
+const RELEASE_FUNDS_ACCOUNTS: usize = 13;
+const RELEASE_FUNDS_ACCOUNTS_LEGACY: usize = 12;
+const ROTATE_BITMAP_ACCOUNTS: usize = 7;
+const ROTATE_BITMAP_ACCOUNTS_LEGACY: usize = 6;
```
sebatustra: nit: should this be RESET_SMT_ACCOUNTS_LEGACY?

indexer/src/indexer/datasource/common/parser/escrow.rs
```diff
@@ -19,7 +19,19 @@ const BLOCK_MINT: u8 = 2;
 // pub(crate) so shared test fixtures can build valid Deposit instruction data.
 pub(crate) const DEPOSIT: u8 = 6;
 const RELEASE_FUNDS: u8 = 7;
-const RESET_SMT_ROOT: u8 = 8;
+const ROTATE_BITMAP: u8 = 8;
+
+// Account counts are the only structural difference between the current and
+// the legacy layouts of these two instructions: both kept their
+// discriminator, so nothing in the instruction data can select between them.
+// A resync backfills from the genesis slot, so legacy releases are still
```
sebatustra: is it necessary to add support for legacy releases if nothing has been deployed on mainnet?

indexer/src/operator/sender/remint.rs
```diff
@@ -382,21 +368,171 @@ pub async fn process_pending_remints(
...
+async fn recorded_release(state: &SenderState, entry: &PendingRemint) -> Option<String> {
+    let nonce = entry.ctx.withdrawal_nonce?;
+
+    match state.storage.get_observed_release(nonce).await {
+        Ok(Some(release)) => Some(format!(
+            "a release for nonce {nonce} is on record (slot {}, signature {}), so it already paid \
+             out and reminting would credit it twice",
+            release.slot, release.signature
+        )),
+        Ok(None) => None,
```
sebatustra: `Ok(None)` here means "the indexer hasn't written a row for this nonce yet", not "no release happened", but it's treated as proof and lets the refund through. Since this is the only gate for every refund whose generation already rotated, read the escrow indexer's committed checkpoint first and route to the same block-and-escalate path as the Err arm when it hasn't covered the slots where a release could have landed

indexer/src/indexer/datasource/common/parser/escrow.rs
Huzaifa696: Deleted it. Legacy releases belong to the abandoned instance, and the backfill drops any transaction that does not touch the configured instance before parsing it, so nothing this decoded was ever stored.

indexer/src/operator/sender/remint.rs
Huzaifa696: Fixed. A missing row can mean no release happened, or that the indexer has not reached that slot yet. It now waits until the checkpoint proves the indexer covered every slot a release could be in.

indexer/src/indexer/datasource/common/parser/escrow.rs
Huzaifa696: removed the legacy constants as per the below comment

sebatustra (approved):

PR #255, opened 2026-08-14
chore: Vendor the DvP swap program from solana-foundation/dvp

gitteri:
The DvP swap program's canonical source moved to [solana-foundation/dvp](https://github.com/solana-foundation/dvp). This repo isn't dropping the feature though: core embeds the program as a precompile, the gateway auth-checks SwapDvp accounts, and integration tests exercise the swap. So rather than keep building a divergent copy from in-tree source, we vendor a snapshot of just the two artifacts we actually consume.

Vendored from `solana-foundation/dvp` @ `39bf82373a6311fd42cc5b9343adc4568da60ef2`.

## What's vendored
- `dvp-swap-program/clients/rust`: the generated `dvp-swap-program-client` crate. Committed here (upstream git-ignores its generated code and publishes no crate), so core/gateway/integration keep depending on it by path.
- `core/precompiles/dvp_swap_program.so`: the compiled program, now a committed real binary (was a symlink into `target/deploy`), rebuilt from upstream HEAD.

## Behavioral changes
- Program id adopts the real devnet id `dvp34bdbcEm4f4FCUjGV4mDAkDshaQR4LkK8fdcsyZq`, replacing the old placeholder `DzG1qJupt6...` (mainnet id will differ and isn't set yet).
- Upstream `SwapDvp` grew 394 to 458 bytes (added `mint_a_authority`/`mint_b_authority`), so the gateway's hand-mirrored `SWAP_DVP_SIZE` moves to 458. The three owner-field offsets (user_a, user_b, settlement_authority) are unchanged, and `CreateDvpBuilder` is byte-identical, so integration `setup.rs` needs no change.

## Removed
In-tree program source, tests, IDL, codegen scripts and build tooling. Dropped the deleted workspace members and pulled the DvP build/test/coverage steps out of the Dockerfile, Makefile and the CI workflows. Also cleaned up the now-orphaned `transfer_hook_fixture`.

## Verification
`cargo check` clean across core, gateway, indexer, integration and the vendored client, including the test trees.

## Re-syncing later
See `dvp-swap-program/README.md`: regenerate the client and rebuild the `.so` in a dvp checkout, copy both in, and update the gateway mirror if the layout changes.
### Coverage Report

| Core | 19,591 | 21,270 | 92.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33882652471) |
| Indexer | 38,763 | 42,485 | 91.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33882652471) |
| Gateway | 1,649 | 1,908 | 86.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33882652471) |
| Auth | 1,312 | 1,506 | 87.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33882652471) |
| Withdraw Program | 129 | 241 | 53.5% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33882652435) |
| Escrow Program | 1,195 | 1,982 | 60.3% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33882652435) |
| E2E Integration | 15,455 | 20,055 | 77.1% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33882652471) |
| **Total** | **78,094** | **89,447** | **87.3%** | |

> Last updated: 2026-09-04 15:10:01 UTC by E2E Integration


.github/workflows/program-integration.yml
```diff
@@ -48,17 +46,13 @@ jobs:
         run: |
           make -C private-channel-escrow-program build
           make -C private-channel-withdraw-program build
-          make -C dvp-swap-program build
 
       - name: Run escrow integration tests
         run: make -C private-channel-escrow-program integration-test-no-build
 
       - name: Run withdraw integration tests
```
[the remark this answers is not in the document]
gitteri: Fair flag, here's the split. The full terminal-op integration suite (Settle/Reclaim/Reject/Cancel/Recover) moved to solana-foundation/dvp along with the program source, where it runs against the program in that repo's CI (128 integration tests). That's its proper home now that this repo vendors the compiled artifact instead of building it.

Against the committed binary here, coverage is:
- `integration/tests/private-channel/rpc/test_swap.rs` exercises create_dvp + fund end to end (past-expiry blocks create, future-expiry funds OK). That is exactly the path that regressed on our zero-rent runtime, and the .so rebuild in this PR (dvp #9) fixes it.
- `gateway/src/auth.rs` checks the SwapDvp account layout (size + owner offsets) against the vendored client.
- This PR now also runs the vendored client's verify + CPI-flag-regression tests in CI (`make dvp-client-test`), which nothing ran before.

Honest gap you're right about: the terminal ops are not exercised against the committed .so in this repo, only in dvp. If we want a smoke test of settle/reclaim here too, porting a minimal litesvm case against the vendored binary is a reasonable follow-up. Flagging rather than doing it in this PR to keep the vendoring change scoped.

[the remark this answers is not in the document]
gitteri: created a followup ticket

gateway/src/auth.rs
```diff
@@ -124,23 +124,19 @@ const DELEGATE_END: usize = 108;
...
-/// Update it with the real program ID once the program is deployed, otherwise
-/// no live DvP account will match and these checks will never fire.
-const DVP_SWAP_PROGRAM: &str = "DzG1qJupt6Khm8s8jB3p93NkhPoiAg2M7vkEhkS15CtC";
+/// DvP swap escrow program ID (devnet deployment).
+const DVP_SWAP_PROGRAM: &str = "dvp34bdbcEm4f4FCUjGV4mDAkDshaQR4LkK8fdcsyZq";
 
 /// Serialized size of a SwapDvp account (SwapDvp::LEN). Smaller DvP-owned
 /// accounts (e.g. the nonce tombstone PDA) are not swaps and are denied.
-const SWAP_DVP_SIZE: usize = 394;
+const SWAP_DVP_SIZE: usize = 458;
```
sebatustra: `SWAP_DVP_SIZE` and `DVP_SWAP_PROGRAM` are now available from the vendored client. adding the client as a gateway dep would reduce the hand written sync risk when size or program id change

gitteri: Done — went a bit further than just size + program id. Added dvp-swap-program-client as a gateway dep and now pull the program id from DVP_SWAP_PROGRAM_ID, and decode with SwapDvp::try_from_bytes (the strict verify helper: exact size check + typed user_a/user_b/settlement_authority fields). That also let me delete the hand-written SWAP_DVP_OWNER_FIELDS byte-offset table — the bit that had actually drifted (its option-tag sibling was off by two pubkeys, see the verify.rs fix in this PR). So no hand-sync left: on the next re-vendor the id, size, and layout all come from the committed client. Same behavior, gateway tests green.

sebatustra (dismissed):

gateway/src/auth.rs
gitteri: Done in f4ed565b6. gateway/src/auth.rs now imports DVP_SWAP_PROGRAM_ID and SwapDvp (and SWAP_DVP_ACCOUNT_LEN from ::verify) straight from the vendored dvp-swap-program-client crate (added as a gateway dep in gateway/Cargo.toml), so DVP_SWAP_PROGRAM and the size are derived from the client rather than hand-mirrored: they can't drift when the id or account size changes upstream. Thanks for the nudge.

.github/workflows/program-integration.yml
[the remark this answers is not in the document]
gitteri: This is intentional and part of the point of the PR. We are no longer maintaining a divergent in-tree DvP program, so its full lifecycle tests (Settle/Reclaim/Reject/Cancel/Recover) move upstream to solana-foundation/dvp, which owns the source now. What this repo actually consumes is still covered: the gateway auth-check of SwapDvp accounts (gateway/src/auth.rs, exercised against the vendored client's SwapDvp/SWAP_DVP_ACCOUNT_LEN) and the integration setup.rs CreateDvpBuilder path. Re-adding a terminal-instruction suite here would mean re-importing the exact source we just vendored away. If a vendored client/binary mismatch is the worry, the re-sync procedure in dvp-swap-program/README.md plus the gateway's now-derived (non-mirrored) constants are the guardrail. Resolving as by-design.

dvp-swap-program/README.md
```diff
@@ -1,146 +1,55 @@
...
-
-```
-program/         on-chain program (no_std, pinocchio-based)
-clients/rust/    Codama-generated Rust client
-clients/typescript/  Codama-generated TypeScript client
-idl/             generated Codama IDL
-tests/integration-tests/   LiteSVM integration tests
-```
+If the `SwapDvp` layout changes, update the hand-mirrored size and field offsets
+in `gateway/src/auth.rs` (`SWAP_DVP_SIZE`, `SWAP_DVP_OWNER_FIELDS`) to match.
```
sebatustra: nit: these two lines no longer apply, commit [f4ed565](https://github.com/solana-foundation/solana-private-channels/commit/f4ed565b6efef1edcc69d4f5cde88eb218e28d3b) removed the two constants

gitteri: Good catch, fixed. Replaced those two lines with a note that the gateway reads SwapDvp size/fields from the vendored client (SwapDvp::try_from_bytes), so a layout change is handled by re-vendoring: no hand-mirrored constants left to update.

sebatustra (approved):

PR #256, opened 2026-08-17
refactor(escrow,indexer): move withdrawal replay protection to an on-chain nonce bitmap

Huzaifa696:
## Summary

Withdrawal replay protection moves from a Sparse Merkle Tree to a fixed 8 KB on-chain nonce bitmap, across both the escrow program and the indexer/operator that drives it.

The nonce model is unchanged, `generation = nonce / 65536`, `bit = nonce % 65536`, but the chain is now the sole replay authority. The off-chain mirror tree, proof generation, rollback and rebuild paths are deleted, and 544 bytes of proof data leave every release transaction.

## Why

The SMT made replay protection a single shared value. One unresolved nonce made the whole root unknowable, which stalled every other withdrawal in the same tree and left the operator retrying proofs against a state it could not confirm. Per-nonce bits are independent: nothing is shared to get stuck on, a delayed release can be resent unchanged, and a divergence names the exact nonces instead of a root mismatch.

## Breaking interface changes

Existing deployments are **not** upgradable in place — see Deployment below.

| Change | Detail |
|---|---|
| `ResetSmtRoot` → `RotateBitmap` | Discriminator stays `8`; arg becomes `expected_generation` |
| New `WithdrawalBitmap` account | Seeds `["withdrawal_bitmap", instance]`, discriminator `3`. Created by `CreateInstance`, so every instance has one by construction |
| `Instance` shrinks 107 → 67 bytes | Drops `withdrawal_transactions_root` and `current_tree_index` |
| `ReleaseFunds` drops `new_withdrawal_root` and `sibling_proofs` | Now `amount`, `user`, `transaction_nonce`. Instance becomes read-only — it only signs the transfer |
| `CreateInstance` gains `withdrawal_bitmap` account + `bitmap_bump` arg | |
| Custom errors 11–14 renumbered | `InvalidWithdrawalBitmap`, `NonceAlreadyUsed`, `NonceOutsideCurrentGeneration`, `UnexpectedGeneration` |
| `ReleaseFundsEvent` 177 → 145 bytes | Loses `new_withdrawal_root`; `RotateBitmapEvent` gains `new_generation` |

## Operator behaviour changes

Same stages, same failure routing, same recovery machinery. What changed is where the truth comes from.

- **`NonceAlreadyUsed` is terminal, where `InvalidSmtProof` was retryable.** The old error meant either a stale local tree or a genuine second release; the operator assumed the first and retried with a new proof. `NonceAlreadyUsed` can only mean the release already landed, so the withdrawal is settled from its recorded signatures instead of retried.
- **A cached generation gates the build, asymmetrically.** The operator trusts its remembered generation only when it says *go*. Anything that would hold a release back is confirmed against the chain first — a release wrongly held would never be sent, never rejected, and so nothing would ever correct the memory that held it.
- **The refund path reads the nonce bit before crediting.** If the read fails, or the bitmap has moved past that nonce, the refund waits rather than proceeding on the signature verdict alone.
- **Boot reconciliation is a directional bit diff, not a root equality check.** Chain ahead of the DB (a release landed, bookkeeping lost) is repaired in place and startup continues. DB ahead of the chain (a release the chain never made) still refuses to start. Either direction names the exact nonces.
- **Rotation blocks on deferred refunds in the generation being cleared.** Wiping the bits would erase the only record of whether a user had already been paid.
- **Queued releases and escalated rows are durable, not in-memory.** A release waiting on the next rotation is written as `Parked` so a crash rebuilds it; escalating an uncertain withdrawal no longer deletes its broadcast signatures, which are exactly what is needed to settle it later.

## Deployment

**Pre-bitmap instances cannot be migrated.** They have no bitmap account and can never gain one, so every release against them fails after this change.

### Coverage Report

| Core | 12,065 | 13,690 | 88.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32033576983) |
| Indexer | 26,841 | 30,176 | 88.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32033576983) |
| Gateway | 1,055 | 1,230 | 85.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32033576983) |
| Auth | 830 | 949 | 87.5% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32033576983) |
| E2E Integration | 12,854 | 16,392 | 78.4% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32033576983) |
| **Total** | **53,645** | **62,437** | **85.9%** | |

> Last updated: 2026-08-17 13:50:10 UTC by E2E Integration


Huzaifa696:
@greptile-apps

sebatustra (approved):

PR #257, opened 2026-08-17
fix(indexer): re-arm the deposit instead of failing it on a transient JIT verdict (issue-285)

Huzaifa696:
## Summary

Closes issue 285. A deposit funds escrow before the operator mints on the PrivateChannel, so its row is the only record that a mint is still owed. When `mint_to` returned `MintNotInitialized`, failures from the JIT `InitializeMint` helper were incorrectly classified as `PermanentFailure` and moved the deposit to terminal `Failed`. These failures can be transient, such as RPC, signer, send, or confirmation errors, and no worker retries `Failed` rows, leaving funded deposits stranded.

These failures are now classified as `Transient` and re-arm the deposit from `processing` to `pending` through the existing cap-gated `try_requeue_prebroadcast` CAS. This does not introduce another value-bearing send: the helper only broadcasts `InitializeMint`, which moves no funds and is idempotent. The failed `mint_to` signature remains in the journal, allowing the processor to re-check it and retry the mint only after the attempt is proven dead. The sender does not write a final status on cap, races, or write errors; recovery remains responsible for escalation and verifies the deposit on-chain first.
### Coverage Report

| Core | 14,024 | 15,713 | 89.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32050980226) |
| Indexer | 35,657 | 39,254 | 90.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32050980226) |
| Gateway | 1,476 | 1,698 | 86.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32050980226) |
| Auth | 908 | 1,059 | 85.7% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32050980226) |
| E2E Integration | 14,344 | 18,258 | 78.6% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32050980226) |
| **Total** | **66,409** | **75,982** | **87.4%** | |

> Last updated: 2026-08-18 15:32:45 UTC by E2E Integration

### DIFF
--- a/indexer/src/operator/sender/transaction.rs
+++ b/indexer/src/operator/sender/transaction.rs
@@ -979,8 +979,12 @@ pub(super) fn handle_confirmation_result<'a>(
                         cleanup_failed_transaction(state, ctx.withdrawal_nonce);
                         state.mint_builders.remove(&txn_id);
                     }
-                    JitOutcome::PermanentFailure(reason) => {
-                        handle_permanent_failure(state, ctx, storage_tx, &reason).await;
+                    // Only deposits reach this block: the guard above proves a
+                    // cached mint builder, and those are cached for the deposit
+                    // mint alone. So there is no nonce-keyed state to unwind
+                    // and no withdrawal remint to compensate here.
+                    JitOutcome::Transient(reason) => {
+                        requeue_deposit_after_jit(state, txn_id, &signature, &reason).await;
                     }
                 }
             }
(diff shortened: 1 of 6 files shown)


indexer/src/operator/sender/transaction.rs
```diff
@@ -1244,6 +1248,81 @@ pub(super) async fn requeue_or_fail_prebroadcast(
...
+        .with_label_values(&[pt, "mint_jit_transient"])
+        .inc();
+
+    // The row leaves this task on every branch below and the next pickup builds
+    // its own builder, so the cached one would only go stale here.
+    state.mint_builders.remove(&txn_id);
+
+    match state
+        .storage
+        .try_requeue_prebroadcast(txn_id, MAX_RECOVERY_REQUEUE_ATTEMPTS)
```
sebatustra, line 1277: should we use `try_demote_processing` instead? I worry about using `try_requeue_prebroadcast` for a txn that was actually broadcasted, does that make sense?

Huzaifa696: The name is historical (it describes its first caller, not a precondition) — the write is just a cap-gated processing → pending CAS that asserts nothing about broadcast state, and the double-mint guard is the journal gate in process_deposit_funds, which only re-mints once the stored signature is proven dead.

PR #258, opened 2026-08-17
fix(indexer): separate a missing checkpoint from a genesis checkpoint and fail closed (issue-278)

Huzaifa696:
## Summary

`get_last_checkpoint` previously treated "no checkpoint row" and "checkpoint at slot zero" as the same value. During the first few seconds of a deployment, the checkpoint row may not exist yet because writes are batched every five seconds. If Yellowstone disconnected during that window, the reconnect logic treated the stream as fresh, resumed from the provider's live position, and could skip slots that arrived while it was disconnected. A `WithdrawFunds` burn in that gap could therefore leave tokens burned with no indexed release.

`get_last_checkpoint` now returns `Option<u64>`, so a missing checkpoint is distinguishable from slot zero. On startup, `run` persists an anchor before creating the Yellowstone datasource: it uses the resolved range floor when backfill ran, or the chain tip when no backfill was needed. Reconnects now use any real anchor, including slot zero, and hold the triggering block until an anchor exists. The checkpoint update remains monotonic and only creates a missing row, leaving an existing checkpoint untouched without an RPC call.
### Coverage Report

| Core | 14,011 | 15,713 | 89.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32051141243) |
| Indexer | 35,746 | 39,356 | 90.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32051141243) |
| Gateway | 1,476 | 1,698 | 86.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32051141243) |
| Auth | 908 | 1,059 | 85.7% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32051141243) |
| E2E Integration | 14,356 | 18,269 | 78.6% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32051141243) |
| **Total** | **66,497** | **76,095** | **87.4%** | |

> Last updated: 2026-08-18 15:23:12 UTC by E2E Integration

### DIFF
--- a/docs/INDEXER.md
+++ b/docs/INDEXER.md
@@ -39,7 +39,12 @@ Recovers missed slots on indexer restart or network issues:
      cannot be proven empty aborts the batch rather than being checkpointed past
    - Process blocks in order
    - Update checkpoint per slot via `CheckpointWriter` (driven by `SlotComplete` events)
-4. Switch to real-time mode (Yellowstone or polling)
+4. For the Yellowstone datasource, persist a startup anchor before the live stream runs, so a
+   durable checkpoint always exists: reconnect gap repair replays from it and withholds live slots
+   rather than advancing the checkpoint without one. The anchor is the resolved backfill range's
+   floor, or the current chain tip when backfill is disabled. RPC polling has no reconnect repair
+   and writes no anchor; it resumes from its configured start slot
+5. Switch to real-time mode (Yellowstone or polling)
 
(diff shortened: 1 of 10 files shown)


sebatustra (commented):
LGTM but CI is red

PR #266, opened 2026-08-19
fix(indexer): start the transaction processor before a one-shot backfill (issue-266)

Huzaifa696:
## Summary

Closes issue 266. Backfill-only mode is used to repair missing finalized deposits or withdrawals, but it previously started the checkpoint writer and backfill producer without starting the transaction processor, which is the component that writes rows and advances the checkpoint. As a result, small repairs could finish successfully while dropping all parsed events, while larger repairs could block on the full instruction channel. A missed deposit could leave funds in escrow without a mint, while a missed withdrawal could leave burned receipt tokens without a release row.

Backfill-only mode now starts and manages the full pipeline: resolve the range, start the writer, spawn the processor, run the backfill, close the sender, join the processor, drain pending events, and verify the final checkpoint before closing the database pool. The processor now receives the escrow instance ID; without it, all escrow instructions are filtered out. The configuration guard is therefore enforced consistently for both backfill and resync. The backfill error is also retained until cleanup so a partial repair preserves the completed prefix, while the final verification fails with `BackfillIncomplete` if the checkpoint did not reach the target.

The same instance-scope validation is now applied to resync before it can drop existing data, and `backfill_only` without `backfill.enabled` is now rejected instead of silently starting live indexing. The live indexing path is unchanged. Tests cover the pipeline, checkpoint verification, configuration guards, idempotent reruns, and write failures that leave the checkpoint safely behind.
### Coverage Report

| Core | 17,749 | 19,421 | 91.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33519179920) |
| Indexer | 37,730 | 41,399 | 91.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33519179920) |
| Gateway | 1,634 | 1,892 | 86.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33519179920) |
| Auth | 1,312 | 1,506 | 87.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33519179920) |
| E2E Integration | 15,216 | 19,594 | 77.7% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33519179920) |
| **Total** | **73,641** | **83,812** | **87.9%** | |

> Last updated: 2026-09-01 15:24:18 UTC by E2E Integration

### DIFF
--- a/indexer/src/indexer/indexer.rs
+++ b/indexer/src/indexer/indexer.rs
@@ -1,5 +1,5 @@
 use crate::config::{ProgramType, ReconciliationConfig};
-use crate::error::{DataSourceError, IndexerError, ReconciliationError};
+use crate::error::{CheckpointError, DataSourceError, IndexerError, ReconciliationError};
 use crate::{
     indexer::{
         checkpoint::{CheckpointMsg, CheckpointWriter},
@@ -21,7 +21,8 @@ use crate::{
     indexer::{
         backfill::{BackfillService, StartupRange},
         checkpoint::{
-            get_last_checkpoint, wait_for_checkpoint_commit, CHECKPOINT_COMMIT_TIMEOUT_SECS,
+            get_last_checkpoint, program_key, wait_for_checkpoint_commit,
+            CHECKPOINT_COMMIT_TIMEOUT_SECS,
         },
     },
     operator::escrow_sweep::CustodySnapshot,
@@ -269,9 +270,28 @@ async fn run_backfill_only(
     storage: Arc<Storage>,
     program_type: ProgramType,
     escrow_instance_id: Option<Pubkey>,
+    configured_start_slot: Option<u64>,
 ) -> Result<(), IndexerError> {
     // Resolve first and fail closed: with no live stream there is no ungated fallback.
     let range = backfill_service.resolve_range().await?;
+
+    // A floor above the durable checkpoint means the slots in between are outside the
+    // fill, yet the gated writer seeds its frontier at that floor and would walk it to
+    // the target, committing a checkpoint over slots nothing ever fetched. Absence of a
+    // checkpoint is the one case a configured start slot may set the floor: it is
+    // initializing a ledger, not skipping one.
+    if let Some(committed) = get_last_checkpoint(&storage, program_type).await? {
+        if range.anchor > committed {
+            return Err(IndexerError::Checkpoint(
+                CheckpointError::StartSlotAheadOfCheckpoint {
+                    setting: "indexer.backfill.start_slot",
+                    program_type: program_key(program_type),
+                    start_slot: configured_start_slot.unwrap_or(range.anchor + 1),
+                    checkpoint: committed,
+                },
+            ));
+        }
+    }
 
     let (instruction_tx, instruction_rx) = mpsc::channel(PIPELINE_CHANNEL_CAPACITY);
     let (checkpoint_tx, checkpoint_rx) = mpsc::channel(PIPELINE_CHANNEL_CAPACITY);
@@ -419,6 +439,7 @@ pub async fn run(
                 storage.clone(),
                 common_config.program_type,
                 common_config.escrow_instance_id,
+                indexer_config.backfill.start_slot,
             )
             .await;
         }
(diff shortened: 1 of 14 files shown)


indexer/src/shutdown_utils.rs
```diff
@@ -541,16 +549,36 @@ pub async fn cleanup_after_backfill(
...
-        Err(_) => warn!("Checkpoint writer drain timed out"),
+        Err(_) => {
+            // Timing out only stops waiting; the task keeps running unless it is cancelled.
+            // Both stages below would then race it: the read could miss a flush that lands
+            // moments later and call a finished backfill incomplete, and the close would
+            // pull the pool out from under a write still in flight. Cancelling first makes
+            // the checkpoint read afterwards a settled fact rather than a snapshot.
+            warn!("Checkpoint writer drain timed out; cancelling it before verifying");
+            checkpoint_handle.abort();
+            let _ = checkpoint_handle.await;
```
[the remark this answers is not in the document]
Huzaifa696: Valid: 10s drain vs sqlx's 30s acquire timeout meant healthy flushes were cancelled by default. Added a dedicated 45s bound for the one-shot path (build-time assert pins it above the pool timeout); the abort stays as a stuck-writer backstop.

indexer/src/shutdown_utils.rs
```diff
@@ -540,17 +566,43 @@ pub async fn cleanup_after_backfill(
...
+            // the checkpoint read afterwards a settled fact rather than a snapshot.
+            //
+            // The bound above is set so that reaching here means the writer is genuinely
+            // stuck rather than slow, which is what makes cancelling it the right call.
+            warn!(
+                "Checkpoint writer still running after {}s; cancelling it before verifying",
+                config.backfill_drain_timeout_secs
+            );
+            checkpoint_handle.abort();
+            let _ = checkpoint_handle.await;
```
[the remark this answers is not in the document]
Huzaifa696: The bound is deliberate: unbounded would let a connection that never answers hold a one-shot job open forever, so the shortfall is reported rather than waited out. What was wrong is that a cancelled writer read as a missing slot; it now returns `BackfillCheckpointUnconfirmed` with the wait duration so the operator knows to re-run rather than hunt a slot, and it stays fail-closed because a cancelled writer leaves the two indistinguishable in the DB.

indexer/src/indexer/indexer.rs
```diff
@@ -311,8 +401,8 @@ pub async fn run(
     }
 
     // 3. Create channels
-    let (instruction_tx, instruction_rx) = mpsc::channel(1000);
-    let (checkpoint_tx, checkpoint_rx) = mpsc::channel(1000);
+    let (instruction_tx, instruction_rx) = mpsc::channel(PIPELINE_CHANNEL_CAPACITY);
```
sebatustra: these two channels are dead in backfill-only mode, since `run_backfill_only` builds its own pair. moving them bellow `if backfill_only {...` would stop the waste

integration/tests/indexer/backfill_only_e2e.rs
```diff
@@ -0,0 +1,408 @@
...
+    assert_eq!(
+        row_a.status, "pending",
+        "a backfilled deposit is queued for the operator, not already serviced"
+    );
+
+    let checkpoint = db::get_checkpoint_slot(&pool, "escrow")
+        .await?
+        .expect("a successful repair must leave a durable checkpoint");
+    assert!(
+        checkpoint >= slot_b,
```
sebatustra, line 319: the module doc claims this test pins "advances the durable checkpoint to the range top". It does not: `slot_b` is the last deposit, and the range top is the tip resolved inside run()

indexer/src/indexer/indexer.rs
```diff
@@ -254,6 +257,93 @@ fn spawn_transaction_processor(
...
+) -> Result<(), IndexerError> {
+    // Resolve first and fail closed: with no live stream there is no ungated fallback.
+    let range = backfill_service.resolve_range().await?;
+
+    let (instruction_tx, instruction_rx) = mpsc::channel(PIPELINE_CHANNEL_CAPACITY);
+    let (checkpoint_tx, checkpoint_rx) = mpsc::channel(PIPELINE_CHANNEL_CAPACITY);
+
+    // Gating to the fill range keeps a failed slot from being leapfrogged by a later one.
+    let mut checkpoint_writer = CheckpointWriter::new(storage.clone());
+    if let Some((from_slot, target)) = range.gap {
```
sebatustra, line 301: from_slot is max(start_slot - 1, last_committed_slot) (backfill.rs:352), and CheckpointState::gated seeds frontier = from_slot (checkpoint.rs:75). So when start_slot sits above the checkpoint, the writer starts its frontier above the checkpoint and the run commits the tip without anything below start_slot ever being fetched.

something like
if from_slot > committed {
     return Err(IndexerError::etc;
 }
in run_backfill_only could prevent this

Huzaifa696: Fixed in b85993f: `run_backfill_only` now refuses when the resolved floor sits above the durable checkpoint, with a unit test and an e2e that pins the checkpoint staying untouched. Guarded on `Option` rather than `from_slot > committed` so a never-indexed ledger can still be initialised from `start_slot` — same rule #284 applies in `resolve_range`, which will supersede this local check when it merges.

sebatustra (approved):

PR #273, opened 2026-08-20
fix(core): require an explicit node mode instead of defaulting to all-in-one (issue-200)

sebatustra:
## Finding 200: Unset node mode starts read and write services

`--mode` had `default_value = "aio"`, so a missing `PRIVATE_CHANNEL_MODE` started both halves of the node instead of failing closed. A container intended as a read node would open the accounts database writable and run the full write pipeline (dedup, sigverify, sequencer, execution, settlement, address index) plus `sendTransaction`.

Two compose variables made this easy to hit: `PRIVATE_CHANNEL_ENABLE_READ` and `PRIVATE_CHANNEL_ENABLE_WRITE`. Nothing in the binary has ever read them, but they look like independent role switches.

## Changes

  - `core/src/bin/node.rs`: `--mode` is now required, `aio` default removed.
  - `core/src/bin/node.rs`: warn at startup if either legacy `ENABLE_*` variable is still set, naming the effective mode.
  - `docker-compose.yml`, `docker-compose.devnet.yml`: removed the six dead `ENABLE_*` lines.
  - `docs/CONFIG.md`: mode is documented as required, startup validation note updated.

## Verification

Ran the built binary. Both drop shapes now exit 2 without starting:

  - Variable unset: `error: the following required arguments were not provided: --mode <MODE>`
  - Variable empty, which is what compose produces from an unset interpolation: `error: a value is required for '--mode <MODE>'
### Coverage Report

| Core | 15,588 | 17,237 | 90.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32510568235) |
| Indexer | 36,841 | 40,456 | 91.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32510568235) |
| Gateway | 1,634 | 1,892 | 86.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32510568235) |
| Auth | 1,312 | 1,506 | 87.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32510568235) |
| E2E Integration | 14,588 | 18,739 | 77.8% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/32510568235) |
| **Total** | **69,963** | **79,830** | **87.6%** | |

> Last updated: 2026-08-21 18:55:01 UTC by E2E Integration


core/src/bin/node.rs
```diff
@@ -21,8 +21,9 @@ use {
     about = "PrivateChannel node that can run in read, write, or all-in-one mode"
 )]
 struct Args {
-    /// Node operation mode
-    #[arg(short, long, default_value = "aio", env = "PRIVATE_CHANNEL_MODE")]
+    /// Node operation mode. Required: a defaulted mode would silently start a
+    /// read deployment as a writer whenever the variable is dropped.
```
Huzaifa696, line 25: Nothing stops someone re-adding default_value = "aio", since no test covers it. Maybe add a parse test asserting the node refuses to start without a mode.

core/src/bin/node.rs
```diff
@@ -256,6 +257,20 @@ async fn main() {
...
     info!("Mode: {:?}", args.mode);
 
+    // Legacy names that read like independent role switches but were never
+    // wired up. --mode is the only role control.
+    for var_name in [
+        "PRIVATE_CHANNEL_ENABLE_READ",
+        "PRIVATE_CHANNEL_ENABLE_WRITE",
+    ] {
+        if std::env::var_os(var_name).is_some() {
+            warn!(
```
Huzaifa696: The stale-ENABLE_* warning runs after arg parsing, so it never fires when the mode is missing, the exact case it's for. 

fix: Check those vars before parsing so the hint reaches the operator.

core/src/bin/node.rs
sebatustra: Added mode_is_required in core/src/bin/node.rs (cargo test --bin node): asserts try_parse_from fails with MissingRequiredArgument naming --mode. Verified it fails if default_value = "aio" is restored.

core/src/bin/node.rs
sebatustra: Right, Args::parse() exited first. Moved the check above it, printing to stderr since logging isn't initialized until after parse. The message drops the effective mode, which isn't known pre-parse.

Huzaifa696 (approved):

PR #276, opened 2026-08-24
fix(indexer): drop the cold-start exemption that left the startup anchor unread (issue-278)

Huzaifa696:
## Summary

The previous fix persisted a startup anchor before the Yellowstone stream opened, but an `ever_streamed` flag exempted the first connection from the gap gate. The anchor was therefore written but never used, leaving the window between startup and the first streamed block unowned. With backfill disabled, the first `SlotComplete` could advance the checkpoint straight past that gap. With backfill enabled, the live-start boundary was only consumed by RPC polling, leaving the slots above the backfill target uncovered. A `WithdrawFunds` burn landing in either window could therefore be missed permanently.

Every Yellowstone connection, including the first, now arms the gap gate on its first live block: the checkpoint is held, the missing window below that block is backfilled, and only then is the block forwarded. The source also carries a startup floor representing the highest slot startup has already accounted for: `live_start_slot - 1` when a range was resolved, otherwise the persisted anchor. This floor widens an existing startup gate rather than allowing a lower stream opening to narrow it, and bounds the first fill to only the remaining uncovered window. `CheckpointState` is unchanged.
### Coverage Report

| Core | 17,750 | 19,421 | 91.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33507369370) |
| Indexer | 37,828 | 41,533 | 91.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33507369370) |
| Gateway | 1,634 | 1,892 | 86.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33507369370) |
| Auth | 1,312 | 1,506 | 87.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33507369370) |
| E2E Integration | 15,148 | 19,576 | 77.4% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33507369370) |
| **Total** | **73,672** | **83,928** | **87.8%** | |

> Last updated: 2026-09-01 13:24:09 UTC by E2E Integration

### DIFF
--- a/indexer/src/indexer/backfill.rs
+++ b/indexer/src/indexer/backfill.rs
@@ -280,20 +280,19 @@ pub async fn ensure_startup_anchor(
 }
 
 /// Highest slot startup owns, one below the live boundary, or the anchor with no backfill.
-/// Bounded both ways: the first gate waits on this slot, so above the tip it waits forever.
+/// Not capped at the chain tip: a node lagging the provider that wrote the checkpoint is normal.
 #[cfg(feature = "datasource-yellowstone")]
 pub fn resolve_startup_floor(
     live_start_slot: Option<u64>,
     anchor: u64,
-    tip: u64,
 ) -> Result<u64, IndexerError> {
     let floor = live_start_slot.map_or(anchor, |slot| slot.saturating_sub(1));
 
-    if floor > tip || floor < anchor {
+    if floor < anchor {
         return Err(DataSourceError::InvalidConfig {
             reason: format!(
-                "startup floor {floor} is outside the recoverable range [{anchor}, {tip}]; \
-                 check backfill.from_slot against the chain the RPC endpoint serves"
+                "startup floor {floor} is below the durable anchor {anchor}; the live boundary \
+                 and the anchor were resolved from different ranges"
             ),
         }
         .into());
@@ -390,6 +389,18 @@ impl BackfillService {
         };
 
         let current_slot = latest_slot_with_retry(&self.rpc_poller).await?;
+
+        // A boundary this far past the chain is a misconfiguration, not lag. Warn rather than
+        // refuse, because a node trailing the provider that wrote the checkpoint is routine.
+        if from_slot.saturating_sub(current_slot) > self.config.max_gap_slots {
+            warn!(
+                "Startup boundary {} is {} slots past the tip {} this endpoint reports; check \
+                 backfill.start_slot against the chain it serves",
+                from_slot,
+                from_slot - current_slot,
+                current_slot
+            );
+        }
 
         // One past the highest slot backfill covers (gap) or the durable checkpoint
         // (no gap); max guards against an RPC node lagging behind the checkpoint.
@@ -617,40 +628,14 @@ mod tests {
     #[cfg(feature = "datasource-yellowstone")]
     #[test]
     fn startup_floor_is_the_slot_below_the_live_boundary() {
-        assert_eq!(
-            resolve_startup_floor(Some(1_001), 500, 2_000).unwrap(),
-            1_000
-        );
+        assert_eq!(resolve_startup_floor(Some(1_001), 500).unwrap(), 1_000);
     }
 
     /// With no backfill there is no range to inherit and the anchor is the boundary itself.
     #[cfg(feature = "datasource-yellowstone")]
     #[test]
     fn startup_floor_falls_back_to_the_anchor() {
-        assert_eq!(resolve_startup_floor(None, 500, 2_000).unwrap(), 500);
-    }
-
-    /// A from_slot set past the chain would gate on slots that can never be produced, which
-    /// freezes the checkpoint for good. Refuse to start rather than stall silently.
-    #[cfg(feature = "datasource-yellowstone")]
-    #[test]
-    fn startup_floor_rejects_a_boundary_above_the_chain_tip() {
-        let err = resolve_startup_floor(Some(9_001), 500, 2_000).unwrap_err();
-        assert!(
-            err.to_string().contains("9000"),
-            "the error must name the rejected floor: {err}"
-        );
-    }
-
-    /// The tip is read after the boundary is resolved, so equality is the normal case for a
-    /// deployment starting at the tip and must not be mistaken for a misconfiguration.
-    #[cfg(feature = "datasource-yellowstone")]
-    #[test]
-    fn startup_floor_accepts_a_boundary_at_the_chain_tip() {
-        assert_eq!(
-            resolve_startup_floor(Some(2_001), 500, 2_000).unwrap(),
-            2_000
-        );
+        assert_eq!(resolve_startup_floor(None, 500).unwrap(), 500);
     }
 
     /// The anchor is the lowest slot startup owns, so a floor under it means the two were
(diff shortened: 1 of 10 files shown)


indexer/src/indexer/datasource/yellowstone/source.rs
```diff
@@ -561,13 +563,12 @@ async fn connect_and_stream(
...
-    // live block arms the gate + backfill so the residual window cannot be leapfrogged.
+    // Set on every connection, the first one included, so that connection's first live block
+    // gates the window below it. A process start and a stream resume are the same event here.
     #[cfg(feature = "datasource-rpc")] gap_ctx: Option<&ReconnectGapCtx>,
     #[cfg(feature = "datasource-rpc")] backfill_handle: &mut Option<tokio::task::JoinHandle<()>>,
-    // Set true once this connection subscribes, so the caller can tell a real streaming
-    // session apart from a connect failure that never reached the reconnect-gating state.
-    #[cfg(feature = "datasource-rpc")] subscribed: &mut bool,
+    // The slot startup already owns, consumed by whichever connection arms the gate first.
+    #[cfg(feature = "datasource-rpc")] startup_floor: &mut Option<u64>,
```
gitteri, line 579: worth having a sanity check for the startup floor to make sure a bad config isn't used?

gitteri (dismissed):

indexer/src/indexer/datasource/yellowstone/source.rs
```diff
@@ -690,15 +685,15 @@ async fn connect_and_stream(
                         block.transactions.len()
                     );
 
-                    // First live block after a reconnect: observe the resume slot and
-                    // arm the gate + concurrent backfill BEFORE forwarding this block,
-                    // so its SlotComplete cannot leapfrog the still-unfilled window.
+                    // Arm the gate and backfill BEFORE forwarding this block, so its
```
gitteri, line 735: startup_floor.take() is unconditional on first arm (source.rs): if that first arm fails after taking it, the next reconnect runs with floor = None. Narrow double-fault could weaken the fail-closed guarantee for a residual window; data still eventually covered by backfill.

indexer/src/indexer/datasource/yellowstone/source.rs
Huzaifa696: Added resolve_startup_floor, which now rejects a floor outside [anchor, tip] at startup, since a floor past the tip would gate on slots that never arrive and freeze the checkpoint for good.

indexer/src/indexer/datasource/yellowstone/source.rs
Huzaifa696: fixed: arm_reconnect_gap now takes &mut Option<u64> and clears the floor only after the gate is sent and the fill is spawned, so a failed arm leaves it for the next connection.

indexer/src/indexer/backfill.rs
```diff
@@ -279,6 +279,29 @@ pub async fn ensure_startup_anchor(
...
+/// Bounded both ways: the first gate waits on this slot, so above the tip it waits forever.
+#[cfg(feature = "datasource-yellowstone")]
+pub fn resolve_startup_floor(
+    live_start_slot: Option<u64>,
+    anchor: u64,
+    tip: u64,
+) -> Result<u64, IndexerError> {
+    let floor = live_start_slot.map_or(anchor, |slot| slot.saturating_sub(1));
+
+    if floor > tip || floor < anchor {
```
sebatustra: the `floor > tip` half of this check would stop the process if the RPC node is behind the checkpoint, is this intentional? simple example that I believe is not unrealistic:

Restart with `backfill.enabled = false` and a durable checkpoint at slot 1000. The RPC node reports finalized slot 998, because it lags a little behind the Geyser provider that wrote that checkpoint.

indexer.rs:662  anchor = ensure_startup_anchor(...)  -> 1000   (existing checkpoint, returned untouched)
indexer.rs:673  tip    = latest_slot_with_retry(...) ->  998   (getSlot on backfill.rpc_url)
backfill.rs:290 floor  = anchor                      -> 1000   (live_start_slot is None)
backfill.rs:292 1000 > 998 -> Err(InvalidConfig)

run returns Err, bin/indexer.rs:350 propagates it with ?, the process exits.

indexer/src/indexer/datasource/yellowstone/source.rs
```diff
@@ -195,36 +189,38 @@ where
...
-                error!(
-                    "Reconnect gap-fill: no durable checkpoint anchor for {:?}; holding the \
-                     checkpoint rather than skipping the gap",
-                    program_type
-                );
-                record_missing_anchor(program_type);
-                if backoff_cancelled(cancel, backoff).await {
-                    return GapFillOutcome::Cancelled;
+        // A cold start is handed the slot startup already owns, so there is nothing to derive.
+        let checkpoint = match floor {
```
sebatustra: if `floor` is `Some` (a cold start for example) the checkpoint is a constant, and so is target (since its a parameter). In that case, if `validate_gap` comes back with `GapTooLarge` once, it will return the same answer as long as that loop lives. is that intended?

indexer/src/indexer/backfill.rs
Huzaifa696: Not intentional, good catch: the tip bound is gone and the floor is only checked against the anchor now, since a node trailing the provider that wrote the checkpoint is routine and `resolve_range` already tolerates it.

indexer/src/indexer/datasource/yellowstone/source.rs
Huzaifa696: Not intended, fixed: the checkpoint is re-read every cycle now and the floor only acts as a lower bound via `max(floor, checkpoint)`, so an operator resync clears an oversized gap without a restart.

indexer/src/indexer/datasource/yellowstone/source.rs
```diff
@@ -712,15 +693,15 @@ async fn connect_and_stream(
...
-                    // arm the gate + concurrent backfill BEFORE forwarding this block,
-                    // so its SlotComplete cannot leapfrog the still-unfilled window.
+                    // Arm the gate and backfill BEFORE forwarding this block, so its
+                    // SlotComplete cannot leapfrog the window still unfilled underneath it.
                     #[cfg(feature = "datasource-rpc")]
                     if let Some(ctx) = gap_ctx {
                         if !armed {
                             armed = true;
                             let safe_to_forward = arm_reconnect_gap(
                                 block.slot,
```
sebatustra, line 742: block.slot here is the first block touching the program (the subscription filters on account_include, source.rs:609-617), so on a cold start this measures idle time, not outage time. Boot with the default max_gap_slots: 1000 and no program traffic for 15 min gives gap = 2250 at source.rs:230, which takes the GapTooLarge branch at source.rs:241 and deadlocks

Huzaifa696: Confirmed and fixed: the subscription now also asks for `slots`, so the gate arms on the slot the stream resumed at rather than the first program-filtered block, which keeps the measured gap the outage instead of the idle stretch.

sebatustra (approved):

PR #277, opened 2026-08-25
fix(indexer): decouple bitmap rotation from the boundary withdrawal (issue-2)

Huzaifa696:
## Summary

`RotateBitmap` was previously dispatched only while processing the withdrawal whose nonce landed on a generation boundary. That made a single row responsible for triggering rotation for every later withdrawal. If that row was quarantined by a build error, swept into `manual_review`, or never written because a losing insert consumed its sequence value, rotation never happened. The chain remained on the old generation, and every later withdrawal parked and retried indefinitely with no visibility into the blockage.

The sender now checks every five seconds whether the lowest withdrawal nonce that still owes a release belongs to a later generation than the on-chain bitmap. If so, it arms a rotation. A single aggregate provides the required state: the lowest outstanding nonce proves the current generation is finished and work exists beyond it, while the upper bound distinguishes a blocked rotation from an idle system. No durable state is needed because the condition is derived from the database and chain on every tick, so a crash cannot permanently lose a rotation trigger. The existing in-flight and remint barriers still determine when the rotation is actually sent.

Any non-terminal withdrawal, including `manual_review`, blocks rotation and therefore later generations until the issue is resolved. This is intentional and is now surfaced through a counter and a log identifying the blocking nonce. The PR adds 21 tests: 14 driver tests using mocks, four Postgres SQL tests, and three end-to-end tests, alongside three updated processor guards. 
### Coverage Report

| Core | 12,065 | 13,690 | 88.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33492044319) |
| Indexer | 27,372 | 30,743 | 89.0% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33492044319) |
| Gateway | 1,055 | 1,230 | 85.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33492044319) |
| Auth | 830 | 949 | 87.5% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33492044319) |
| E2E Integration | 13,054 | 16,506 | 79.1% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33492044319) |
| **Total** | **54,376** | **63,118** | **86.1%** | |

> Last updated: 2026-09-01 10:08:28 UTC by E2E Integration


indexer/src/operator/sender/proof.rs
```diff
@@ -139,6 +143,163 @@ fn generation_mismatch(nonce: u64, nonce_generation: u64, chain_generation: u64)
...
+        return;
+    };
+
+    // Everything waiting is still inside that window, so nothing needs a
+    // rotation and the chain need not be read to say so.
+    if state
+        .cached_generation
+        .is_some_and(|cached| highest / NONCES_PER_GENERATION <= cached)
+    {
+        return;
```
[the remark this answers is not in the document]
Huzaifa696: Correct that the cache is confirmed-sourced, but a stale lead can't persist: the waiting releases keep retrying, each on-chain refusal calls refresh_generation which corrects the cache, and a restart clears it entirely, so the rotation arms on the next tick. Reworded the comment that claimed the cache can never lead.

indexer/src/operator/sender/transaction.rs
```diff
@@ -398,6 +398,9 @@ fn rearm_failed_rotation(state: &mut SenderState, error_msg: &str) {
              unopened generation stays unreleasable until a rotation lands."
         );
         state.rotation_in_flight = None;
+        // The driver starts a fresh rotation once nothing is in flight, and a
+        // count left at the limit would abandon that one on its first failure.
+        state.rotation_rearm_attempts = 0;
```
gitteri, line 403: **Permanent-abandonment cap removed.** `give_up` now resets `rotation_rearm_attempts` to 0. Combined with the new 5s re-origination driver, a rotation the chain deterministically rejects (bad operator authority, escrow-program bug) gets re-derived and re-broadcast every ~5s indefinitely: paying fees and emitting `rotation_lost` on a loop with no hard stop after `MAX_ROTATION_REARMS`. Largely intended by the re-derive design, but the pre-PR permanent cap is gone. Worth a bounded backoff or a dead-letter state so a deterministically-failing rotation eventually stops and pages instead of spinning forever.

indexer/src/operator/sender/proof.rs
```diff
@@ -139,6 +143,163 @@ fn generation_mismatch(nonce: u64, nonce_generation: u64, chain_generation: u64)
...
+    }
+
+    let chain_generation = match state.refresh_generation().await {
```
gitteri, line 205: **Silent rotation stall on persistent read failures.** If `refresh_generation` here (or `unreleased_withdrawal_nonce_bounds` / `read_unreleased_bounds`) keeps erroring during a DB/RPC outage, every 5s tick returns early with only a `warn` log and no metric. Next-generation withdrawals wedge for the whole outage, but no `OPERATOR_TRANSACTION_ERRORS` series moves, so dashboards/alerts show nothing. Note the inversion: the less-severe correct-withholding case *does* get a counter (`rotation_blocked_by_lower_nonce`), while this hard-failure path is invisible. Suggest emitting an error counter on these early-return read failures.

indexer/src/operator/sender/proof.rs
```diff
@@ -139,6 +143,163 @@ fn generation_mismatch(nonce: u64, nonce_generation: u64, chain_generation: u64)
...
+    // Nothing is waiting past the window the chain is on, so nothing is blocked.
+    if highest_generation <= chain_generation {
+        state.rotation_blocked_passes = 0;
+        return;
+    }
+
+    // Withholding is correct here, but only a human clears it, so report it.
+    // Once per threshold, not once per block: a counter that stops ticking lets
+    // the alert resolve while the stall is still running.
+    state.rotation_blocked_passes = state.rotation_blocked_passes.saturating_add(1);
```
gitteri, line 250: **False-positive alert on slow-but-progressing releases.** The `rotation_blocked_by_lower_nonce` counter increments for any live lower nonce held >5min, including a release still churning in `Processing`/in-flight but making progress. Under sustained RPC congestion the last release of a closing generation can page the operator for an "incident" that needs no manual resolution and self-clears, contradicting the runbook's promise that a rising count is "a row someone has to resolve." Consider excluding actively-progressing statuses (in-flight / recently-submitted) from the blocked condition so the alert stays a true has-to-be-resolved signal.

indexer/src/operator/sender/proof.rs
```diff
@@ -139,6 +143,163 @@ fn generation_mismatch(nonce: u64, nonce_generation: u64, chain_generation: u64)
     .into()
 }
 
+/// Five minutes at the origination interval. A boundary crossing looks blocked
+/// while the closing generation's last row settles, and that clears itself.
+const ROTATION_BLOCKED_ALERT_PASSES: u32 = 60;
```
gitteri: **"5 minutes" contract split across two files with no compile-time link.** `ROTATION_BLOCKED_ALERT_PASSES = 60` here and `ROTATION_ORIGINATION_INTERVAL = 5s` in `mod.rs` together encode the documented 5-minute alert threshold, but nothing ties them together. Changing the interval to e.g. 10s silently makes this alert fire at 10 minutes while `docs/runbooks/_glossary.md` and `withdrawal_manual_review.md` still say five, with no test to catch the drift. Suggest deriving the pass count from the interval (`5min / ROTATION_ORIGINATION_INTERVAL`) or a `const_assert`/comment linking the two so the documented contract can't silently break.

indexer/src/operator/sender/proof.rs
```diff
@@ -396,6 +559,433 @@ mod tests {
...
+            push_withdrawal_with_nonce(&mock, *id, *nonce, *status);
+        }
+        let mut state = sender_state_with_storage(url, mock);
+        state.instance_pda = Some(Pubkey::new_unique());
+        state
+    }
+
+    fn blocked_count() -> f64 {
+        crate::metrics::OPERATOR_TRANSACTION_ERRORS
+            .with_label_values(&["escrow", "rotation_blocked_by_lower_nonce"])
```
gitteri, line 625: **Test asserts a metric label prod never emits.** `blocked_count()` reads the counter under the `escrow` program label, and `originator_state` builds on `sender_state_with_storage` which hardcodes `ProgramType::Escrow` while forcing `instance_pda = Some`. But the driver only runs for the `Withdraw` role (`instance_pda` is `Some` only for Withdraw), so production emits this counter under the `withdraw` label. That escrow+`Some(instance_pda)` combination is one `run_sender` never produces, so a regression in the real emitted label (or its registration) would pass these tests unnoticed. Suggest driving the test state from `ProgramType::Withdraw` so the asserted label matches prod.

gitteri (commented):
Automated review pass (xhigh). No high-severity correctness bug survived scrutiny; the change is well-engineered and heavily tested. Five lower-severity findings below, roughly in priority order. #4 (unbounded rotation retry) and #2 (silent-stall blind spot) are the two I'd act on before merge; the rest are alert-quality / maintainability / test-fidelity cleanups.

indexer/src/operator/sender/proof.rs
Huzaifa696: Fixed: every read failure in the driver now increments rotation_origination_read_failed (both the bounds query and the generation read, plus the two conversion guards), so a DB or RPC outage moves a series instead of only writing a log line.

indexer/src/operator/sender/proof.rs
Huzaifa696: Fixed: ROTATION_BLOCKED_ALERT_PASSES is now derived from ROTATION_ORIGINATION_INTERVAL rather than hardcoded, with a test asserting their product is still exactly five minutes, so retuning the interval moves the threshold with it instead of silently retiming the alert.

indexer/src/operator/sender/proof.rs
```diff
@@ -139,6 +144,187 @@ fn generation_mismatch(nonce: u64, nonce_generation: u64, chain_generation: u64)
...
+
+    // Start the search at the cached generation so the usual pass, which finds
+    // nothing to do, skips the older nonces instead of scanning all of them. The
+    // cache sits at or behind the chain, and a refused release corrects it if not.
+    let cached_floor = state
+        .cached_generation
+        .map(|cached| cached.saturating_mul(NONCES_PER_GENERATION))
+        .unwrap_or(0);
+    let Some((_, highest)) = read_unreleased_bounds(state, cached_floor).await else {
+        return;
```
sebatustra: should we also reset `rotation_blocked_passes` in these early returns ("nothing is blocked" paths)? I think an earlier crossing (from a previous generation) can leave the tick at a non zero value, thus incrementing a metric when it shouldn't

indexer/src/operator/sender/proof.rs
```diff
@@ -579,6 +604,24 @@ mod tests {
...
+    }
+
+    /// The runbooks promise the withheld-rotation alert fires after five minutes,
+    /// but the threshold is a pass count and the passes come from a timer in
+    /// another file. Changing either without this would move the alert silently.
+    #[test]
+    fn the_blocked_alert_reports_after_five_minutes() {
+        assert_eq!(
+            super::super::ROTATION_ORIGINATION_INTERVAL * ROTATION_BLOCKED_ALERT_PASSES,
+            std::time::Duration::from_secs(300),
```
sebatustra: should we use `ROTATION_BLOCKED_ALERT_AFTER` instead of hardcoding `Duration::from_secs(300)`?

indexer/src/operator/sender/proof.rs
Huzaifa696: read_unreleased_bounds now distinguishes "nothing owed" from "couldn't read", so every pass that establishes no block clears the streak while read failures deliberately leave it standing.

indexer/src/operator/sender/proof.rs
Huzaifa696: Done, split into two assertions so a failure says which drifted: ROTATION_BLOCKED_ALERT_AFTER == 300s pins the figure the runbooks quote, and INTERVAL * PASSES == ROTATION_BLOCKED_ALERT_AFTER catches an interval that doesn't divide it.

sebatustra (approved):
LGTM

PR #282, opened 2026-08-28
fix(core): reject v0 transactions declaring address table lookups at admission (issue-182)

Huzaifa696:
## Summary

`send_transaction_impl` and `simulate_transaction` used a stub address loader that ignored V0 `address_table_lookups` and returned no loaded addresses. As a result, a V0 transaction declaring address-table lookups was accepted as if it had no lookups. Sanitization did not catch this because it validates instruction indices against the declared lookup counts, while the stub loader returned only the static keys. Program IDs and signers remained static, so the allowlist and admin-key checks were not bypassable; the issue was that lookup accounts were missing from scheduler lock extraction, executor preloading, and the address index.

The fix rejects transactions that declare address-table lookups at admission. A shared `has_address_table_lookups` predicate sits beside the existing allowlist check, and both RPC entry points return `-32602` before building the transaction. Using `SimpleAddressLoader::Disabled` was rejected because `SanitizedTransaction::try_new` calls `load_addresses` for every V0 transaction, which would also reject valid lookup-free V0 transactions. Address resolution is not possible because no address lookup table program is allowlisted, so any declared lookup cannot be resolved. Declared-but-unused lookups are rejected as well, matching real validator behaviour.
### Coverage Report

| Core | 16,554 | 18,203 | 90.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33169895017) |
| Indexer | 37,615 | 41,281 | 91.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33169895017) |
| Gateway | 1,634 | 1,892 | 86.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33169895017) |
| Auth | 1,312 | 1,506 | 87.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33169895017) |
| E2E Integration | 15,019 | 19,374 | 77.5% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33169895017) |
| **Total** | **72,134** | **82,256** | **87.7%** | |

> Last updated: 2026-08-28 13:01:44 UTC by E2E Integration


sebatustra (approved):
LGTM

PR #286, opened 2026-08-31
fix(core): distinguish an unreadable account store from an absent account (issue-150)

Huzaifa696:
## Summary

The storage accessors `get_account_shared_data` and `get_accounts` returned `None` for both a genuinely missing account and a failed read. `getAccountInfo` therefore returned a null account, while `getTokenAccountBalance` returned a client error telling the caller not to retry. More importantly, the executor preloads accounts through the same path: a failed read made a real funded account look nonexistent, and the settler could then write that empty state over the real row. The lamport-conservation check reads the same cache, so it could not detect the problem.

Both accessors now return `Result` with a typed `AccountLoadError`. `Backend` represents a failed query and is retried under a total time budget rather than an attempt count, since the pool's 30-second acquire timeout could otherwise make four attempts block the executor for two minutes. `Corrupt` represents an account that cannot be deserialized and is treated as fatal because the stored bytes cannot change; database errors are also not retried. If an account preload fails, the executor aborts the batch before execution or writes, so the node stops instead of persisting state based on incomplete reads, while dropped transactions remain resubmittable. Redis remains a cache: stale or missing entries fall back to Postgres.

The block and transaction side of this finding was fixed previously; this closes the account side. Adds 19 unit tests, two over-the-wire integration tests, and an operator runbook because a corrupt row now causes the node to stop.
### Coverage Report

| Core | 18,548 | 20,264 | 91.5% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33616666584) |
| Indexer | 38,617 | 42,331 | 91.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33616666584) |
| Gateway | 1,648 | 1,907 | 86.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33616666584) |
| Auth | 1,312 | 1,506 | 87.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33616666584) |
| E2E Integration | 15,312 | 19,912 | 76.9% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33616666584) |
| **Total** | **75,437** | **85,920** | **87.8%** | |

> Last updated: 2026-09-02 10:38:15 UTC by E2E Integration


core/src/accounts/get_accounts.rs
```diff
@@ -7,13 +7,194 @@ use {
...
+pub(crate) fn reset_test_retry() {
+    set_test_retry(LOAD_MAX_ATTEMPTS, LOAD_BACKOFF_BASE_MS);
+    set_test_budget_ms(LOAD_TOTAL_BUDGET_MS);
+}
+
+/// Whether a failed query could plausibly succeed if tried again. A database
+/// that answered with an error answered: a missing table or a rejected
+/// permission will say the same thing every time, so only transport and pool
+/// failures are worth a second attempt.
+fn is_transient(error: &sqlx::Error) -> bool {
```
sebatustra, line 107: can there be transient `sqlx::Error::Database` errors? they are excluded completely

core/src/accounts/get_accounts.rs
```diff
@@ -7,13 +7,194 @@ use {
...
+            }
+        }
+    }
+}
+
+impl std::error::Error for AccountLoadError {}
+
+/// Bounded retry for a whole-query failure. A blip must not restart the node,
+/// but a sustained outage must still reach a verdict in well under a second.
+const LOAD_MAX_ATTEMPTS: u32 = 4;
```
sebatustra, line 42: how is this LOAD_MAX_ATTEMPTS determined? could real database interrptions last longer, thus returning Backend error too soon? also that would kill the node right?

core/src/accounts/bob.rs
sebatustra, line 229: what would happen with these drained settled_accounts if `let accounts = self.accounts_db.get_accounts(&miss_keys).await?;` fails? could those accounts be lost?

core/src/accounts/get_accounts.rs
Huzaifa696: Good catch, fixed: `is_transient` now retries `Database` errors whose SQLSTATE is class 08 (connection), 53 (resources) or 57 (operator intervention) plus 40001/40P01, while the 42xx family still fails on the first attempt.

core/src/accounts/get_accounts.rs
Huzaifa696: The real bound is `LOAD_TOTAL_BUDGET_MS` (5s), sized to outlast a blip like a failover or a recycled connection rather than a real outage, and yes it stops the node on purpose: a restart is recoverable, a block settled on accounts we could not read is not.

core/src/accounts/bob.rs
Huzaifa696: No, those are acknowledgements of writes already durable in Postgres, so draining them only updates in-memory clean/dirty bookkeeping and a failed fetch returns before anything else in BOB is touched.

sebatustra (approved):

PR #288, opened 2026-08-31
fix(core): retry settlement and drain the pipeline in order instead of discarding executed transactions (issue-160)

Huzaifa696:
## Summary

The settler discarded executed transactions on three paths. A failed blocktime tick broke the worker loop with no retry, so one transient Postgres error could stop block production. The final flush cleared its buffer after a failed commit with only a `warn!` and no signatures. On shutdown, every stage also raced the same cancellation token, so a clean restart could drop already-admitted work at four points, from the sequencer's queued input through to the address-index writer.

Settlement now retries within a 15s budget, with each commit bounded at 5s. The timeout covers only the Postgres commit, never the publishes after it: those are not idempotent, and republishing a blockhash would add a second copy to the dedup window, potentially evicting the first and dropping the replay cache for a still-live blockhash. The 15s budget is tied to `transaction_expiration_ms`: because a stall stops block production while the blockhash window is measured in blocks, the stall itself consumes the expiration window, so the retry budget caps the overshoot at twice that window. The timestamp is sampled once per call, so a retry after an unacknowledged commit rewrites the same rows rather than creating a second blockhash that a replica may already serve. The heartbeat is updated on every attempt, preventing a stalled settler from continuing to report as healthy.

Shutdown now closes admission at the RPC layer, after which each stage exits when its input closes, draining in order under a single deadline.

### Coverage Report

| Core | 19,587 | 21,270 | 92.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33645253246) |
| Indexer | 38,617 | 42,331 | 91.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33645253246) |
| Gateway | 1,648 | 1,907 | 86.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33645253246) |
| Auth | 1,312 | 1,506 | 87.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33645253246) |
| E2E Integration | 15,350 | 20,002 | 76.7% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33645253246) |
| **Total** | **76,514** | **87,016** | **87.9%** | |

> Last updated: 2026-09-02 15:48:28 UTC by E2E Integration


core/src/stages/settle.rs
```diff
@@ -849,13 +1020,20 @@ async fn settle_transactions(
...
-        )
-        .await?;
+        ),
+    )
+    .await
+    .map_err(|_| {
+        SettleError::Other(format!(
+            "commit exceeded {commit_timeout:?} and was abandoned"
+        ))
+    })??;
```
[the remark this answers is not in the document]
Huzaifa696: Fixed in 07f79ab: the block insert moved above the counter and now returns `(xmax = 0)`, so a replayed commit increments `transaction_count` only when the slot is genuinely new.

core/src/nodes/node.rs
```diff
@@ -474,15 +478,22 @@ impl NodeHandles {
...
+        let deadline = tokio::time::Instant::now() + DRAIN_DEADLINE;
         for worker in self.workers {
-            match tokio::time::timeout(Duration::from_secs(5), worker.handle).await {
+            let remaining = deadline.saturating_duration_since(tokio::time::Instant::now());
+            match tokio::time::timeout(remaining, worker.handle).await {
                 Ok(Ok(_)) => info!("{} stopped gracefully", worker.name),
                 Ok(Err(e)) => error!("{} error: {:?}", worker.name, e),
-                Err(_) => warn!("{} shutdown timeout", worker.name),
+                Err(_) => warn!("{} did not drain within the shutdown deadline", worker.name),
             }
```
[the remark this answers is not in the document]
Huzaifa696: Fixed in 07f79ab: `shutdown()` now awaits `&mut worker.handle` and calls `abort()` on timeout, and reports an incomplete drain instead of logging "complete" unconditionally.

integration/tests/private-channel/test_ordered_shutdown_e2e.rs
```diff
@@ -0,0 +1,353 @@
...
+        carried > 0.0,
+        "shutdown landed on an empty pipeline, so the drain was never exercised"
+    );
+    let accepted_at_shutdown = loader.accepted_so_far();
+
+    handles.shutdown().await;
+
+    // Admission must be closed once shutdown has started.
+    let after_shutdown = client
+        .send_transaction(&memo_tx(solana_sdk::hash::Hash::default(), 9_999))
```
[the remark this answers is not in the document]
Huzaifa696: Fixed in 07f79ab: switched to `Hash::new_unique()`, which also makes clear the refusal under test is admission rather than an expired blockhash.

core/src/nodes/node.rs
```diff
@@ -474,25 +483,97 @@ impl NodeHandles {
...
+            let remaining = deadline.saturating_duration_since(tokio::time::Instant::now());
+            match tokio::time::timeout(remaining, &mut worker.handle).await {
                 Ok(Ok(_)) => info!("{} stopped gracefully", worker.name),
                 Ok(Err(e)) => error!("{} error: {:?}", worker.name, e),
-                Err(_) => warn!("{} shutdown timeout", worker.name),
+                Err(_) => {
+                    // Dropping a JoinHandle detaches the task rather than
+                    // stopping it, so an in-process restart would leave this
+                    // stage holding its pool and channels beside the new one.
+                    worker.handle.abort();
```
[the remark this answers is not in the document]
Huzaifa696: Fixed in 48ef1e5: `shutdown()` now awaits the handle after `abort()`, bounded by a 2s grace so a task with no await point cannot hold the drain open, and logs an error if one ignores the abort.

core/src/nodes/node.rs
```diff
@@ -474,25 +488,132 @@ impl NodeHandles {
...
-            match tokio::time::timeout(Duration::from_secs(5), worker.handle).await {
+        // One deadline for the whole drain, not one per worker: the workers are
+        // awaited in pipeline order, so a per-worker budget would multiply by
+        // the number of stages and overrun the container's stop grace period,
+        // which kills the process mid-drain and loses what the order preserved.
+        let deadline = tokio::time::Instant::now() + DRAIN_DEADLINE;
+        let mut overran = false;
+        for mut worker in self.workers {
+            let remaining = deadline.saturating_duration_since(tokio::time::Instant::now());
+            match tokio::time::timeout(remaining, &mut worker.handle).await {
```
sebatustra, line 517: wait_for_any_worker_quit waits until any worker task finishes, then returns its name. It never removes the finished worker from the list:

```
  let futures: Vec<_> = self
      .workers
      .iter_mut()                                    // 473: borrows, does not move
      .enumerate()
      .map(|(idx, worker)| {
          let future = (&mut worker.handle).map(move |_| idx);   // 476: polls the handle
          Box::pin(future)
      })
      .collect();

  let (completed_idx, _result, _remaining) = futures::future::select_all(futures).await;  // 481
  let worker_name = self.workers[completed_idx].name().to_string();                        // 482: still in the Vec
```

Line 476 polls the JoinHandle and line 481 is where one of those polls returns Ready. That poll consumes the task's result. But self.workers still holds all 10 handles, including the consumed one.

shutdown() then walks the whole list and polls every handle a second time:
```
  for mut worker in self.workers {
      let remaining = deadline.saturating_duration_since(tokio::time::Instant::now());
      match tokio::time::timeout(remaining, &mut worker.handle).await {   // 511: second poll
```

Why it is wrong:
A tokio JoinHandle can only be polled to completion once. The second poll panics:

```
  pub(super) fn try_read_output(self, dst: &mut Poll<super::Result<T::Output>>, waker: &Waker) {
      if can_read_output(self.header(), self.trailer(), waker) {   // true once complete
          *dst = Poll::Ready(self.core().take_output());
      }
  }

  match mem::replace(unsafe { &mut *ptr }, Stage::Consumed) {
      Stage::Finished(output) => output,
      _ => panic!("JoinHandle polled after completion"),
  }
```

So node.rs:511 panics as soon as the loop reaches the worker that already finished. The panic escapes shutdown(), escapes main (core/src/bin/node.rs:247-248, #[tokio::main]), and the process aborts.

core/src/stages/settle.rs
```diff
@@ -687,7 +727,144 @@ fn compute_blockhash(
...
+            // failed to reach is rebuilt from Postgres on the next boot.
+            Err(SettleError::AddressIndexWriterGone) => {
+                return Err(SettleError::AddressIndexWriterGone)
+            }
+            Err(e) => e,
+        };
+
+        // Re-read: cancellation may have arrived while the attempt was running,
+        // and a second attempt would then run past the node's drain deadline.
+        if shutdown_token.is_cancelled() || Instant::now() + backoff >= deadline {
```
sebatustra: here we return on the first failure whenever the token is cancelled. The retry loop at settle.rs:811 never runs a second iteration. So the 15 second budget (settle.rs:62) and the exponential backoff (settle.rs:68-70, settle.rs:862) are all dead code once shutdown starts.

That would be reasonable if the drain had no time to spare, but it has plenty. DRAIN_DEADLINE is 8 seconds (core/src/nodes/node.rs:37). A single 2 second attempt leaves roughly 6 seconds unused.

The consequence is severe because a failed settle does not just skip a block. It throws the whole buffer away and stops the stage:

```
  // core/src/stages/settle.rs:532-544
  Err(SettleError::Other(msg)) => {
      error!("Failed to settle transactions: {}", msg);
      // Cleared here so the final flush below does not
      // spend a second budget on the same doomed batch.
      discard_buffer(                       // 536: drops the buffer AND the queue
          &msg,
          &mut processing_results,
          &mut execution_results_rx,
          &metrics,
      );
      metrics.settler_buffered_account_bytes(0);
      break;                                // 543: leaves the worker loop
  }
```

core/src/nodes/node.rs
```diff
@@ -474,25 +488,132 @@ impl NodeHandles {
...
+                    // stopping it, so an in-process restart would leave this
+                    // stage holding its pool and channels beside the new one.
+                    worker.handle.abort();
+                    overran = true;
+
+                    // `abort()` only schedules cancellation, so wait for the task
+                    // to actually end. Bounded, because a task that never yields
+                    // cannot be cancelled at all and must not hold the drain open
+                    // past the container's stop grace period.
+                    let stopped = tokio::time::timeout(ABORT_GRACE, &mut worker.handle)
```
sebatustra: finding: The shutdown budget can exactly consume the container stop grace, so the process is killed mid-drain


  Two separate constants:

 ```
  pub const DRAIN_DEADLINE: Duration = Duration::from_secs(8);
  const ABORT_GRACE: Duration = Duration::from_secs(2);
```

DRAIN_DEADLINE is a single shared budget for all workers (node.rs:507, node.rs:510). ABORT_GRACE is not:
```
  Err(_) => {
      worker.handle.abort();               // 518
      overran = true;
      let stopped = tokio::time::timeout(ABORT_GRACE, &mut worker.handle)   // 525: per worker
          .await
          .is_ok();
```

Line 525 runs inside the for loop at node.rs:509, after remaining at node.rs:510 has already reached zero. So it is spent once per overrunning worker, on top of the 8 second deadline.

Why it is wrong

The comment at node.rs:35-36 claims the 8 seconds "still leaves room under the container default stop grace period". The branch's own test states what that period is:

```
  // integration/tests/private-channel/test_ordered_shutdown_e2e.rs:36-38
  const CONTAINER_STOP_GRACE: Duration = Duration::from_secs(10);
```
8 + 2 = 10. One stuck worker consumes the entire window, leaving zero seconds for the process to exit. There is no compose file in the repo that raises the grace period.

An Aio node has 10 workers. Once the deadline expires, every remaining worker gets remaining == 0 at node.rs:510, so each one falls into the Err(_) arm and each one can spend up to ABORT_GRACE. Worst case: 8 + 2 × 10 = 28 seconds.

In practice most tasks are cancelled quickly because abort() lands at the next await point, so 2 or 3 seconds of accumulation is the realistic case. That is still enough to blow past 10 seconds.

File: core/src/stages/settle.rs [hunk 8]
```diff
@@ -687,7 +727,144 @@ fn compute_blockhash(
...
 
+/// Signatures named per log line, so one discard cannot emit a single
+/// unreadable record while still naming every transaction it dropped.
+const DISCARD_SIGNATURES_PER_LINE: usize = 100;
+
+/// Record executed transactions that could not be committed, then drop them.
+///
+/// The only place a buffer clears without a commit, so the record is written
+/// first. Queued results fold in; a DB-backed one would need the dead database.
+fn discard_buffer(
```
sebatustra, line 739: finding: discard_buffer misses batches that arrive while it is draining, so they vanish with no record

discard_buffer is the one place a buffer is thrown away without a commit. It pulls everything queued behind the buffer so the record covers those too.

Why it is wrong: The loop at settle.rs:744 is a snapshot, not a close. Every try_recv frees one slot in the bounded channel, and the executor is very likely parked on a send waiting for exactly that slot:

```
  async fn send_one(results_tx: &mpsc::Sender<ExecutedBatch>, batch: ExecutedBatch) -> SendOutcome {
      match results_tx.send(batch).await {     // 372: awaits with no cancellation
          Ok(()) => SendOutcome::Sent,
          Err(_) => SendOutcome::ChannelClosed,
      }
  }
```

That send has no shutdown escape any more. The SendOutcome::ShuttingDown variant was deliberately removed (core/src/stages/execution.rs:332-338). So the executor sits on execution.rs:372 until a slot opens, which settle.rs:744 provides.

Once discard_buffer returns, settle.rs:543 breaks the loop and run_settle_worker returns at settle.rs:684.
execution_results_rx is dropped there. Anything the executor pushed in between is dropped by the channel's own Drop, with no signature list and no counter increment

core/src/nodes/node.rs
Huzaifa696: Fixed in bbd9a97: `wait_for_any_worker_quit` now removes the worker whose output that poll consumed, so `shutdown` never re-polls it (confirmed the panic first, `core.rs:422`); note this predates the PR, pre-change `shutdown` polled the same list after the same `select_all`.

core/src/stages/settle.rs
Huzaifa696: Fixed in bbd9a97: a cancelled settle now retries within a `SETTLE_SHUTDOWN_BUDGET` (4s) measured from cancellation, with the attempt cap at 1500ms so two attempts fit; you were right that the drain has time to spare, though `DRAIN_DEADLINE` is shared by all workers rather than the settler alone, so the budget is sized explicitly instead of assuming the remainder.

core/src/nodes/node.rs
Huzaifa696: Fixed in bbd9a97: the reserve is now a single shared deadline (`ABORT_RESERVE`) and `DRAIN_DEADLINE` drops to 6s, so the whole shutdown is 8s no matter how many workers overrun; a regression test with six stuck workers measured 20.0s before the change.

core/src/stages/settle.rs
Huzaifa696: Fixed in bbd9a97: `discard_buffer` closes the receiver before draining so nothing can land behind the snapshot, and the executor now records the batch a closed queue hands back plus any chunks that never left, which also closes the upstream half of the asymmetry.

sebatustra (approved):

PR #289, opened 2026-08-31
fix(core): read the settler's resume point from the tip block row (issue-136)

Huzaifa696:
## Summary

The settle worker decided where to resume block production by assembling the chain tip from two tables: `MAX(slot) FROM blocks` for the height, `metadata.latest_blockhash` for the hash. `get_latest_blockhash` encodes genuine absence as an error, so the caller could not tell "no blocks yet" from "the read failed" and settled for `.ok()`, mapping every database error to `None`. Both reads absent means `last_block = None`, which `settle_transactions` treats as genesis: slot 0 with a default hash, written through blind upserts with no parent check. Live today, the two tables can also disagree, and the settler then chains the next block off the metadata hash instead of the hash the tip block carries, writing a broken `previous_blockhash` link into history.

The settler now reads one authoritative row, `SELECT slot, data FROM blocks ORDER BY slot DESC LIMIT 1`, decoded into the existing `BlockInfo`. `get_tip` returns `Result<Option<BlockInfo>, TipError>`, so `Ok(None)` means the ledger genuinely holds no blocks and is the only thing readable as genesis; every read or decode failure propagates. Slot and hash come from one row and cannot drift, and returning the row proves it exists. Emptiness is proven, not assumed: with no block row, `get_tip` rejects a ledger whose metadata still names a tip. It takes `PostgresAccountsDB`, because a writer's resume point must never come from a cache.
### Coverage Report

| Core | 17,831 | 19,550 | 91.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33405508108) |
| Indexer | 37,606 | 41,281 | 91.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33405508108) |
| Gateway | 1,634 | 1,892 | 86.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33405508108) |
| Auth | 1,312 | 1,506 | 87.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33405508108) |
| Withdraw Program | 129 | 241 | 53.5% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33405508142) |
| Escrow Program | 1,195 | 1,982 | 60.3% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33405508142) |
| DvP Swap Program | 270 | 1,179 | 22.9% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33405508142) |
| E2E Integration | 15,114 | 19,514 | 77.5% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33405508108) |
| **Total** | **74,821** | **85,966** | **87.0%** | |

> Last updated: 2026-08-31 15:41:13 UTC by E2E Integration


sebatustra (commented):
implementation looks correct to me but are we not missing the write part of the auditors' suggestion? 

> The block commit should also atomically check the expected parent slot or blockhash before upserting a new block

I see no changes regarding that part, is that intentional?

sebatustra (approved):

PR #290, opened 2026-08-31
fix(core): separate slot from block height with an idle heartbeat block (issue-236)

Huzaifa696:
## Summary

The settler previously produced and stored a block on every 100 ms tick, so an idle node could write ~864,000 rows per day. The tick cannot simply be skipped because it also mints the blockhash that feeds the deduplication window. `slot` therefore remains the tick counter, while `block_height` now counts blocks actually produced: one for each tick that carries work, plus one heartbeat block per second while idle. Blockhashes are minted per produced block, making `lastValidBlockHeight` the last block height at which the blockhash remains valid. `getSlot` still reports the live tick, using a durable counter that suppressed ticks update without creating block rows. This keeps slot progression consistent across node types while reducing idle block production by roughly 10x.

The operator's finality classifier previously derived block height from `context.slot`, which was only correct while slot and block height were equal. It now reads `getBlockHeight`; without this change, live withdrawal broadcasts could be incorrectly treated as expired. Blocks now set `parent_slot` to the previous stored block, which the indexer's continuity walk requires, and the indexer now ends each walk at a produced block rather than at the chain tip, which usually has no block. Cache continuity is also checked against the parent block, since a heartbeat block can be ten slots after the block it extends.

`getSlot` can now report slots for which no block exists, and `getBlock` returns `-32007` for those slots, matching Solana's behavior rather than returning `null` and leaving clients retrying indefinitely. `transaction_expiration_ms` now maps to `max_blockhashes`, which remains 150.
### Coverage Report

| Core | 20,629 | 22,371 | 92.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33871600274) |
| Indexer | 39,077 | 42,803 | 91.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33871600274) |
| Gateway | 1,648 | 1,907 | 86.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33871600274) |
| Auth | 1,312 | 1,506 | 87.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33871600274) |
| E2E Integration | 15,666 | 20,441 | 76.6% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33871600274) |
| **Total** | **78,332** | **89,028** | **88.0%** | |

> Last updated: 2026-09-04 13:14:13 UTC by E2E Integration


core/src/stages/settle.rs
```diff
@@ -455,10 +544,23 @@ pub async fn start_settle_worker(args: SettleArgs) -> WorkerHandle {
...
                         let num_results = processing_results.len();
+                        if !should_produce_block(
+                            num_results > 0,
+                            last_block.is_none(),
+                            tick.saturating_duration_since(last_block_at),
+                        ) {
+                            // The chain is on this slot even though no block
+                            // backs it. Only reached when the tick carried no
+                            // work, so the loaded path never pays for it.
+                            publish_idle_slot(&postgres_db, redis_db.as_ref(), next_slot).await;
```
sebatustra: aren't we still committing to postgres on every tick? the reduction is real for rows but not of writes

indexer/src/indexer/datasource/rpc_polling/rpc.rs
```diff
@@ -383,8 +385,20 @@ impl RpcPoller {
             return Ok((vec![], latest_slot));
         }
 
-        let to_slot = std::cmp::min(from_slot + max_slots as u64, latest_slot);
-        Ok(((from_slot..to_slot).collect(), latest_slot))
+        // The same window the batch would have covered, so the enumeration here is
+        // the one the walk needs rather than an extra range.
+        let window_end = std::cmp::min(from_slot + max_slots as u64, latest_slot).saturating_sub(1);
```
sebatustra, line 401: When the window holds no produced block, the poller returns nothing, sleeps, and re-asks the same range. from_slot never advances.

That's correct when blocks are close together, but it creates a dependency that did not exist before this branch: batch_size must be at least HEARTBEAT_INTERVAL / blocktime_ms. Nothing validates it. The two numbers live in different services' configs (blocktime_ms on the node, batch_size in indexer-private-channel.toml) with no shared constant and no note in the docs.

At defaults the gap is 10 slots and batch_size = 50, so there's 5x headroom today.

Example. An operator sets --blocktime-ms 10 to cut latency. The idle gap becomes 100 slots. The 50-slot window can no longer contain a heartbeat block, so get_slots_to_process returns empty on every poll, forever. The indexer stops and logs nothing; it looks exactly like "caught up."

core/src/stages/settle.rs
Huzaifa696: Yes, intentional: issue 236 is storage growth, not IOPS, and the idle write is a single-row metadata upsert that never grows the table.

indexer/src/indexer/datasource/rpc_polling/rpc.rs
Huzaifa696: Good catch, fixed in dbcf538: an empty window now looks past itself with getBlocksWithLimit, so batch_size caps the batch rather than deciding whether the poller can advance and the coupling stops existing.

core/src/stages/settle.rs
```diff
@@ -52,6 +52,33 @@ pub const CACHE_MIRROR_COOLDOWN: Duration = Duration::from_secs(60);
...
+    postgres_db: &PostgresAccountsDB,
+    redis_db: Option<&RedisAccountsDB>,
+    slot: u64,
+) {
+    if let Err(e) = crate::accounts::current_slot::set_current_slot(postgres_db, slot).await {
+        warn!("Failed to publish the idle slot: {:#}", e);
+        return;
+    }
+    if let Some(redis) = redis_db {
+        crate::accounts::current_slot::mirror_current_slot(redis, slot).await;
```
sebatustra: should we give the idle mirror the same gates the batch mirror has for redis?

"mirroring" is the settler copying each committed batch into redis, and the code wraps that in three separate protections:

```
if let Some(redis) = redis_db {
    let mirrored = tokio::time::timeout(          // (1) 100ms hard cap
        CACHE_MIRROR_BUDGET,
        mirror_batch_to_cache(redis, next_slot, block_info.parent_slot, ...),
    ).await;

    let failed = match mirrored { ... };
     if failed {
        metrics.redis_cache_write_failed();
        redis.record_cache_failure();             // (2) failure tally
    }
  }

async fn mirror_batch_to_cache(...) -> MirrorOutcome {
    if redis.is_mirroring_paused() {              // (3) skip a cache given up on
        return MirrorOutcome::Skipped;
    }

```

the new idle path has none of the three, and it also continues past the give-up check above

Fix. Give the idle mirror the same two gates the batch mirror has

core/src/rpc/get_block_impl.rs
```diff
@@ -13,14 +13,27 @@ pub async fn get_block_impl(
...
-        return Ok(None);
+        // Solana reports a slot it passed without producing as skipped, and one
+        // it has not reached as not available. Most slots here carry no block, so
+        // a null would leave a stock client retrying the majority of them.
+        let tip = read_deps
+            .accounts_db
+            .get_current_slot()
+            .await
+            .map_err(|e| custom_error(JSON_RPC_SERVER_ERROR, format!("Failed to get slot: {}", e)))?
+            .unwrap_or(0);
```
sebatustra: getBlock(0) on a node that has produced nothing answers "skipped" instead of "not available"

The problem. On a fresh node, get_current_slot() returns None — no current_slot row, and the fallback MAX(slot) FROM blocks finds nothing either. unwrap_or(0) turns "there is no chain" into "the tip is slot 0", and slot 0 then satisfies slot <= tip.

fresh node, no blocks, settler not yet past its 1s start delay:
    getBlock(0)   → -32007  "Slot 0 was skipped, or missing due to ledger jump"   ← wrong
    getBlock(1)   → -32004  "Block not available for slot 1"                      ← right
    getBlock(999) → -32004                                                        ← right

Slot 0 is genesis — the one slot on this chain that is never skipped (settle.rs:16 forces a block at genesis). Telling a client it was skipped tells them not to retry the one slot that is about to exist a moment later. Small window (the settler produces genesis on its first tick, 1 s after start) and a small blast radius, but it's the exact opposite of the truth.

Fix. Distinguish "no chain" from "tip is slot 0" instead of flattening both to 0

core/src/stages/settle.rs
Huzaifa696: Gated now: the idle mirror skips a paused cache and is bounded by CACHE_MIRROR_BUDGET, so a stalled Redis can no longer hold the settler tick.

core/src/rpc/get_block_impl.rs
Huzaifa696: Fixed: the tip read matches on the Option now, so with no tip every slot including genesis answers -32004 rather than skipped.

docs/RPC_COMPATIBILITY_ANALYSIS.md
```diff
@@ -49,25 +49,65 @@ Ordered from most divergent → closest match.
...
-| `getSlot` | `minContextSlot` ignored. Returns latest stored slot or 0. |
-| `getBlockHeight` | `minContextSlot` ignored. Returns the latest stored slot or 0, the same read as `getSlot`: SPC keeps slot and block height equal by construction, so the two can never disagree. This is the clock a client compares a stored `lastValidBlockHeight` against to prove a status-less signature can no longer land. |
+| `getSlot` | `minContextSlot` ignored. Returns the live slot, or 0 on a fresh node. The settler ticks it every `blocktime_ms` and publishes it whether or not that tick produced a block, so it advances continuously while the node is idle and stays monotonic across a restart. **Most slots carry no block.** An idle node produces one block per heartbeat, so about nine in ten slots are empty and `getBlock` on one returns `null`, exactly as it does for a skipped slot on Solana. Use `getBlocks`/`getBlocksWithLimit` to find slots that do hold a block, and `getBlockHeight` for the count of blocks produced. |
```
sebatustra: getBlock contract is documented two contradictory ways in the same doc

docs/RPC_COMPATIBILITY_ANALYSIS.md:52 vs :63, and docs/CONFIG.md:41

This branch changes getBlock on an empty slot from null to an error. Line 63 documents it correctly:

A slot with no block is an error, not a null, matching Solana: -32007 (SlotSkipped) … -32004 (BlockNotAvailable)

But line 52 — the getSlot row, which is where an integrator lands when they read "most slots carry no block" — still says the opposite:

 …about nine in ten slots are empty and getBlock on one returns null, exactly as it does for a skipped slot on Solana.

  And docs/CONFIG.md:41:

 …so most slots carry no block and getBlock on one returns null.

Two problems. First, the code returns an error, never null (core/src/rpc/get_block_impl.rs:36-39 — the Ok(None) arm is gone entirely). Second, the trailing clause is wrong about Solana too: Agave returns -32007 for a skipped slot, not null. So a client author reading the getSlot row writes if (block === null) retry() and gets an exception instead — on ~90% of slots.

Fix: replace both null mentions with -32007, and drop "exactly as it does for a skipped slot on Solana" from :52 or restate it as "as -32007, matching Solana".

indexer/src/indexer/datasource/rpc_polling/rpc.rs
```diff
@@ -5,7 +5,12 @@ use serde_json::json;
 use solana_sdk::commitment_config::CommitmentLevel;
 use solana_transaction_status::UiTransactionEncoding;
 use std::collections::HashMap;
-use tracing::debug;
+use tracing::{debug, warn};
+
+/// How far past an empty window the poller reaches for the next producer. The
+/// widest legal idle gap is a thousand slots, so this leaves tenfold room while
+/// keeping a hole in the ledger out of a batch that carries a slot per entry.
+const MAX_LOOKAHEAD_SLOTS: u64 = 10_000;
```
sebatustra: The batch_size coupling from review comment #2 was moved, not removed: two new unvalidated constants, one of which fails open

dbcf538 correctly removed the hard stall where batch_size decided whether the poller could advance. But the fix introduced two new constants that encode the same assumption about the node's idle gap, in two different files, neither derived from nor validated against blocktime_ms:

`  const MAX_LOOKAHEAD_SLOTS: u64 = 10_000;
`
`  const PRODUCER_LOOKBACK_SLOTS: u64 = 1_000;
`
"The widest legal idle gap is a thousand slots" is not true of anything in the codebase. The real widest gap is HEARTBEAT_INTERVAL / blocktime_ms (core/src/stages/settle.rs:58) — 10 at defaults, and unbounded as blocktime_ms shrinks. Nothing in run_node validates the ratio (core/src/nodes/node.rs:192-203 checks only blocktime_ms != 0 and max_blockhashes !=  0).

Fix: (a) derive both constants from the node's advertised cadence or at minimum cross-reference HEARTBEAT_INTERVAL in the comments and correct the "thousand slots" claim; (b) make the unwrap_or(tip) / Err arms return Result and let the caller fail with a named error rather than silently reproducing the unwitnessable-tail case.

docs/RPC_COMPATIBILITY_ANALYSIS.md
Huzaifa696: Fixed: the getSlot row and CONFIG.md both say -32007 now, and the Solana null comparison is gone.

indexer/src/indexer/datasource/rpc_polling/rpc.rs
Huzaifa696: Both bounds now derive from one MAX_IDLE_GAP_SLOTS and an unreadable listing errors; an empty window still returns the tip, since the walk witnesses that tail from above.

core/src/stages/settle.rs
```diff
@@ -52,6 +52,51 @@ pub const CACHE_MIRROR_COOLDOWN: Duration = Duration::from_secs(60);
...
+/// Record a tick that produced no block, so `getSlot` keeps moving while the
+/// node is idle. Best effort: the in-memory counter is what production uses, and
+/// a failure here costs a stale published slot, never a wrong block.
+async fn publish_idle_slot(
+    postgres_db: &PostgresAccountsDB,
+    redis_db: Option<&RedisAccountsDB>,
+    slot: u64,
+) {
+    if let Err(e) = crate::accounts::current_slot::set_current_slot(postgres_db, slot).await {
+        warn!("Failed to publish the idle slot: {:#}", e);
```
sebatustra: The idle-slot Postgres write is the only unbounded await on the settler's tick path, and its failure is warn-only. The Redis half of this path got both gates after review feedback. The Postgres half got neither a timeout nor a failure path.

docs/CONFIG.md
```diff
@@ -152,6 +158,14 @@ Truncation deletes in batches, and each batch commits the new retention floor
 (`getFirstAvailableBlock`) in the same transaction as its deletions. A partial or
 aborted run therefore never advertises history it has already removed.
 
+**`--keep-slots` is a slot span, not a block count.** Blocks are sparse relative
```
sebatustra, line 161: Postgres has a retention floor nothing enforces.
fix:  Two cheap guards: warn (or refuse) in truncate_slots when keep_slots < 10 * max_blockhashes, and log at warn in load_dedup_state when blocks.len() < max_blockhashes naming both numbers. The second is one line and turns a silent transaction-dropping window into a startup log line.

core/src/stages/settle.rs
Huzaifa696: Bounded now: the idle write gets a 100ms budget like the mirror, and a timeout or error is warned and counted in a new settler_idle_slot_publish_failed metric; the heartbeat block one second later carries the failure into the retry budget, so /health stays with the path that owns it.

docs/CONFIG.md
Huzaifa696: Both guards added: truncate takes --max-blockhashes and --blocktime-ms and warns when keep_slots is below the derived floor, and boot warns naming both numbers when the window rebuilt short and genesis is not in it.

sebatustra (approved):

PR #291, opened 2026-09-01
fix(core): reject slot commits that do not extend the stored ledger (issue-134)

Huzaifa696:
## PR description

`write_batch_postgres` committed a slot with unconditional upserts and carried no
writer identity, and the settler reads the tip once at startup then derives every
later slot in memory. Two write-capable nodes against one primary therefore both
build slot N+1 from the same parent: the later `blocks` upsert wins, while account
rows, transaction rows and the transaction count keep effects from both. A reader
could see one writer's block beside another writer's state, a ledger no serial
execution produced. Dedup is per-process, so both nodes also execute the same
signed transaction.

The block insert is now conditional on no block existing at or above its slot, and
`rows_affected() == 0` aborts the whole batch. The slot primary key covers the
simultaneous case: both writers clear the guard and one aborts on `23505`. Guarding
on the block's own slot rather than `parent_slot` was deliberate, since gaps are
unreachable from the settler and enforcing contiguity would have broken two
existing rpc fixtures. A writer lease, a session advisory lock on a dedicated
connection, refuses a second write or aio node at startup and re-proves ownership
every 5s, stopping the node when it cannot; without that the primary-key race picks
the survivor arbitrarily and can leave the live writer holding no lease.
`transaction_count` keeps its read-modify-write, now safe because only one writer
can commit a slot.
### Coverage Report

| Core | 21,224 | 23,058 | 92.0% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34110655136) |
| Indexer | 39,077 | 42,803 | 91.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34110655136) |
| Gateway | 1,648 | 1,907 | 86.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34110655136) |
| Auth | 1,704 | 1,923 | 88.6% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34110655136) |
| E2E Integration | 15,964 | 20,674 | 77.2% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34110655136) |
| **Total** | **79,617** | **90,365** | **88.1%** | |

> Last updated: 2026-09-07 11:17:55 UTC by E2E Integration


core/src/nodes/node.rs
```diff
@@ -199,6 +202,16 @@ pub async fn run_node(config: NodeConfig) -> Result<NodeHandles, Box<dyn std::er
     // Create a single shutdown token for all services
     let shutdown_token = CancellationToken::new();
 
+    // Taken after config validation so a bad config still fails without touching
+    // Postgres, and before any write-path work so a duplicate node is refused
+    // before it repairs indexes or serves RPC. Losing the lease later cancels the
+    // same token, so the node stops rather than running on without it.
+    let writer_lease = if matches!(config.mode, NodeMode::Write | NodeMode::Aio) {
+        Some(WriterLease::acquire(&config.accountsdb_connection_url, shutdown_token.clone()).await?)
```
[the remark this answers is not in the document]
Huzaifa696: Good catch, fixed in 50db0bc. `WriterLease` now implements `Drop`, which cancels the stop token so the heartbeat unlocks and closes the session even when the lease is dropped on a failed startup path. Pinned by `writer_lease_dropped_without_release_still_frees_the_lock`.

core/src/nodes/node.rs
```diff
@@ -479,13 +493,26 @@ impl NodeHandles {
...
                 Ok(Err(e)) => error!("{} error: {:?}", worker.name, e),
                 Err(_) => warn!("{} shutdown timeout", worker.name),
             }
         }
 
+        // Released after the workers, so a settler still draining keeps the lease.
+        // A worker that outlives the join timeout is the gap: a replacement node
+        // could take the lease while that worker is still committing.
+        if let Some(lease) = self.writer_lease {
+            lease.release().await;
```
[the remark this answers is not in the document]
Huzaifa696: Agreed, fixed in 50db0bc. `shutdown` now tracks whether every worker joined in time and only releases the lease when they did; a timed-out worker is detached rather than stopped, so the lease is held until the process exits instead. Pinned by `shutdown_keeps_the_lease_when_a_worker_outlives_the_join_timeout` and its clean-path counterpart.

core/src/accounts/writer_lease.rs
```diff
@@ -0,0 +1,290 @@
...
+            .context("Failed to open the writer lease connection")?;
+
+        let acquired: bool = sqlx::query_scalar("SELECT pg_try_advisory_lock($1)")
+            .bind(WRITER_LEASE_LOCK_ID)
+            .fetch_one(&mut conn)
+            .await
+            .context("Failed to acquire the writer lease")?;
+
+        if !acquired {
+            return Err(anyhow!(
```
sebatustra, line 278: finding: A write node that dies ungracefully holds the lease for ~2 hours, and nothing can start in its place

A session advisory lock lives exactly as long as its Postgres backend. That backend lives as long as the TCP connection. When the write node exits, crashes, or is SIGKILLed, the kernel closes the socket, Postgres sees EOF, and the lock frees in milliseconds — that case is fine.

The case that is not fine is the host disappearing: a hypervisor kill, a kernel panic, a network partition, a k8s node going NotReady. No FIN is ever sent. The backend stays blocked in recv() holding the lock, and Postgres only notices when TCP keepalives expire. tcp_keepalives_idle defaults to 0 (use the OS default), which on Linux is net.ipv4.tcp_keepalive_time = 7200 plus 9 probes at 75s — about 2 hours 11 minutes.

For that whole window every replacement write node dies at run_node (core/src/nodes/node.rs:209-213) with "Another write-capable node already holds the writer lease". The gateway routes sendTransaction to a single write URL (docs/CONFIG.md:59), so the deployment is write-dead until an operator finds the zombie backend and kills it by hand — and there is nothing in docs/ telling them that, or that this is what happened.

core/src/accounts/writer_lease.rs
```diff
@@ -0,0 +1,290 @@
...
+        let reason = match verdict {
+            Ok(Ok(true)) => continue,
+            Ok(Ok(false)) => "pg_locks reports the lease is no longer held",
+            // A checked-out connection never heals, so an error means the session ended.
+            Ok(Err(_)) => "the probe query failed on the lease session",
+            Err(_) => "the probe did not answer within the timeout",
+        };
+        error!("Writer lease ownership could not be proven ({reason}); stopping the node");
+        node_shutdown.cancel();
+        return;
```
sebatustra, line 214: finding: A false-positive probe closes the lease session, freeing the lock while the node is still draining — and this is what makes hold() a no-op

```
  error!("Writer lease ownership could not be proven ({reason}); stopping the node");
  node_shutdown.cancel();
  return;                       // <-- `conn` is dropped here
```

That return drops the owned PgConnection. The socket closes, the backend exits, and the advisory lock is genuinely released — including in the probe_timeout case, where the lock was never actually lost.

Now trace what the node is doing at that moment. node_shutdown.cancel() starts a drain that keeps writing:

  - the settler breaks out of its loop and runs a final flush that commits another whole block (settle.rs:599-641);
  - the address_index_writer keeps flushing address_signatures rows;
  - NodeHandles::shutdown allows 5 seconds per worker, sequentially (node.rs:526).

So for several seconds the node commits slots with the lock already handed back. A replacement can acquire it, get_tip (settle.rs:343-349), and start producing from a tip the draining node is about to move. Both then target the same slot, theguard kills one — and if it kills the replacement, you get a crash loop against a node that is supposedly shutting down.

core/src/accounts/writer_lease.rs
```diff
@@ -0,0 +1,290 @@
...
+        let verdict =
+            tokio::time::timeout(LEASE_PROBE_TIMEOUT, lock_is_still_held(&mut conn)).await;
+        let reason = match verdict {
+            Ok(Ok(true)) => continue,
+            Ok(Ok(false)) => "pg_locks reports the lease is no longer held",
+            // A checked-out connection never heals, so an error means the session ended.
+            Ok(Err(_)) => "the probe query failed on the lease session",
+            Err(_) => "the probe did not answer within the timeout",
+        };
+        error!("Writer lease ownership could not be proven ({reason}); stopping the node");
```
sebatustra, line 207: finding: Lease loss and duplicate-writer rejection are log-only, in a repo that already has the metric, the alert and the runbook for the identical mechanism

where: core/src/accounts/writer_lease.rs:76-83, core/src/accounts/write_batch.rs:33-40

The heartbeat computes a perfectly good reason — "pg_locks reports the lease is no longer held", "the probe query failed on the lease session", "the probe did not answer within the timeout" — and throws it into a log line. Those are, almost word for word, the reason label values the indexer emits (docs/runbooks/sender_lock_lost_runbook.md:45-51: not_held / probe_error / probe_timeout).

The indexer's version of this mechanism has:

  - a counter, private_channel_operator_sender_lock_lost_total, labelled by reason (indexer/src/metrics.rs:225-244, incremented at lock_connection.rs:220-229);
  - a critical Grafana alert;
  - a 170-line runbook with the triage SQL, the pgbouncer caveat, and the escalation path.

core/src/accounts/write_batch.rs
```diff
@@ -216,8 +228,9 @@ async fn write_batch_postgres(
         .map_err(|e| format!("Failed to bulk upsert transactions: {}", e))?;
     }
 
-    // Read-modify-write inside BEGIN…COMMIT: safe because all writers serialize
-    // via this path and MVCC returns the caller's own last commit.
+    // Read-modify-write inside BEGIN…COMMIT: the settler always extends the tip by
```
sebatustra: finding: transaction_count is still a read-modify-write; its safety now rests on an invariant enforced 30 lines later and in  another module

The original finding recommended: "Make transaction-count updates atomic as well, for example with a locked row or SQL-side increment, instead of read-modify-write replacement." This PR kept the RMW and replaced the justification:

// Read-modify-write inside BEGIN…COMMIT: the settler always extends the tip by
// one slot, so the block insert below lets only one writer commit that slot and
// a rejected writer's increment rolls back with the rest of its batch.

The argument is correct today: next_slot = lb.slot + 1 unconditionally (settle.rs:717-720), settle.rs:834 is the only non-test caller of write_batch, and the loser's INSERT … ON CONFLICT DO UPDATE SET value = $1 — which does blindly overwrite the winner's newer count under READ COMMITTED — is rolled back with its transaction.

But look at what the safety of a counter now depends on:

  1. the block insert is reached at all — it is gated on if let (Some(block_info), Some(block_data) (write_batch.rs:259), and the counter's if tx_count > 0 block is not;
  2. the settler never skips a slot;
  3. settle.rs remains the sole production caller.

None of those is checked where the counter is written. Point 1 is the sharp one: write_batch(&settlements, txs, None) is a valid call on a pub API (core/src/accounts/traits.rs:149) that increments the durable counter and upserts accounts and transactions with no guard at all. It's test-only today; nothing marks it as unsafe tomorrow.

An SQL-side increment removes all three dependencies and is shorter than the code it replaces:

```
  INSERT INTO metadata (key, value) VALUES ('transaction_count', $1)
  ON CONFLICT (key) DO UPDATE
    SET value = <count(metadata.value) + $2 encoded>
```

If TransactionCount's encoding makes that awkward in SQL, SELECT … FOR UPDATE on the row costs one word and makes the RMW correct on its own terms rather than by appeal to a caller.

core/src/accounts/writer_lease.rs
Huzaifa696: Fixed: the lease session now sets its own TCP keepalives (60s idle, 3 probes 15s apart), so a vanished host is reaped in under two minutes instead of ~2h11m, and `docs/runbooks/writer_lease_unavailable.md` covers the zombie-backend triage and the pgbouncer caveat.

core/src/accounts/writer_lease.rs
Huzaifa696: Fixed: an unprovable verdict now parks the heartbeat instead of returning, so the session (and the lock) stays until shutdown reports its workers stopped, which is also what makes `hold()` real; pinned by `an_unprovable_probe_keeps_the_lease_session_until_it_is_stopped`.

core/src/accounts/writer_lease.rs
Huzaifa696: Agreed and deferred to a follow-up: the counter, Grafana alert and runbook are an additive observability layer over reason strings that already match the indexer labels, and I would rather not fold them into a branch whose subject is the lock itself.

core/src/accounts/write_batch.rs
Huzaifa696: Agreed, deferred: the counter is now skipped on a replayed slot and still rides on the block insert one statement above it, so an SQL-side increment is the right removal of that dependency but belongs with the surviving upsert rather than here.

core/src/accounts/writer_lease.rs
```diff
@@ -0,0 +1,451 @@
...
+    /// while a detached worker can still commit.
+    pub fn hold(self) {
+        std::mem::forget(self);
+    }
+}
+
+impl Drop for WriterLease {
+    /// A lease dropped on an error path still has to free the lock, so cancel and
+    /// let the heartbeat close the session on its way out.
+    fn drop(&mut self) {
```
sebatustra: Dropping NodeHandles without shutdown() releases the writer lease while every worker keeps running. 
The fix that matches the rest of the design: make the drop path as conservative as the still_running path. Either don't implement Drop on WriterLease at all (force release() / hold(), so a dropped lease keeps the lock until the process exits), or give NodeHandles a Drop that calls lease.hold() on the lease it still owns. The startup error path in run_node:246-254  already routes through shutdown(), so it no longer needs Drop to clean up for it.

core/src/accounts/writer_lease.rs
```diff
@@ -0,0 +1,451 @@
...
+        // leave the connection unusable.
+        tokio::select! {
+            biased;
+            _ = stop.cancelled() => break,
+            _ = tokio::time::sleep(interval) => {}
+        }
+
+        let verdict =
+            tokio::time::timeout(LEASE_PROBE_TIMEOUT, lock_is_still_held(&mut conn)).await;
+        let reason = match verdict {
```
sebatustra: One slow pg_locks probe takes the whole write side of the deployment down, with no retry and no metric. A single SELECT ... FROM pg_locks that takes more than 2 seconds stops the only write-capable node in the deployment. LEASE_PROBE_TIMEOUT is 2s and there is no tolerance — not "two consecutive failures", not one retry. pg_locks takes lightweight locks across every lock partition, and this query runs every 5 seconds forever against the primary that is simultaneously absorbing the settler's commits. A checkpoint storm, a lock-table scan behind a long-running VACUUM, or a  2-second network stall is enough.

core/src/accounts/writer_lease.rs
Huzaifa696: Fixed by removing `Drop` entirely rather than adding one to `NodeHandles`: only `release()` frees the lock now, so a dropped or leaked lease behaves like `hold()` and every owner is covered, pinned by `a_dropped_lease_keeps_the_lock` and `dropping_node_handles_without_shutdown_keeps_the_lease`.

core/src/accounts/writer_lease.rs
Huzaifa696: Fixed at the root: the probe now runs on its own short-lived connection keyed to the lease backend's `(pid, backend_start)`, so it can no longer poison the lease session, and only a definitive `not_held` stops the node while an unanswered probe is retried for a 30s budget before `probe_unavailable`, with `writer_lease_probe_total{outcome}` and `writer_lease_lost_total{reason}` for the retries you noted were invisible.

sebatustra (dismissed):

sebatustra (approved):

PR #292, opened 2026-09-03
fix(auth): bound the listener instead of serving through bare axum::serve (issue-224)

sebatustra:
## finding-224: auth served through bare `axum::serve`

`axum::serve` has no connection limit and no timeouts, so a client could open sockets and sit idle or trickle headers indefinitely. The existing rate limits are charged per request, so nothing applied until a request arrived.

## Fix

`auth/src/serve.rs` replaces it with a configured hyper loop, matching `gateway::serve`: global and per-IP connection caps, `header_read_timeout` for slowloris, TCP keepalive, and backoff on accept errors. Each request is bounded
  by a middleware in `build_app` that sheds as 503 through `AppError`, so the response keeps its JSON body and CORS headers.

### Coverage Report

| Core | 19,587 | 21,270 | 92.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34103842373) |
| Indexer | 38,763 | 42,485 | 91.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34103842373) |
| Gateway | 1,648 | 1,907 | 86.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34103842373) |
| Auth | 1,704 | 1,923 | 88.6% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34103842373) |
| Withdraw Program | 129 | 241 | 53.5% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34103842361) |
| Escrow Program | 1,195 | 1,982 | 60.3% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34103842361) |
| DvP Swap Program | 270 | 1,179 | 22.9% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34103842361) |
| E2E Integration | 15,459 | 20,055 | 77.1% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34103842373) |
| **Total** | **78,485** | **89,863** | **87.3%** | |

> Last updated: 2026-09-07 09:54:49 UTC by E2E Integration


auth/src/lib.rs
```diff
@@ -127,10 +139,34 @@ pub fn build_app(state: AppState, cors_allowed_origin: &str) -> Router {
...
+/// every other failure here has; a bare status from a timeout layer would be
+/// the one response a browser could neither read nor parse.
+async fn enforce_request_timeout(
+    State(limit): State<Duration>,
+    request: Request,
+    next: Next,
+) -> Result<Response, AppError> {
+    tokio::time::timeout(limit, next.run(request))
+        .await
+        .map_err(|_| AppError::Unavailable)
```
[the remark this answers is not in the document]
sebatustra: `consume_challenge` is now generic over the executor like `insert_verified_wallet`, and the route runs both in one transaction, so nothing commits unless the wallet is stored.

auth/src/lib.rs
```diff
@@ -127,10 +139,34 @@ pub fn build_app(state: AppState, cors_allowed_origin: &str) -> Router {
...
         .layer(cors)
         .with_state(state)
 }
 
+/// Sheds a request that outran `limit`, whatever it was waiting on.
+///
+/// Reported as 503, not 408. The usual cause is server-side contention (the
+/// pool or the Argon2 queue), and 408 tells the client it may repeat the request
+/// unchanged, which turns a struggling database into a retry storm. 503 is
+/// already what this service returns when it sheds for the Argon2 cap.
```
Huzaifa696: This argues a shed must be 503 so clients don't treat it as a retryable-unchanged failure, but every sqlx::Error still maps to 500 here. Should PoolTimedOut/PoolClosed join the request-timeout and Argon2-cap sheds as 503? The 5s acquire_timeout makes this reachable much more often than before.

auth/src/serve.rs
```diff
@@ -0,0 +1,449 @@
...
+        };
+
+        // Enable OS TCP keepalive so a peer that vanishes without a close (its
+        // network dropped) is detected and the socket reclaimed. Best-effort:
+        // log and keep serving if it fails.
+        let keepalive = TcpKeepalive::new()
+            .with_time(limits.tcp_keepalive_idle)
+            .with_interval(limits.tcp_keepalive_interval);
+        if let Err(e) = SockRef::from(&stream).set_tcp_keepalive(&keepalive) {
+            warn!("Failed to set TCP keepalive for {ip}: {e}");
```
Huzaifa696: This warn! fires once per accepted connection when set_tcp_keepalive fails, which is the per-connection amplification deliberately avoids for refusals. Can it fold into the 10s periodic report, or log once?

auth/src/config.rs
```diff
@@ -42,4 +52,113 @@ pub struct Config {
     /// regardless of which IPs they come from.
     #[arg(long, env = "AUTH_USERNAME_ATTEMPTS_PER_MINUTE", default_value = "5")]
     pub auth_username_attempts_per_minute: NonZeroU32,
+
+    /// Maximum concurrent client connections. Connections beyond this are
+    /// dropped so a flood cannot exhaust sockets, tasks or memory before the
+    /// per-request throttle is ever reached.
+    #[arg(long, env = "AUTH_MAX_CONNECTIONS", default_value = "1024")]
```
Huzaifa696, line 59: With the common 1024 soft RLIMIT_NOFILE, accept() hits EMFILE before the semaphore ever fills, so the cap never engages and we fall through to warn-and-backoff at serve.rs:163-170. Should we lower the default, or set ulimits.nofile for the auth service in docker-compose.yml and note the requirement in the README?

auth/src/lib.rs
sebatustra: AppError::Db now sheds 503 for the errors that already flip /health unhealthy (Io, PoolTimedOut, PoolClosed, WorkerCrashed); 500 is left for a query the database answered and rejected.

auth/src/serve.rs
sebatustra: Fixed. Warns on the first failure only, via a bool beside the refusal counters.

auth/src/config.rs
sebatustra: fixed but with neither of the recommendations: as it turns out, write-node, read-node and gateway already set ulimits.nofile: 65536, and auth got the gateway's connection limits without its fd budget. Added the same block to both compose files plus a README line on the requirement.

Huzaifa696 (approved):

PR #295, opened 2026-09-07
fix(indexer): separate live workers from a destructive resync with a shared/exclusive advisory lock (issue-68)

Huzaifa696:
### Summary

**Problem.** `ResyncService::run` could drop and recreate tables while a live indexer or operator was still using the database. The indexer had no advisory lock, and the operator’s lock used a key that `DROP TABLE` could not affect. A resync could therefore delete pending transactions, release-signature journals, checkpoints, and nonce state while workers still believed they owned them.

**Fix.** Workers now hold a shared advisory lock for their entire lifetime, while resync takes the same lock exclusively. Postgres prevents resync from running while workers are active. The lock session is detached from the pool to avoid `Pool::close()` waiting for the 15s shutdown timeout. Ownership is re-checked every 5s and synchronously before the drop. Resync also creates the schema under the lock, requires `--destroy-existing-data`, and refuses to run during an unresolved reconciliation halt.

**Reliability.** The lock connection enables TCP keepalives so dead hosts do not leave shared locks blocking resync for hours. A timeout during the lock probe is treated as inconclusive rather than proof that the lock was lost, and retries continue within a 30s budget. Connection errors on the lock session itself still indicate a genuinely lost lock.
### Coverage Report

| Core | 22,444 | 24,324 | 92.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34535420542) |
| Indexer | 39,969 | 44,301 | 90.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34535420542) |
| Gateway | 1,662 | 1,957 | 84.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34535420542) |
| Auth | 1,762 | 1,982 | 88.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34535420542) |
| E2E Integration | 16,741 | 22,290 | 75.1% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34535420542) |
| **Total** | **82,578** | **94,854** | **87.1%** | |

> Last updated: 2026-09-10 23:06:26 UTC by E2E Integration


indexer/src/indexer/resync.rs
```diff
@@ -350,16 +409,31 @@ impl ResyncService {
...
+            result = backfill_service.run(instruction_tx.clone()) => {
+                result.map_err(|e| {
+                    error!(
+                        "Backfill service failed during resync from slot {} to {}: {}",
+                        genesis_slot, current_slot, e
+                    );
+                    e
+                })?;
+            }
+        }
```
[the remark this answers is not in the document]
Huzaifa696: The fill, the processor drain and the checkpoint cleanup are now one future under a single select against the lock token, so a loss during any of them aborts both writers and returns LiveStateLockLost. The existing mid-rebuild test also asserts no durable frontier survives the loss.

indexer/src/indexer/resync.rs
```diff
@@ -282,6 +334,13 @@ impl ResyncService {
...
+        // The heartbeat bounds a silently lost lock to one interval, which is fine for a
+        // worker that only has to stop. The drop is irreversible, so prove ownership
+        // synchronously here instead of trusting the last tick.
+        live_lock.ensure_held().await.inspect_err(|e| {
+            error!("Refusing to drop tables: {}", e);
+        })?;
+
         // Step 1: Drop existing tables
         info!("Dropping existing database tables...");
         self.storage.drop_tables().await.map_err(|e| {
```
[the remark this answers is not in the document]
Huzaifa696: drop_tables now runs on the lock-holding session via a fenced path, inside one transaction, so the drop cannot outlive the lock and a session lost part way rolls back instead of leaving a half-dropped schema. ensure_held() is kept only to turn an already-lost lock into a clean refusal, and its comment now says so.

indexer/src/storage/common/storage/live_lock.rs
```diff
@@ -0,0 +1,377 @@
...
+            heartbeat_interval,
+            LIVE_LOCK_UNCONFIRMED_BUDGET,
+            task_stop.clone(),
+            move |reason, detail| report_lock_lost(role, &on_lost, reason, detail),
+            move || {
+                let session = probe_session.clone();
+                async move { session.probe().await }
+            },
+        )
+        .await;
```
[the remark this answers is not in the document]
Huzaifa696: The live-state lock now has its own cancellation token in both the indexer and the operator, so a lost lock is distinguishable from a ctrl-c at the shutdown site. On that path both roles abort their writers, skip the graceful drain, and exit non-zero.

indexer/src/storage/common/storage/sender_lock.rs
```diff
@@ -46,28 +46,65 @@ async fn run_lock_heartbeat<P, Fut>(
...
-                return;
-            }
+            // Unanswered. Falls through to the budget below.
+            Err(_) => {}
         }
+
+        // A timeout says the server is slow, not that the lock is gone. Retry while the
+        // budget lasts: our session is still open, so it still holds the lock, and
+        // nothing the lock protects is at risk meanwhile.
+        if confirmed_at.elapsed() < budget {
```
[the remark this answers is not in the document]
Huzaifa696: The heartbeat now clamps both the sleep and the probe timeout to the budget remaining since the last confirmation, so 30s is a true upper bound and a hung probe is rejected at 30s rather than ~35s. The clamp applies only after an unanswered probe, since clamping unconditionally would spin on a zero budget or a busy connection.

indexer/src/indexer/indexer.rs
```diff
@@ -386,6 +410,30 @@ pub async fn run(
...
+    // either. Refusing here is the point: the supervisor retries once the resync exits.
+    // The lock gets a token of its own so a lost lock stays distinguishable from an
+    // interrupt at the shutdown site. The two earn different treatment: only one of them
+    // means a resync may already be dropping the tables underneath us.
+    let live_lock_lost = CancellationToken::new();
+    let _live_lock = storage
+        .try_acquire_live_lock(
+            LiveLockMode::Shared,
+            INDEXER_LOCK_ROLE,
+            live_lock_lost.clone(),
```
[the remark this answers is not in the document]
Huzaifa696: The token is now observed across the whole startup sequence rather than only at supervision: schema init and reconciliation run under a guard, and the startup fill sits inside one future watched by a select that aborts both writers on loss. Backfill-only takes the token too and runs its fill, drain and checkpoint flush as one watched future, so it can no longer commit a frontier after the lock is gone.

indexer/src/storage/common/storage/live_lock.rs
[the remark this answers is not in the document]
Huzaifa696: Both roles now share one stop_signal helper that puts the lock-loss branch first under biased;, so a lost lock wins whenever both are ready in the same poll. The operator keeps its existing biased; property, since the combined signal branch still sits ahead of every task-completion branch. One correction: the indexer shutdown_signal had no biased; at all, so there the hazard was random branch choice rather than a fixed wrong order.

indexer/src/storage/common/storage/live_lock.rs
```diff
@@ -0,0 +1,410 @@
...
+        match self {
+            Self::Postgres(handle) => handle.run_fenced(f).await,
+            #[cfg(any(test, feature = "test-mock-storage"))]
+            Self::Noop => Err(StorageError::LiveStateLockLost),
+        }
+    }
+}
+
+impl LiveLockHandle {
+    async fn ensure_held(&self) -> Result<(), StorageError> {
```
sebatustra, line 243: A false-positive lock loss makes resync drop every table and then refuse to rebuild, leaving an empty database

Resync's pre-drop guard asks the server whether the lock is still held, but it never asks itself whether its own heartbeat has already given up on the lock:

```
  // resync.rs:340 — the only thing between the pre-flight and the drop
  live_lock.ensure_held().await.inspect_err(|e| {
      error!("Refusing to drop tables: {}", e);
  })?;
```
```
  // resync.rs:347
  info!("Dropping existing database tables...");
  self.storage.drop_tables_fenced(&live_lock).await
```

ensure_held is a fresh pg_locks probe and nothing more (live_lock.rs:220-234). It does not consult lock_lost. And the heartbeat, when it declares a loss, deliberately leaves the session open and healthy:

```
  // live_lock.rs:318-322
  // A loss verdict can be a false positive on a healthy session, and closing
  // now would free the lock while the caller is still winding down. Keep the
  // session until the guard is dropped, then let it go.
  task_stop.cancelled().await;
  task_session.close().await;
```

So after a false loss verdict the session still holds the lock, ensure_held() returns Ok, and the drop goes through. The very next thing that runs is:

```
  // resync.rs:479-492
  tokio::select! {
      biased;
      _ = lock_lost.cancelled() => {
          error!("Live-state lock lost during the rebuild; stopping it");
          processor_abort.abort();
          checkpoint_abort.abort();
          return Err(IndexerError::Storage(StorageError::LiveStateLockLost));
      }
      result = rebuild => result?,
  }
```

indexer/src/storage/common/storage/live_lock.rs
```diff
@@ -0,0 +1,410 @@
...
+            probe_advisory_lock_held(conn, self.session.key),
+        )
+        .await
+        {
+            Ok(Ok(true)) => Ok(()),
+            _ => Err(StorageError::LiveStateLockLost),
+        }
+    }
+
+    async fn run_fenced<T, F>(&self, f: F) -> Result<T, StorageError>
```
sebatustra, line 260: The fenced drop has no timeout, and a blocked DROP TABLE reads as Busy forever, wedging resync while it holds the exclusive lock

indexer/src/operator/operator.rs
sebatustra, line 1: The processor-death and critical-task-exit paths still drain and flush after the lock is lost

Every branch this PR added stops writers outright on a lost lock, with the same reasoning each time — "Closing the checkpoint channel is the writer's cue to flush what it has, which would commit a durable frontier over a database this run only half rebuilt". Two paths were missed.

Indexer. The token is moved into stop_signal, so the ProcessorEnded arm structurally cannot check it:

Operator. The token is cloned here, but the five critical_exit branches never look at it, and all of them fall through to a full graceful drain:

core/src/accounts/truncate.rs
```diff
@@ -81,9 +81,18 @@ pub async fn truncate_slots(
 
     let pool = db.pool.clone();
 
+    // The lock belongs to one session, so acquire and release have to run on the
+    // same connection. Taken from the pool, a statement lands on whichever
+    // connection is free, and unlocking from a different one is a silent no-op
+    // that strands the lock on the acquiring connection until it is recycled.
+    let mut lock_conn = pool
```
sebatustra, line 89: truncate_slots can return a lock-holding connection to the pool, and discards the report when it does

The connection-affinity fix is correct and the tightened test is a genuine improvement. The release path is not:

```
  // truncate.rs:108-117
  let released = sqlx::query_scalar::<_, bool>("SELECT pg_advisory_unlock($1)")
      .bind(TRUNCATE_ADVISORY_LOCK_ID)
      .fetch_one(&mut *lock_conn)
      .await
      .context("Failed to release advisory lock")?;     // ← early return
  if !released {
      return Err(anyhow!(
          "Truncation advisory lock was not held by the releasing session"
      ));
  }

  result
```

If that query errors rather than returning false, ? returns early and lock_conn — a PoolConnection — drops straight back into the pool still holding CONTRUC.

indexer/src/storage/common/storage/live_lock.rs
Huzaifa696: Good catch, and the sequencing is exactly as described. The pre-drop guard now fails if either the server says the lock is gone or this process own heartbeat has already declared a loss, so a false verdict refuses the drop instead of destroying the tables and then refusing to rebuild. Parking still holds the lock through the wind-down; it just no longer permits new destructive work.

indexer/src/storage/common/storage/live_lock.rs
Huzaifa696: Correct, and the Busy arm made it invisible. Fenced work is now bounded by a client-side timeout and the lock session carries a Postgres lock_timeout, so a DROP queued behind a table lock fails cleanly rather than wedging resync while it holds the exclusive lock.

indexer/src/operator/operator.rs
Huzaifa696: Both paths confirmed. The indexer now clones the token before it is moved into stop_signal so the ProcessorEnded arm can check it, and the operator checks it after the critical_exit branches. On a lost lock both stop outright instead of draining and flushing.

core/src/accounts/truncate.rs
Huzaifa696: Agreed, that was a real hole in the fix. Both the failed-release and not-held paths now detach the connection instead of letting it return to the pool, so a session that cannot be proven unlocked has its socket closed and the lock dies with it. The truncation error is no longer silently dropped either.

indexer/src/indexer/resync.rs
```diff
@@ -350,50 +416,81 @@ impl ResyncService {
...
+            biased;
+            _ = lock_lost.cancelled() => {
+                error!("Live-state lock lost during the rebuild; stopping it");
+                // Both writers are stopped outright rather than drained. Closing the
+                // checkpoint channel is the writer's cue to flush what it has, which
+                // would commit a durable frontier over a database this run only half
+                // rebuilt and leave no gap for a later run to detect.
+                processor_abort.abort();
+                checkpoint_abort.abort();
+                return Err(IndexerError::Storage(StorageError::LiveStateLockLost));
```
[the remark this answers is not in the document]
Huzaifa696: Confirmed, abort only requests cancellation and an in-flight statement can still land after it returns. Every lock-loss path now aborts and then waits for the writers to actually stop before returning, through one shared helper, bounded by a timeout that logs if a writer outlives it.

sebatustra (approved):

PR #296, opened 2026-09-07
fix(core): fail closed when the dedup restore window has a gap (issue-148)

Huzaifa696:
**Problem.** On restart, the dedup cache and live blockhash set were rebuilt from the newest blocks in Postgres without verifying they formed an unbroken chain tail. A missing block could leave its transactions out of the dedup cache while their blockhashes remained valid, allowing a transaction to execute twice. Slot adjacency is not sufficient because idle slots can advance without storing blocks.

**Fix.** `get_last_blocks` now validates the window using block heights, where each stored block increments the height exactly once. The newest height must match the durable `block_height` counter, which also catches blocks missing from the top. Both reads use the same `REPEATABLE READ` snapshot to avoid inconsistent results during a commit. Gaps at the oldest end are safe because their blockhashes are already expired; any other gap is treated as a fatal startup error.
### Coverage Report

| Core | 22,721 | 24,614 | 92.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34592799085) |
| Indexer | 39,969 | 44,301 | 90.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34592799085) |
| Gateway | 1,662 | 1,957 | 84.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34592799085) |
| Auth | 1,762 | 1,982 | 88.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34592799085) |
| E2E Integration | 16,770 | 22,354 | 75.0% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34592799085) |
| **Total** | **82,884** | **95,208** | **87.1%** | |

> Last updated: 2026-09-11 12:09:33 UTC by E2E Integration


core/src/accounts/get_last_blocks.rs
```diff
@@ -23,12 +27,32 @@ pub async fn get_last_blocks(db: &AccountsDB, limit: usize) -> Result<Vec<BlockI
...
     let rows = sqlx::query("SELECT data FROM blocks ORDER BY slot DESC LIMIT $1")
         .bind(limit as i64)
-        .fetch_all(pool.as_ref())
+        .fetch_all(&mut *tx)
         .await
         .context("Failed to query the most recent blocks")?;
 
+    // The raw counter, not the height reader: that one falls back to the highest
+    // slot when the key is gone, which would compare a slot against a height.
+    let durable_height = super::get_block_height::read_block_height_counter(&mut *tx).await?;
```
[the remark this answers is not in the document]
Huzaifa696: Acknowledged, and accurate about the behaviour. This is a deliberate design decision: the deployment is greenfield, so there is no pre-counter ledger to upgrade in place, and a missing counter alongside block rows is treated as corruption rather than an upgrade state. If in-place upgrade ever becomes real, this is the first assumption to revisit.

core/src/accounts/truncate.rs
```diff
@@ -81,9 +81,18 @@ pub async fn truncate_slots(
...
     let pool = db.pool.clone();
 
+    // The lock belongs to one session, so acquire and release have to run on the
+    // same connection. Taken from the pool, a statement lands on whichever
+    // connection is free, and unlocking from a different one is a silent no-op
+    // that strands the lock on the acquiring connection until it is recycled.
+    let mut lock_conn = pool
+        .acquire()
+        .await
+        .context("Failed to reserve a connection for the truncation lock")?;
```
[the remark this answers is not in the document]
Huzaifa696: Valid, and reproduced: at max_connections=1 the truncation timed out after 32s, while the pre-fix code passed in 4s. Fixed in 4c39f307 by opening the lock session with PgConnection::connect_with outside the pool, so it consumes no pooled connection at any pool size. This also matches writer_lease.rs, which avoids a pooled lock because sqlx does not reset a returned connection. Covered by truncate_runs_on_a_single_connection_pool.

core/src/accounts/get_last_blocks.rs
```diff
@@ -39,5 +63,168 @@ async fn get_last_blocks_postgres(db: &PostgresAccountsDB, limit: usize) -> Resu
...
+
+    let (newest_slot, newest_height) = previous.expect("the loop ran over a non-empty slice");
+    match durable_height {
+        Some(durable) => ensure!(
+            newest_height == durable,
+            "the newest restored block at slot {newest_slot} has height {newest_height} but the \
+             durable block height counter is {durable}, so the dedup window is not the chain tip; \
+             restore from backup",
+        ),
+        None => bail!(
```
sebatustra, line 122: A missing block_height counter is a permanent boot failure, and the fallback written for exactly that case is now dead code. 

Two paths in this repo now disagree about the same situation. get_block_height_postgres documents and handles it:

```
  // core/src/accounts/get_block_height.rs:41-50
  match read_block_height_counter(pool.as_ref()).await? {
      Some(height) => Ok(Some(height)),
      // A node upgrading in place has no counter yet. The last stored block
      // carries the height the old build assigned it, so continuing from there
      // keeps every lastValidBlockHeight a client holds valid.
      None => last_block_height_postgres(db).await,
  }
```

But load_dedup_state runs at nodes/node.rs:347, long before settle ever reads a tip, so on a Write/Aio node the bail at get_last_blocks.rs:120 fires first and last_block_height_postgres is never reached. The node also can't heal itself: the counter is only written as a side effect of committing a block (write_batch.rs:273), and committing a block requires starting.

The ledger shape that hits this is concrete, not hypothetical. block_height (the metadata key) landed in de449f9 (#290); the last BlockInfo layout change was 5c81be0 (#275), which added transaction_message_hashes. A ledger written by any build between those two decodes fine, has contiguous heights (pre-#290 settle.rs set block_height: Some(next_slot) and produced a block every slot), and has no counter row — so it passes the gap check and dies on the counter check. And last_block_height_postgres's MAX(slot) would return exactly the right answer for that ledger, because height was slot back then.

The suggested recovery is also wrong for this case: "restore from backup" restores a backup with the same missing key.

core/src/accounts/get_last_blocks.rs
```diff
@@ -39,5 +63,168 @@ async fn get_last_blocks_postgres(db: &PostgresAccountsDB, limit: usize) -> Resu
...
+        if let Some(durable) = durable_height {
+            bail!(
+                "the ledger has no block rows but the durable block height counter is {durable}, \
+                 so every block in the dedup window is missing; restore from backup"
+            );
+        }
+        return Ok(());
+    }
+
+    // Only the oldest blocks may be absent: their transactions can only name a
```
sebatustra: An interior gap bricks the node, though the PR's own safety argument gives a provably safe automatic recovery.

That comment states the general rule, then the code applies it only to the oldest end (where it happens for free, because LIMIT slides the window down). The identical reasoning covers any gap:

  - The missing block at height h held transactions whose recent_blockhash came from a block at height ≤ h-1.
  - If you keep only heights h+1..=T as the restored window, every hash at or below h is not in live_blockhashes, so dedup.rs:281-287 drops those transactions as unknown before dedup is even consulted.
  - Blocks above h are all present, so every transaction that named a still-live hash is still in the cache.

So "truncate the live window at the first gap" is fail-closed, not fail-open. What genuinely must stay fatal is the top gap — newest_height != durable at :114-119 — because those blocks' transactions used hashes that remain live.

Huzaifa696: Agreed, and fixed. The window now drops every block at or below the gap and keeps the contiguous suffix, which takes those blockhashes out of the live set so their transactions are refused as unknown rather than replayed. A warning names the heights either side, since it is still data loss worth investigating. The top gap stays fatal for exactly the reason you gave.

core/src/accounts/get_last_blocks.rs
Huzaifa696: Right that the advice was wrong, and that is fixed: the message now says a backup carries the same missing key and points at restoring the counter from the newest stored block's height. Keeping the case fatal though, since greenfield rules out the pre-#290 ledger while a counter deleted from a live one is real corruption this must catch, and MAX(slot) would wave it through on a genesis-only ledger. The asymmetry with the height reader is now documented on both sides.

sebatustra (dismissed):

sebatustra (approved):

PR #299, opened 2026-09-08
Merge `origin/main` into the OtterSec audit branch

Huzaifa696:
## What this does, and why

`ottersec-audit` carries the OtterSec remediation, including the withdrawal redesign that replaced the sparse Merkle tree with an on-chain nonce bitmap. `origin/main` carries the Cantina remediation (59 commits squash-merged as `5c81be03`) plus 17 later fixes, all written against the SMT design. The two lineages diverged four months ago at `977327b9` and neither is an ancestor of the other, so `ottersec-audit` cannot reach `main` without absorbing it first.

The hazard is not textual. `git merge origin/main` produces 51 conflicted files, but the dangerous part is what it does *not* flag: because ours deleted whole SMT modules that theirs never touched, git resolved large regions of theirs' SMT code in as clean additions e.g 185 stale symbols in `sender/remint.rs` alone. 

So this branch is not one commit. It is a merge commit that resolves the tree deliberately, followed by eleven ports that re-express, on the bitmap design, every Cantina fix that still applies but whose implementation was SMT-shaped.

## Design decisions

- **One merge commit, the ports as follow-ups.** The merge only resolves conflicts and deletes the SMT-only code. Nineteen Cantina fixes still apply to us, but they were written for the SMT, so each is rewritten for the bitmap as its own commit. Put them inside a 51-file merge and nobody can review them: `git show` does not display changes that came from neither branch.
- **In the operator tree, keep our files and add main's fixes back by hand.** For `operator/{operator,processor,recovery}.rs`, `operator/sender/*.rs` and `operator/utils/{instruction_util,mod}.rs` we keep our version of the whole file. Git had already mixed 185 lines of dead SMT code into these files without marking them as conflicts, so our committed version is the cleaner starting point.
- **In `core`, take main's version where it already covers more.** OtterSec issue-11 and Cantina issue-150 are the same bug fixed twice, and main's fix goes further: it makes `AccountsDB::get_accounts` return a `Result` and carries that through settle and the streamer. Do not combine both sides of `stage_metrics.rs`; each defines the same four names, so the crate would not compile.
- **Delete the old SMT code rather than leave it sitting there.** Out go five storage methods, the `indexer_state.owed_rotation_target` column, four error variants, three metric labels and the `tree-rotation-alerts` group. Dead code that still compiles is exactly how the frontier and `trip_halt` bugs above slipped in: nothing forces anyone to look at it.
- **Do not take the withdrawal dequeue nonce frontier.** `get_and_lock_pending_transactions_internal` keeps our version and `SKIP LOCKED`. Cantina #26 refuses to hand out a withdrawal while any lower nonce is unfinished. The SMT needed that, because releasing out of order broke the proof for the lower leaf. The bitmap does not: `WithdrawalBitmap::consume_nonce` flips one bit by position and never looks at the bits below it. Keeping the rule would only cost us liveness, since a single `manual_review` row would block every higher nonce. #26 is a HIGH finding and this drops its fix, so it needs written sign-off. The two audits disagree here; see the Cantina #26 section.
- **Settle transaction outcomes on `finalized`, never `confirmed`.** Both confirmation-poll paths and `attempt_remint` now require `finalized()` before they write a row terminal. Main had already moved to this and the merge lost it. It matters here because the read side (`classify_endpoint`) always required `finalized`, so a `confirmed` write meant the two halves disagreed about what counts as settled, and a forked-out block could be recorded as paid.
- **Keep both reconciliation checks; they do not clash.** Cantina #196's supply check runs in `reconcile_against_snapshot`, ours in `classify_and_report`. They only look contradictory because each measures the gap the other way round: main's "supply exceeds custody" is the same thing we call a shortfall, not the surplus we stopped blocking on.
- **Accept that `admin-ui` is deleted.** Main removed the app as dead code. Our only change to it was a manual bitmap-rotation button, and the automatic driver (`d91ed1e0`) is the supported way to do that now.
- **Use program id `9tgHa1DcnaSSUtmMsst8ovKTe1Gfxzezn27KnH9xXYeU`.** Main already switched to it everywhere: env files, docs and Makefiles. Going back to our `GokvZqD2…` would mean editing all of them again. See the deployment note below.
- **Keep our storage function signatures where only the argument order differs.** `get_stale_{parked,processing}_transactions` stay as `(threshold, limit, transaction_type)`. `set_pending_remint` keeps our signature but main's SQL body, which is safe to run twice, and that is what lets the #229 handoff retry without checking first.
- **`write_slot` takes the observed releases, and does so inside the retry.** We keep main's `finalize_and_checkpoint`: it returns a `Result` and retries the whole write, killing the process once the retries run out. Our observed-release write becomes an argument to `write_slot` instead of a step after the loop. Moving it back out compiles and looks correct, and quietly brings back the exact bug OtterSec's bitmap review raised against our code.

## Audit coverage

**All 28 OtterSec findings are preserved.** Thirteen have fixes that exist only on ours and survive because those files were resolved to our version: role-scoped recovery (#1), the `InitializeMint2` gate (#4), directional reconciliation (#5), the account-keys mirror (#9), remint to the initiator (#13), duplicate account keys (#14), the bounded halt sweep (#15), the stalled-withdrawal clear (#20), the rotation decoupling (bitmap review #2), the bitmap itself (on-chain suggestion #1), the gateway role cache (Web2 #1) and the signing-challenge binding (Web2 #2 and #5). The rest are main-side or predate the divergence. Two are satisfied differently and are called out on purpose:

- **Off-chain #12** ("distinguish missing accounts from load errors"). This branch drops ours' implementation in favour of main's Cantina #150, which we checked does strictly more. The finding's intent survives; the code implementing it does not, so an auditor diffing #12 against this branch should not read that as a regression.
- **Bitmap review #1** is a finding against ours' own refund gate, and its patch is main's, not ours: it is closed by adopting main's retried slot-write path, described in the `write_slot` decision above. It is the one finding this merge actually *fixes* rather than just carries over: a Cantina fix is what closes an OtterSec one.

**Nineteen Cantina findings are re-expressed on the bitmap design** as these eleven commits, in dependency order:

| # | Commit | Findings |
|---|---|---|
| 1 | `8abfca48` | `#210` restrict the deferred remint queue to the Withdraw role |
| 2 | `d09f555a` | `#198`, `#207`, `#209` pre-broadcast requeue, head rescue, unsupported-mint park (with `41ffcb01`) |
| 3 | `05160554` | `#229`, `#240` (Cantina `#274`, `#249`) retry the PendingRemint handoff, classify the journal not the stash |
| 4 | `26fa4d7b` | finding-7, `#230`, `#206` remint journal, one-live-attempt index, deferred outcomes |
| 5 | `d9e98725` | `#197`, `#201`, `#186` ownership-claimed write-ahead persist and JIT mint re-fires |
| 6 | `438acbf7` | `#187`, `#192`, `#243` coverage-proved absence, corroborated Dead on a fallback RPC |
| 7 | `a845db0d` | `#242`, `#223` slot-bound bitmap proof for recovery re-arming |
| 8 | `ddc9ac15` | `#200`, `#205` finality-gated reopened deposit re-mints, never terminalize a persisted mint |
| 9 | `07e78232` | `#280` length-check `getSignatureStatuses` on both poll paths |
| 10 | `a94d2b80` | `#231`, `#285` deposit `SendDurability`; stop terminalizing funded deposits |
| 11 | `3a69755c` | `#1` key mint and remint idempotency memos on the durable source event id |

Three further commits are housekeeping: `29af011c` repairs test failures from the merge; `6231016` restores transaction-error redaction on ungated error-bearing methods; and `786e527` replaces main's leftover `withdrawal_dequeue_respects_nonce_frontier` test, which still asserted the Cantina #26 behaviour this branch rejects, with one that pins the chosen behaviour instead.

A high-effort review of the whole branch afterwards found eleven defects of its own, closed by these eight commits. Three of them (`ee2de92`, `fd3604f`, and half of `f730740`) are Cantina fixes the merge itself had dropped, which is the failure mode this branch's structure exists to catch:

| Commit | Fixes |
|---|---|
| `f730740` | `MockStorage` still implemented the nonce frontier this branch removes, and wrote the refund-authorising flag to the wrong copy of the row |
| `ee2de92` | both confirmation-poll paths and `attempt_remint` settled on `confirmed()` where main used `finalized()`, so a forked-out block could be recorded as paid |
| `f34b772` | a dead-signature withdrawal with no configured instance was re-armed with no on-chain check, allowing a second release for a nonce already spent |
| `c324bd5` | boot promotion could complete a row whose refund a previous process had already claimed and broadcast, then delete the claim record |
| `e3529d8` | the halt sweep quarantined rows without their release journal, leaving them invisible to the sweep that repairs them |
| `fd3604f` | a dropped in-flight permit starved the JIT mint-init retry under sustained load |
| `e1a3e9d` | redaction was decided from the raw JWT, so a demoted operator kept raw errors until the token expired |
| `12abf5c` | `parse_create_instance` was never re-indexed for the eight-account layout, so every account after `instance` decoded as the wrong key |

`66abf09` records the round: 1895 tests pass, every resolution check listed above still holds, and each fix is named to the test that pins it.

**One known gap.** The two SQL changes in `c324bd5` and `e3529d8` are exercised only through `MockStorage`, not against real Postgres. Two of the defects above were in that same mock, so this deserves a Postgres-level test before the branch lands.

## Deployment constraint, read this before deploying

`9tgHa1DcnaSSUtmMsst8ovKTe1Gfxzezn27KnH9xXYeU` currently hosts the **SMT-era** program on devnet. The merged tree's program is the **bitmap** program. The merged indexer must not run against devnet until the bitmap program is upgraded at that address, or every release fails on an unknown instruction discriminator. This is a deploy-ordering constraint, not a code change, and it is the biggest operational risk in this merge.

Two related deployment notes:

- Cantina #196 turns the channel RPC into a hard boot gate for the escrow indexer: `reconcile_against_snapshot` returns `MissingChannelRpc` when `source_rpc_url` is absent or blank. An escrow indexer deployed without it stops booting after this merge where it previously started. `private-channel-deploy/vars/dev.yml` sets it; production is unverified from this tree.
- The schema union is additive only, so a rolled-back binary still runs against a migrated database. Rollback for the branch itself is `git reset --hard d91ed1e0` after tagging `backup/pre-main-merge-<date>`.

### Coverage Report

| Core | 21,713 | 23,554 | 92.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34335576897) |
| Indexer | 38,417 | 42,584 | 90.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34335576897) |
| Gateway | 1,661 | 1,956 | 84.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34335576897) |
| Auth | 1,747 | 1,967 | 88.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34335576897) |
| E2E Integration | 16,292 | 21,656 | 75.2% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34335576897) |
| **Total** | **79,830** | **91,717** | **87.0%** | |

> Last updated: 2026-09-09 10:21:42 UTC by E2E Integration


gateway/src/lib.rs
```diff
@@ -366,22 +734,36 @@ impl Gateway {
...
-        // Public methods need neither a token nor a role check.
+        // Public methods need neither a token nor a role check, but redaction is a
+        // separate axis: getSignatureStatuses is ungated and still error-bearing.
         if !is_gated(method) {
-            return None;
+            return Ok(redacts_transaction_errors(
+                auth_header,
+                decoding_key,
+                method,
+            ));
```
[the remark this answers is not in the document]
Huzaifa696: correct, and not covered by e1a3e9d, which only fixed the gated path. Fixed in 47aefbd: the DB role re-check now runs before the ungated branch returns, so both paths decide redaction from the confirmed role, and an unreachable auth DB redacts rather than taking the public method down. Pinned by test_demoted_operator_loses_raw_errors_on_ungated_status.

indexer/src/operator/sender/transaction.rs
```diff
@@ -1622,7 +2057,11 @@ pub(super) async fn route_poll_results(
 ) {
     for (mut tx, status_opt) in results {
         match status_opt {
-            Some(status) if status.satisfies_commitment(CommitmentConfig::confirmed()) => {
+            Some(status) if status.satisfies_commitment(CommitmentConfig::finalized()) => {
```
sebatustra, line 2060: since both poll gates are now moved to `finalized()`, we need to increase `MAX_POLL_ATTEMPTS_CONFIRMATION` and `DEFAULT_CONFIRMATION_POLL_INTERVAL_MS`, the current values give a poll budget of ~2s  vs the ~12s needed, thus making all deposits fail

indexer/src/operator/sender/transaction.rs
sebatustra, line 667: this is `confirmed()`, but PR description claims that "Both confirmation-poll paths and attempt_remint now require finalized() before they write a row terminal."

monitoring/provisioning/alerting/alert-rules.yml
sebatustra, line 1: do we not have alerts for `rotation_blocked_by_lower_nonce` and `rotation_lost`? these are outages that should have a page

indexer/src/operator/sender/mint.rs
```diff
@@ -414,6 +412,94 @@ async fn mint_authority_check_with_backoff(
...
+                    (ConsumedMintKind::Remint, rest)
+                } else {
+                    continue;
+                };
+
+            let Some(source_event_id) = SourceEventId::from_encoded(encoded) else {
+                return Err(format!(
+                    "channel mint {} carries an idempotency memo that does not parse to a \
+                     current-scheme source-event-id (legacy memo scheme); resync cannot reconcile \
+                     across the memo cutover - drain the operator and follow the cutover runbook",
```
sebatustra: what would be this runbook?

MUST_FIX_ROUND2.md
sebatustra, line 1: was this file commited by accident?

indexer/src/storage/postgres/db.rs
sebatustra, line 863: does this "reassigns nonces from zero" means that on a resync we leave the DB issuing nonces from 0 against a bitmap on generation G with bits already set?

indexer/src/operator/db_transaction_writer.rs
```diff
@@ -73,6 +73,9 @@ impl DbTransactionWriter {
                 update.status,
                 update.counterpart_signature.clone(),
                 update.processed_at.unwrap_or_else(Utc::now),
+                // Theirs' release_signatures provenance column is out of scope
+                // for this merge, so nothing is written to it yet.
+                None,
```
sebatustra, line 79: did we drop the cantina fix here?

indexer/src/operator/db_transaction_writer.rs
sebatustra: this comment appears to be stale, there is no "local tree" anymore

Huzaifa696:
On the poll budget: the numbers are right but the chain is not, so I do not think a change is needed. Both sites that gate on `finalized()` poll the **channel**, never Solana. Mints and `InitializeMint` are fire-and-forget and polled on `state.rpc_client`, which `SenderState::dest_finality` tags `FinalityRpc::channel` for the Escrow role (the role that mints); `attempt_remint` sends and confirms on `state.source_rpc_client`, which `source_finality` tags `channel` for the Withdraw role. The channel has no 32-slot finalization: `get_signature_statuses_impl` returns `Finalized` for every stored transaction, so the gate is met on the first poll and the ~2s budget is ample. Solana finalization would apply only to the synchronous `send_and_confirm` path, and that stays on `confirmed()` here and on main. A full local `resync_integration` run (deposit/mint heavy) shows no confirmation timeouts.

Your other three are all valid and are fixed in `a8510bc`:
- line 667: correct, the PR description overclaimed. It now says only the two async sites plus `attempt_remint` moved, and that the sync release path stays `confirmed()`.
- rotation alerts: added `rotation-lost` (critical, fires immediately, nothing re-dispatches an abandoned rotation) and `rotation-blocked-by-lower-nonce` (warning held 15m, since withholding is correct and only matters when it persists).
- the runbook: `docs/runbooks/mint_memo_cutover.md` now exists and the error names it. Flagged as untested guidance, so worth a read from someone who has run a drain.

indexer/src/operator/sender/transaction.rs
Huzaifa696: Both finalized() gates poll the channel, not Solana: dest_finality tags rpc_client channel for the Escrow role (which mints), source_finality does the same for attempt_remint on Withdraw. The channel returns Finalized for every stored transaction, so the gate is met on the first poll and ~2s is ample.

indexer/src/operator/sender/transaction.rs
Huzaifa696: Correct, the description overclaimed. Fixed: it now says only the two async sites plus attempt_remint moved, and that this synchronous release path stays confirmed() as it is on main.

monitoring/provisioning/alerting/alert-rules.yml
Huzaifa696: Agreed, both were unalerted. Added in 9cbaba9: rotation-lost is critical and fires immediately since nothing re-dispatches an abandoned rotation; rotation-blocked-by-lower-nonce is a warning held 15m because withholding is correct and only matters when it persists.

indexer/src/operator/sender/mint.rs
Huzaifa696: It did not exist. Written in a8510bc as docs/runbooks/mint_memo_cutover.md and the error now names it. Flagged as untested guidance, so worth a read from someone who has run a drain.

MUST_FIX_ROUND2.md
Huzaifa696: Yes, an automated step committed it before the other review notes were untracked. Untracked in 9e89a16, file kept on disk. No history rewrite since the branch is already under review.

indexer/src/storage/postgres/db.rs
Huzaifa696: Good catch, and yes: drop_tables drops withdrawal_nonce_seq and init_schema recreates it START 0, while the bitmap keeps its generation and bits. 04f299d adds a pre-flight that refuses a withdraw resync unless the bitmap is fresh, failing closed if it cannot be read. Renumbering is not offered because it rewrites which withdrawal a nonce names.

indexer/src/operator/db_transaction_writer.rs
Huzaifa696: Deferred rather than dropped: the column is migrated but nothing on this branch reads it, and adopting it adds a field to every TransactionStatusUpdate. The comment said only out of scope, which read like an oversight; c19dc65 states the reasoning.

indexer/src/operator/db_transaction_writer.rs
Huzaifa696: Correct, stale. Reworded in c19dc65 to name validate_bitmap_consistency, which is what actually refuses the next boot once Completed rows and bitmap bits diverge.

sebatustra (approved):

PR #206, opened 2026-07-20
fix(indexer): defer pre-broadcast remint failures instead of ManualReview

sebatustra:
Fix: defer pre-broadcast remint failures instead of ManualReview (finding-283)

`execute_deferred_remint` sent every attempt_remint error to ManualReview, so a transient source-chain RPC/signer blip on a fresh compensation flipped the row off PendingRemint and hid it from recovery. Fix: reclassify pre-broadcast / nothing-sent failures from Failed → Defer so they flow through the existing bounded-retry path.

Recommendations:
  - Typed before/after-broadcast outcome — already modeled by RemintAttempt (Confirmed/Defer/Failed); corrected the routing.
  - Keep build/blockhash/signing failures in PendingRemint, bounded counter + requeue — done; these now Defer into defer_or_escalate (persists counter + deadline, requeues, row stays PendingRemint).
  - Defer + retry when the earlier-remint lookup is unavailable — done; get_remint_signatures failure now Defer.
  - Persist signature before broadcast, reconcile before retry — already implemented; also why Uncertain/unparseable cases stay Failed (fail-closed).

Tests updated for defer/requeue;
### Coverage Report

| Component | Lines Hit | Lines Total | Coverage | Artifact |
|-----------|-----------|-------------|----------|----------|
| Core | 11,846 | 13,489 | 87.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29930273692) |
| Indexer | 29,463 | 32,733 | 90.0% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29930273692) |
| Gateway | 1,325 | 1,515 | 87.5% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29930273692) |
| Auth | 785 | 903 | 86.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29930273692) |
| Withdraw Program | - | - | - | - |
| Escrow Program | - | - | - | - |
| DvP Swap Program | - | - | - | - |
| E2E Integration | 13,100 | 16,565 | 79.1% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29930273692) |
| **Total** | **56,519** | **65,205** | **86.7%** | |

> Last updated: 2026-07-22 15:40:41 UTC by E2E Integration

PR #196, opened 2026-07-10
fix(indexer): give runtime reconciliation teeth and an independent supply check

Huzaifa696:
## Summary

Closes issue 14. The runtime reconciliation loop only logged or optionally sent a webhook when it detected a balance mismatch, allowing processing to continue even in the presence of a potential insolvency. It also compared escrow custody only against the database ledger, never validating the actual on-chain supply of minted channel tokens. As a result, operator-key over-issuance or minting bugs could go undetected by the indexer's primary detective control.

This change replaces the previous balance reconciliation with a single on-chain invariant: **channel supply must never exceed escrow custody**. The runtime loop now verifies this using on-chain state, while using the database only to enumerate mints and compute a conservative in-flight envelope. A confirmed `supply > custody` breach that exceeds the envelope for three consecutive finalized observations triggers a durable fail-closed halt, freezing processing across operators, quarantining active withdrawals, forcing the service unhealthy, and issuing a mandatory webhook notification.

The implementation adds a persistent reconciliation halt flag, startup validation of the same supply invariant, and comprehensive unit, integration, and end-to-end test coverage. Existing custody-versus-ledger reconciliation has been removed in favor of the stronger supply-based check, ensuring the indexer independently verifies solvency using on-chain data while remaining resilient to transient RPC inconsistencies.
### Coverage Report

| Component | Lines Hit | Lines Total | Coverage | Artifact |
|-----------|-----------|-------------|----------|----------|
| Core | 11,486 | 13,120 | 87.5% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29254131807) |
| Indexer | 27,414 | 30,691 | 89.3% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29254131807) |
| Gateway | 1,325 | 1,515 | 87.5% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29254131807) |
| Auth | 785 | 903 | 86.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29254131807) |
| Withdraw Program | - | - | - | - |
| Escrow Program | - | - | - | - |
| DvP Swap Program | - | - | - | - |
| E2E Integration | 12,874 | 16,176 | 79.6% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/29254131807) |
[...]

PR #301, opened 2026-09-09
Land the OtterSec remediation and the withdrawal nonce bitmap on main

Huzaifa696:
## What this lands

This merges `ottersec-audit` into `main`. It brings the OtterSec audit remediation, the redesign that replaces the withdrawal sparse Merkle tree with an on-chain nonce bitmap, and the Cantina fixes that had to be rewritten for that design. Forty-seven commits, 191 files.

`ottersec-audit` already absorbed `main` in #299, so this merge is clean: the reconciliation between the two audit lineages was done there, reviewed there, and is not repeated here. Read #299 for why each conflict was resolved the way it was. This description covers what `main` gains and what changes for anyone running it.

## What changes on-chain

The escrow program no longer keeps a sparse Merkle tree of released withdrawals. It keeps a bitmap: one bit per nonce in the current generation, in a `withdrawalBitmap` PDA. `ResetSmtRoot` is gone and `RotateBitmap` replaces it, advancing the generation and clearing the bits once every nonce in the closing generation has released. `ReleaseFunds` checks that the nonce belongs to the current generation and that its bit is clear, then sets it. Out-of-order release inside a generation is safe by construction, because each bit is independent.

The program id is `9tgHa1DcnaSSUtmMsst8ovKTe1Gfxzezn27KnH9xXYeU`, the same id `main` already uses. See the deployment section: that address currently hosts the tree-era program on devnet.

## What changes in behaviour

**Withdrawal releases.** The indexer reads the bitmap instead of maintaining a local tree, so there is no local state that can drift from the chain. The rotation driver derives what is owed from the unreleased nonce bounds on every tick, including after a restart, rather than from a durable target column. A rotation that fails to send is re-armed a bounded number of times, then abandoned with an alert.

**Settlement is written on `finalized`, on the channel side.** The two asynchronous poll paths and the remint path require a finalized status before writing a row terminal. All three poll the channel, which reports every stored transaction as finalized immediately, so this costs no latency. The synchronous release path stays on `confirmed()` exactly as it is on `main` today.

**The withdrawal dequeue no longer holds later nonces behind an unresolved lower one.** Cantina #26 added that rule for the tree, where releasing out of order invalidated the lower leaf's proof. The bitmap has no such dependency, and OtterSec #15 explicitly recommends letting later nonces proceed. This drops a HIGH finding's fix on purpose and needs written sign-off from the Cantina response owner; the two audits contradict each other here and both should be told.

**Startup reconciliation only blocks on a custody shortfall.** A surplus is attacker-inducible, since anyone can send tokens to the escrow, so it warns instead of refusing to start. Cantina #196's supply-versus-custody check still runs first and still halts; the two measure the gap in opposite directions and do not conflict.

**A resync refuses to run against a bitmap that has advanced.** A resync restarts the nonce sequence from zero. Against a bitmap at generation G with bits set, every new release would be rejected and boot would compare rebuilt rows against bits belonging to other withdrawals. The pre-flight now reads the bitmap and fails closed, before dropping anything, if it is not fresh or cannot be read. A withdraw indexer configured without an escrow instance id or an RPC that can read the bitmap will now refuse a resync it previously ran.

**Deposits are never terminalized by the sender on a send or build failure.** A write-ahead-persisted mint stays `processing` for recovery to reconcile against its journaled signature. Previously a build failure could mark a funded deposit failed.

## Audit coverage

**All 28 OtterSec findings are closed.** Thirteen are fixes that exist only on this branch: role-scoped recovery (#1), the `InitializeMint2` gate (#4), directional reconciliation (#5), the account-keys mirror (#9), remint to the initiator (#13), duplicate account keys (#14), the bounded halt sweep (#15), the stalled-withdrawal clear (#20), the rotation decoupling (bitmap review #2), the bitmap itself (on-chain suggestion #1), the gateway role cache (Web2 #1) and the signing-challenge binding (Web2 #2 and #5). Two are worth calling out because they are satisfied differently from how they were filed:

- **Off-chain #12** is closed by `main`'s Cantina #150 rather than by this branch's original implementation, because #150 does strictly more. An auditor diffing #12 against this branch should not read that as a regression.
- **Bitmap review #1** is a finding against this branch's own refund gate, and it is closed by adopting `main`'s retried slot-write path. It is the one OtterSec finding that the merge with `main` fixed rather than merely preserved.

**Nineteen Cantina findings were rewritten for the bitmap.** Their original fixes were written against the tree and did not survive resolving the operator files to this branch's version, so each was re-expressed as its own commit with its own tests:

| Commit | Findings |
|---|---|
| `8abfca48` | #210 restrict the deferred remint queue to the Withdraw role |
| `d09f555a` | #198, #207, #209 pre-broadcast requeue, head rescue, unsupported-mint park |
| `05160554` | #229, #240, #274, #249 retry the PendingRemint handoff, classify the journal not the stash |
| `26fa4d7b` | finding-7, #230, #206 remint journal, one-live-attempt index, deferred outcomes |
| `d9e98725` | #197, #201, #186 ownership-claimed write-ahead persist and JIT mint re-fires |
| `438acbf7` | #187, #192, #243 coverage-proved absence, corroborated Dead on a fallback RPC |
| `a845db0d` | #242, #223 slot-bound bitmap proof for recovery re-arming |
| `ddc9ac15` | #200, #205 finality-gated reopened deposit re-mints, never terminalize a persisted mint |
| `07e78232` | #280 length-check `getSignatureStatuses` on both poll paths |
| `a94d2b80` | #231, #285 deposit `SendDurability`, stop terminalizing funded deposits |
| `3a69755c` | #1 key mint and remint idempotency memos on the durable source event id |

Every other Cantina finding arrived intact through the merge, with one deliberate exception, #26, described above.

**A full-branch review then found eleven further defects**, three of them Cantina fixes the merge had dropped, all closed:

| Commit | Fixes |
|---|---|
| `f7307401` | `MockStorage` still implemented the removed nonce frontier and wrote the refund-authorising flag to the wrong copy of the row |
| `ee2de921` | the two async poll sites and `attempt_remint` settled on `confirmed()` where `main` used `finalized()` |
| `f34b772e` | a dead-signature withdrawal with no configured instance was re-armed with no on-chain check |
| `c324bd53` | boot promotion could complete a row whose refund a previous process had already claimed and broadcast |
| `e3529d8a` | the halt sweep quarantined rows without their release journal, making them unrepairable |
| `fd3604fd` | a dropped in-flight permit starved the JIT mint-init retry under load |
| `e1a3e9d5` | redaction was decided from the raw JWT, so a demoted operator kept raw errors until the token expired |
| `12abf5c8` | `parse_create_instance` was never re-indexed for the eight-account layout |
| `47aefbd0` | the ungated `getSignatureStatuses` branch skipped the DB role check, leaking raw errors to a demoted operator |

## Deployment, read before merging

**The devnet program at `9tgHa1` is the tree-era program.** This tree's program is the bitmap program. The merged indexer must not run against devnet until the bitmap program is upgraded at that address, or every release fails on an unknown instruction discriminator. This is a deploy-ordering constraint, not a code change, and it is the largest operational risk in landing this.

**The escrow indexer now requires `source_rpc_url`.** Cantina #196 makes the channel RPC a hard boot gate: an escrow indexer deployed without it stops booting where it previously started. The dev deploy vars set it; production should be confirmed before rollout.

### Coverage Report

| Core | 21,830 | 23,680 | 92.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34376814182) |
| Indexer | 38,796 | 42,989 | 90.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34376814182) |
| Gateway | 1,662 | 1,957 | 84.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34376814182) |
| Auth | 1,762 | 1,982 | 88.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34376814182) |
| Withdraw Program | 129 | 241 | 53.5% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34376814192) |
| Escrow Program | 905 | 1,720 | 52.6% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34376814192) |
| E2E Integration | 16,399 | 21,776 | 75.3% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34376814182) |
| **Total** | **81,483** | **94,345** | **86.4%** | |

> Last updated: 2026-09-09 17:21:52 UTC by E2E Integration


indexer/src/indexer/resync.rs
```diff
@@ -104,6 +114,60 @@ impl ResyncService {
...
+            instance = %instance,
+            bitmap = %bitmap_pda,
+            "Checking the withdrawal bitmap has issued no nonces before any destruction..."
+        );
+        let rpc = RpcClientWithRetry::with_retry_config(
+            rpc_url.clone(),
+            RetryConfig::default(),
+            CommitmentConfig::confirmed(),
+        );
+        let bitmap = fetch_consumed_nonces(&rpc, &bitmap_pda, None)
```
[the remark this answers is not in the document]
Huzaifa696: Valid, fixed in 7d20a250: the bitmap is now read bound to the node's finalized tip via minContextSlot with a 120s tip-freshness check, and any completed withdrawal row refuses the resync locally before an RPC is consulted at all.

indexer/src/operator/sender/transaction.rs
```diff
@@ -129,113 +108,46 @@ impl SenderState {
...
-                        error!("Unreleased-nonce gate read failed: {e}");
-                        return Err(
-                            ProgramError::RotationGateUnavailable { target_tree_index }.into()
-                        );
+                        self.pending_rotation = Some(builder);
+                        return Err(e);
                     }
-                }
+                };
+                builder.expected_generation(expected_generation);
```
sebatustra, line 151: expected_generation exists so a replayed rotation is rejected instead of advancing twice. The re-arm path defeats it.

```
  let expected_generation = match self.refresh_generation().await { ... };
  builder.expected_generation(expected_generation);      // unconditional overwrite
  self.rotation_in_flight = Some(builder.clone());
```

rearm_failed_rotation puts that already-bound builder back into state.pending_rotation, and the next tick sends it straight back through the code above, which re-reads the chain and rebinds. Sequence:

  1. Rotation bound to G, broadcast, lands (chain now G+1).
  2. check_transaction_status returns Err(TransactionError::Rpc) (any RPC blip while polling, transaction_util.rs:134-140) or send_signed errors after the tx already propagated → handle_permanent_failure → rearm_failed_rotation.
  3. Re-dispatch reads G+1, binds G+1, rotates again → chain G+2.

  Generation G+1 is now closed forever

indexer/src/operator/sender/state.rs
```diff
@@ -95,265 +97,461 @@ impl SenderState {
...
+        "Validating withdrawal bitmap against completed withdrawals"
+    );
 
-    let instance_data = rpc_client
-        .get_account_data(&instance_pda)
-        .await
-        .map_err(|_| AccountError::AccountNotFound {
-            pubkey: instance_pda,
-        })?;
+    let bitmap = fetch_consumed_nonces(rpc_client, &bitmap_pda, None).await?;
```
sebatustra: The boot bitmap diff is the one bitmap read that isn't slot-bound, and it can halt startup

```
let bitmap = fetch_consumed_nonces(rpc_client, &bitmap_pda, None).await?;   // :212
```

Everywhere else the branch is careful about this. verify_release_landed (state.rs:136-169) anchors on a finalized blockhash's context slot and passes it as min_context_slot, with the rationale "a lagging backend errors instead of serving an older snapshot whose clear bit would read as proof of non-release". refuse_if_bitmap_advanced goes further and proves the tip is fresh against wall clock. The boot gate passes None, and so does confirm_divergence, so the confirmatory re-read can hit the same lagging backend.

A load-balanced RPC that routes both reads to a lagging node yields clear bits for nonces the DB has as Completed, db_only is non-empty, and the operator refuses to start

integration/tests/indexer/rotation_driver_e2e.rs
sebatustra, line 1: rotation_driver_e2e is declared but never run

The target exists and holds the three highest-value scenarios for this redesign (rotation_happens_with_no_boundary_row_present, a_manual_review_row_blocks_rotation_until_resolved, rotation_is_re_derived_after_a_restart_that_dropped_the_arm). Both Makefiles enumerate test targets one by one, and neither lists it:

It compiles under clippy, so nothing complains. Add it to integration/Makefile:integration-test, integration-coverage-indexer, and the root ci-integration-test-indexer group

indexer/src/operator/sender/transaction.rs
Huzaifa696: Confirmed and fixed: the binding is now taken once and survives the re-arm, so a rotation that landed but was reported failed comes back bound to its original generation and the program refuses it as UnexpectedGeneration instead of advancing a second time. It is released only when the rotation settles or is abandoned.

indexer/src/operator/sender/state.rs
Huzaifa696: Valid, both now bind to a finalized context slot. The confirmation re-read takes its own fresh anchor rather than reusing the first, since re-reading at the original slot would return the identical snapshot and defeat the point of the re-read.

integration/tests/indexer/rotation_driver_e2e.rs
Huzaifa696: Good catch, it had never run. Now wired into the prod-feature group of both integration Makefiles and the root CI indexer group, plus a fourth scenario that drives a landed-but-reported-failed rotation through the re-arm and pins the finding above.

sebatustra (approved):

PR #306, opened 2026-09-10
fix(core): enforce the account-data limit before the bytes are fetched (issue-276)

Huzaifa696:
The executor checks account sizes before loading their data. BOB serves accounts already in memory, while only missing accounts are checked in the database without loading the full data.

Transactions requiring over 64 MiB are rejected with the same error the SVM would return. Preloading is limited to 256 MiB, with each account counted once. Anything over the limit is deferred to a follow-up batch.

BOB is also capped by bytes, never evicting the current preload or unsettled writes, and results drop readonly account data so evicted accounts are not kept alive; account reads stream, so resident account data stays within 1 GiB plus one preload. 27 tests were added, including one real-node integration test.

### Coverage Report

| Core | 23,906 | 25,847 | 92.5% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34865310120) |
| Indexer | 39,979 | 44,301 | 90.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34865310120) |
| Gateway | 1,662 | 1,957 | 84.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34865310120) |
| Auth | 1,762 | 1,982 | 88.9% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34865310120) |
| E2E Integration | 16,998 | 22,664 | 75.0% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34865310120) |
| **Total** | **84,307** | **96,751** | **87.1%** | |

> Last updated: 2026-09-14 16:52:25 UTC by E2E Integration


core/src/stages/execution.rs
```diff
@@ -935,13 +1153,37 @@ pub async fn execute_batch(
...
+                error_metrics: TransactionErrorMetrics::default(),
+                execute_timings: ExecuteTimings::default(),
+                balance_collector: None,
+            });
+        for _ in 0..num_oversized {
+            results
+                .processing_results
+                .push(Err(TransactionError::MaxLoadedAccountsDataSizeExceeded));
+        }
+        regular_transactions.extend(oversized_transactions);
```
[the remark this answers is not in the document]
Huzaifa696: Correct that the order changes, but nothing depends on it: a `ConflictFreeBatch` has disjoint write sets so order carries no state semantics, and the reordered entries are failures that write nothing. The derived blockhash already folds in a nanosecond timestamp (`compute_blockhash`, settle.rs), so it is unreproducible by construction and no consumer can verify it against a transaction order.

core/src/stages/execution.rs
```diff
@@ -350,6 +382,121 @@ pub(crate) enum SendOutcome {
...
+/// batch actually ran.
+pub(crate) enum BatchOutcome {
+    Continue { executed: usize },
+    Stop,
+}
+
+/// Run one scheduled batch, in as many sub-batches as its preload budget needs.
+/// The loop lives here so each piece reaches the settler before the next is
+/// fetched, which keeps one preload's transient cost to the budget.
+pub(crate) async fn process_batch(
```
sebatustra, line 395: The budget bounds one fetch, not resident memory. a single malicious batch still materialises multiple GiB

plan_preload caps what one preload pulls at 256 MiB. But process_batch then runs the leftovers back-to-back in a loop, and BOB keeps everything it loaded:

```
  // execution.rs:483-501
  if deferred.is_empty() {
      return BatchOutcome::Continue { executed };
  }
  batch = ConflictFreeBatch { transactions: deferred.into_iter()... };
```

Nothing frees the previous sub-batch's accounts before the next one loads. BOB's only ceiling is an entry count, not bytes:

```
  // bob.rs:67-71
  /// Must stay well above the max distinct account keys a single batch can
  /// reference ... At ~1 KiB/account this bounds the clean set near ~1 GiB.
  const DEFAULT_MAX_CACHE_ENTRIES: usize = 1_000_000;
```

That "~1 KiB/account" assumption is exactly what this attack breaks. cap_evict (bob.rs:507) only fires above 1,000,000 entries; the age sweep (bob.rs:483) only evicts entries synced more than an hour ago, and every account this batch just loaded is stamped synced_since = now.

Concrete attack, post-fix. The original PoC (34 × 10 MiB per tx) is now rejected, good. But the attacker just shrinks each transaction to sit under the gate

core/src/accounts/get_accounts.rs
sebatustra, line 1: fetch_all is unchanged, so one "256 MiB" preload transiently costs ~2×

```
  let rows = with_retry(|| async {
      sqlx::query("SELECT pubkey, data FROM accounts WHERE pubkey = ANY($1)")
          .bind(&pubkey_bytes)
          .fetch_all(pool.as_ref())          // whole result set retained
          .await
  }).await...;

  for row in rows {                           // rows still alive for the whole loop
      let data: Vec<u8> = row.get("data");    // second copy of this blob
      match bincode::deserialize::<AccountSharedData>(&data) {
          Ok(account) => result[index] = Some(account),   // third copy accumulates
```
While the loop runs, memory holds: the complete Vec<PgRow> (the full budget), plus every account decoded so far (the full budget again), plus two copies of the row being decoded. A 256 MiB preload peaks near 520 MiB.

The finding asked for this specifically: "Replace fetch_all for caller-selected account blobs with bounded streaming or chunked reads. Track cumulative serialized and decoded bytes while reading..."


It also means the real per-preload ceiling is double what MAX_BATCH_PRELOAD_BYTES and docs/CORE.md advertise — worth either halving the constant or switching to fetch + a running byte counter.

core/src/stages/execution.rs
Huzaifa696: Agreed, fixed in 4f8a4e8e: BOB now also evicts by bytes (1 GiB cap, largest clean entries first, never the current preload's keys or unsettled writes), and results drop readonly account data so evicted accounts are not kept alive until settlement.

core/src/accounts/get_accounts.rs
Huzaifa696: Measured it at about 1.3x rather than 2x (240 MiB of rows peaked at 316 MiB, since each row is dropped as the loop consumes it), but switched to streaming with borrowed bytes anyway in 4f8a4e8e, which peaks at 258 MiB (1.07x).

sebatustra (approved):

PR #307, opened 2026-09-10
Fix/issue 106 107 cantina and transferhook support

sebatustra:
# Escrow: Cantina 106 and 107, transfer-hook support

## 107

`BlockMint` closed the `AllowedMint` PDA that `ReleaseFunds` also needs, so blocking a mint froze the balances already in it. `AllowedMint` now has two independent gates and is never closed. `BlockMint` sets both as absolute values, so `false` re-opens one. `AllowMint` doubles as the re-allow path.

The indexer mirrors the deposit gate in `mints.status`. The withdrawal gate is mirrored onto the `mints` row and read per withdrawal in the pre-flight, which parks the row instead of broadcasting a release the program rejects. No operator restart is needed for either.

## 106

`AllowedMint` pins the reviewed `decimals`, `token_program`, the set of Token-2022 extensions the mint carried, and whether it had a freeze authority. `Deposit` re-reads the mint and rejects with `MintProfileChanged` on a mismatch; an admin clears it by re-allowing, which re-pins the profile.

  Two asymmetries, both deliberate:

  - Freeze authority is compared one way. It can be revoked but never re-added, so losing one is the same mint behaving more safely, while gaining one means a recreate.
  - The metadata, group and group-member bits are recorded but not compared. An issuer writes them onto a live mint as a routine step, so comparing them would stop deposits on a mint that never changed. They carry no power over transfers, and a recreate hiding behind them still has to match decimals, token program, freeze authority and every other extension bit.

The mask pins *which* extensions exist, not their contents: a fee raised, a hook program swapped or a permanent delegate rotated leaves it identical. Freeze authority, pausable, permanent delegate and transfer fee stay accepted and are documented as issuer trust, with off-chain detection only — the operator reads the pinned profile per withdrawal, checks the pooled escrow ATA for a freeze, and checks live pause and drain state for the mints whose profile says they can.

`ReleaseFunds` does not re-check the profile. Closing a mint requires zero supply, so the escrow can only hold a balance while the mint is unclosable, and gating withdrawals on admin-reviewed state is the 107 bug.

## Transfer hooks

`Deposit` and `ReleaseFunds` forward everything past their fixed accounts to the token program, which resolves the mint's `ExtraAccountMetaList` and calls the hook. Ported from solana-foundation/dvp, including the guard: a forwarded account keeps its writable flag and never its signer bit. An `ExtraAccountMetaList` can name the fee payer, and without the strip the hook gets a live signature for that wallet. Two integration tests fail if the strip is removed.

The hook program is not pinned in `AllowedMint`. A hook receives read-only accounts, the strip covers payer and user, and `spl-tlv-account-resolution` de-escalates the transfer authority, so the worst a hook can do is revert a transfer. Pinning would add nothing against that and would strand withdrawals whenever a hook authority swaps to a new program.

Clients append the hook accounts themselves; codama drops trailing accounts, so the IDL cannot describe them. The withdrawal operator resolves them per withdrawal, because an `ExtraAccountMetaList` can key off the amount and destination, and parks the row when the validation account is missing or the resolved accounts exceed what a release transaction can carry.

Tested against a real hook program, in LiteSVM and on a test validator.
### Coverage Report

| Core | 24,416 | 25,765 | 94.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34980339548) |
| Indexer | 40,539 | 44,022 | 92.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34980339548) |
| Gateway | 1,653 | 1,918 | 86.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34980339548) |
| Auth | 1,758 | 1,962 | 89.6% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34980339548) |
| Withdraw Program | 129 | 241 | 53.5% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34980339463) |
| Escrow Program | 1,115 | 2,022 | 55.1% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34980339463) |
| E2E Integration | 17,107 | 22,346 | 76.6% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34980339548) |
| **Total** | **86,717** | **98,276** | **88.2%** | |

> Last updated: 2026-09-15 14:59:03 UTC by E2E Integration


private-channel-escrow-program/program/src/state/allowed_mint.rs
```diff
@@ -27,28 +37,76 @@ impl Discriminator for AllowedMint {
...
     pub const LEN: usize = 1 + // discriminator
-        1; // bump
-
-    pub fn new(bump: u8) -> Self {
-        Self { bump }
+        1 + // bump
+        1 + // deposits_blocked
+        1 + // withdrawals_blocked
+        1 + // decimals
+        32; // token_program
```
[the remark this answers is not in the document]
sebatustra: Correct for a live deployment, but not applicable here: the only deployment is devnet, and the migration path is to create a fresh instance rather than upgrade the old accounts. AllowedMint PDAs are seeded on the instance PDA, so a new instance gets a clean namespace and the pre-upgrade accounts are simply orphaned. Adding a realloc path to AllowMint would mean carrying permanent migration code for a one-time devnet reset, so we're taking the reset instead.

This does mean the layout change is not upgrade-safe, and it needs a real migration path if the program is ever live with allowlisted mints at upgrade time.

indexer/src/operator/processor.rs
```diff
@@ -611,14 +612,37 @@ async fn check_withdrawal_mint_supported(
...
+        )));
+    };
+
+    // release_funds rejects a withdrawal-blocked mint on-chain, and that failure is
+    // permanent, so letting it broadcast would terminalize the row through the
+    // remint path. Parking keeps it re-armable once the admin re-opens the gate.
+    //
+    // Checked before the existence floor below so a blocked mint is never cached as
+    // clear, which is what lets a re-armed row proceed without an operator restart.
+    if allowed_mint.withdrawals_blocked {
```
Huzaifa696, line 648: This check only runs on the first withdrawal of a mint per process because of the `has_existence_floor` early return at the top of the function. If an admin blocks withdrawals while the operator is up (which is when they'd actually do it), releases for that mint keep going out, the program rejects them and the sender remints the row instead of parking it. The runbook says to restart the operator after a block, but that's easy to miss mid-incident.

I'd either read the AllowedMint account on every withdrawal just for the gate (one getAccountInfo), or have the indexer mirror `withdrawals_blocked` into `mints` from the BlockMint args it already parses and check that column here. The second one also lets the gateway refuse the burn up front.

Huzaifa696 (commented):
The trident harness (`tests/trident-tests/fuzz_escrow.rs`) only calls AllowMint during setup, so none of the new gate transitions get fuzzed. Would be good to add BlockMint with both flag combinations and the re-allow path to the instruction mix, doesn't have to be this PR.

private-channel-escrow-program/program/src/processor/allow_mint.rs
```diff
@@ -90,7 +89,7 @@ pub fn process_allow_mint(
         token_program_info,
     )?;
 
-    let allowed_mint = AllowedMint::new(args.bump);
+    let allowed_mint = AllowedMint::new(args.bump, mint_decimals, *token_program_info.address());
```
Huzaifa696: Pinning decimals and token program covers the arithmetic, but it leaves the main MintCloseAuthority case from 106 open: close at zero supply, recreate at the same address with the same decimals/program but a new permanent delegate, pause config, hook, or a freeze authority the original didn't have. Deposit accepts it and nothing forces a re-review. The interaction guide already calls this out as a residual.

Cheap way to close it: fold `get_extension_types()` into a u64 bitmask and record whether `freeze_authority` is set, store both here, and compare them in deposit next to the decimals check. ~9 bytes of state and no CPI. If the team would rather not pin, I'd at least put a written risk acceptance in the PR so Cantina can close 106 as acknowledged rather than partial.

indexer/src/operator/utils/mint_util.rs
```diff
@@ -269,6 +309,177 @@ impl MintCache {
...
+
+    /// Hook program the mint's `TransferHook` points at, or `None` for a mint
+    /// with no hook.
+    async fn transfer_hook_program(
+        &mut self,
+        mint: &Pubkey,
+    ) -> Result<Option<Pubkey>, OperatorError> {
+        let mint_str = mint.to_string();
+
+        if self.transfer_hook_cache.get(&mint_str) == Some(&false) {
```
Huzaifa696: Same class of problem as the greptile comment above, and it also applies to `extension_flags_cache`: the db reset in `upsert_mints_batch` on re-allow never reaches a running operator because it reads the in-memory copy first, so pause/delegate pre-flights keep evaluating the old profile until restart.

One fix covers both: when the indexer sees a re-AllowMint for a mint, drop that mint from `transfer_hook_cache`, `freeze_authority_cache` and `extension_flags_cache` so the next withdrawal re-resolves. If the extension bitmask lands on-chain the hook half goes away on its own, but the flags half still needs the invalidation.

Huzaifa696 (commented):
The new extension bitmask treats any change in the mint's extension list as "this mint was closed and recreated." But three extensions (metadata, group, group member) can be legitimately added to a live mint by its issuer, which is a routine step. When that happens, deposits for that mint stop until an admin re-allows it.

Fix: Ignore those three bits when comparing in Deposit, since they carry no power over transfers or escrow. The mask still catches the extensions that matter, like permanent delegate, pause, or hook.

Huzaifa696 (dismissed):

Huzaifa696 (approved):

PR #310, opened 2026-09-11
feat(core): accept v1 transactions by moving the workspace to Agave 4.2.2

Huzaifa696:
The v1 format only exists in the Agave 4.x crates, which ship together with the SVM, so the whole Solana dependency set moves to Agave 4.2.2. Incoming transactions are decoded with wincode, because bincode's layout of a v1 transaction is not the bytes the sender signed. v1 transactions can be up to 4096 bytes, and legacy and v0 stay at 1232. Rent becomes `Rent::free()`, since 4.2 removed the custom rent hook. Every batch shares one compiled runtime environment, so programs are never recompiled per batch. The channel is gasless, so a v1 transaction's compute budget is ignored.

This PR assumes greenfield deployments, so any database that already holds transaction rows must reset its transactions table first. Stored rows now use a versioned format holding the signed transaction bytes and our own metadata types. No Agave release can change that format, and compute units now come back instead of null. Rows from earlier builds are refused, never misread. `getTransaction` and `getBlock` return `-32015` instead of dropping the connection. The test geyser plugin is rebuilt for 4.2.2. keychain 1.4 hides the variable name in operator configuration errors.

A new end-to-end test sends legacy and v1 transactions through a real node, including one larger than 1232 bytes, and reads them back after a restart. The shutdown drain test is fixed and now runs in CI. CI retries the plugin-backed validator suites only when they hit a known upstream race, anza-xyz/agave#12799.

**Where the lines are.** Almost 10,000 of the changed lines are the lockfile. Of the rest, about 1,020 lines are where the review effort belongs, marked in the table below.

| category | review focus | added | removed | what it is |
|---|---|---:|---:|---|
| Stored row format | yes | 444 | 129 | versioned row, the transaction as signed wire bytes, fully owned metadata, one encoder and one decoder |
| SVM setup and execution | yes | 140 | 85 | processor rebuilt on the 4.2 cache API, one shared environment, `Rent::free()` |
| Ingress decode | yes | 107 | 63 | shared `decode_transaction`, wincode, size cap chosen by version |
| RPC reads | yes | 44 | 6 | `-32015` in place of `.unwrap()` when encoding a transaction for a reply |
| Mechanical API fallout | no | 156 | 213 | callback signature, removed trait method, import path moves, readers switched to the row decoder |
| Tests | no | 1,297 | 570 | new unit and end-to-end tests, plus struct literal updates in existing ones |
| Version pins | no | 154 | 116 | manifests, `versions.env`, `rust-toolchain.toml`, CI action |
| CI wiring | no | 55 | 8 | shutdown and v1 end-to-end tests added to CI, retry script for plugin-backed suites |
| Fixtures | no | 4 | 0 | committed stored rows that pin the format |
| Geyser plugin | no | binary | binary | rebuilt `.so` for 4.2.2 and its checksum |
| Lockfile | no | 3,844 | 6,115 | `Cargo.lock`, generated |
### Coverage Report

| Core | 24,416 | 25,765 | 94.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34898567254) |
| Indexer | 40,048 | 43,421 | 92.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34898567254) |
| Gateway | 1,653 | 1,918 | 86.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34898567254) |
| Auth | 1,758 | 1,962 | 89.6% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34898567254) |
| Withdraw Program | 129 | 241 | 53.5% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34898567172) |
| Escrow Program | 905 | 1,714 | 52.8% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34898567172) |
| E2E Integration | 17,027 | 22,193 | 76.7% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/34898567254) |
| **Total** | **85,936** | **97,214** | **88.4%** | |

> Last updated: 2026-09-14 22:06:29 UTC by E2E Integration


core/src/processor.rs
```diff
@@ -104,54 +136,67 @@ pub fn create_transaction_batch_processor<AccountsDB: TransactionProcessingCallb
...
 /// This functions is also a mock. In the Agave validator, the bank pre-checks
 /// transactions before providing them to the SVM API. We mock this step in
 /// PrivateChannel, since we don't need to perform such pre-checks.
+///
+/// Every transaction is framed with the same budget regardless of what it asked
+/// for. A v1 transaction carries its own compute unit limit, loaded data limit,
+/// heap size and priority fee in the message, and all four are ignored here on
+/// purpose: the channel is gasless, and compute budget instructions are already
+/// refused at admission, so honouring the message fields would be the only place
+/// a sender could influence its own budget.
```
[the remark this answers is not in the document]
Huzaifa696: Addressed in 00621c2b: a new end-to-end test sends a v1 transaction asking for 1 CU, a 1-byte loaded-data cap, a 1 SOL priority fee and max heap, and asserts it still lands and reports real compute usage.

indexer/src/operator/utils/rpc_util.rs
```diff
@@ -355,9 +355,18 @@ impl RpcClientWithRetry {
...
                     min_context_slot,
                 };
-                self.rpc_client
-                    .get_account_with_config(pubkey, config)
-                    .await
+                // The client returns the encoded form now, so decode it back to
+                // keep this wrapper's contract. Base64 is requested above, and a
+                // value that will not decode is reported as absent, which is the
+                // same answer the caller already handles for a missing account.
+                let response = self
```
sebatustra, line 362: An account that fails to decode is reported as "account does not exist"

The comment defends this: "a value that will not decode is reported as absent, which is the same answer the caller already  handles for a missing account."

But those are not the same answer. None here means "the chain says this account does not exist." A decode failure means "we could not read what the chain said." The callers of this wrapper include the operator's reconciliation and insolvency checks, which read escrow/custody accounts and mint supplies. A decode failure that reads as absence turns an integrity fault into a routine zero.

What makes this worth fixing rather than arguing about: this same PR deletes the test that stated the opposite principle.

core/src/rpc/send_transaction_impl.rs
sebatustra, line 1: The deserializer's allocation cap was dropped and nothing replaced it

Before, in both send_transaction_impl and simulate_transaction_impl:

```
  // Use bincode options matching Agave's decode_and_deserialize
  let bincode_options = bincode::options()
      .with_limit(PACKET_DATA_SIZE as u64)
      .with_fixint_encoding()
      .allow_trailing_bytes();
```

After, in core/src/rpc/decode.rs:

`wincode::deserialize::<VersionedTransaction>(tx_data).map_err(|e| { ... })
`
with_limit was doing real work: it bounded how much memory the deserializer could try to allocate while reading attacker-controlled bytes, not just how many it would read. wincode::deserialize uses the default config, whose preallocation ceiling is 4 MiB

indexer/src/operator/utils/rpc_util.rs
Huzaifa696: Fixed in 4d2d334d: undecodable account data is now an Err in every account wrapper (and in the channel supply read, which no longer calls the client's panicking decode), with tests.

core/src/rpc/send_transaction_impl.rs
Huzaifa696: Fixed in 4d2d334d: decoding now uses a wincode config capping preallocation at 256 KiB (main's short_vec reserved up to ~5 MiB via Vec::with_capacity, so with_limit never bounded this), with tests for a huge declared count and the densest valid transactions.

indexer/src/operator/utils/signer_util.rs
sebatustra, line 1: indexer/src/operator/utils/signer_util.rs:79 — startup no longer tells you which env var is missing.

keychain 1.4 redacts SignerError in both Display (#[error("Invalid private key format")]) and Debug ([REDACTED]). Every SignerError::InvalidPrivateKey(format!("{} not set", var)) in load_signer still builds the message, but nothing can print it. The panic an operator sees is:

ADMIN_SIGNER must be configured: SignerError::InvalidPrivateKey([REDACTED])

The tests were adapted to this by destructuring the variant (detail() at :265) rather than fixing the production path. Env var names are not secrets — log the name before returning, or build the expect message from it. Same loss at :82, where OPERATOR_SIGNER_INSTANCE collapses Err(_) into a generic warn and silently falls back to the admin signer.

Cargo.toml
sebatustra: the yellowstone pin comment now contradicts the pins under it.

indexer/src/operator/utils/signer_util.rs
Huzaifa696: Fixed in 3de9d8dd: load_signer now returns a local error whose config reasons (env var names only, never values) survive keychain redaction, so the admin panic and the operator fallback warn both name the variable, and the tests assert the printed message.

Cargo.toml
Huzaifa696: Fixed in 3de9d8dd: the comment now matches the pins (client 13.5 requires proto 12.7, the Solana 4.x line) and the v15.2.1 plugin tag in versions.env.

sebatustra (approved):

PR #317, opened 2026-09-15
convert client types at the trident boundary, add fuzz CI

sebatustra:
## What

Trident fuzz tests were broken by the Agave 4.2.2 bump (#310) and nothing caught it, because CI never built them.

  - **Type split.** The generated clients moved to solana-pubkey 4 / solana-instruction 3; trident-fuzz 0.12 is pinned to solana-sdk 2.x. Added `ToClient`/`ToTrident` conversions at the builder boundary (11 call sites).
  - **Borsh feature.** Both client crates derive Borsh over `Pubkey` without declaring `solana-pubkey/borsh` — they were borrowing it from the dvp client via workspace feature unification. `cargo check --workspace` passed, `cargo check -p` did not. Now declared on both.
  - **Dead patch.** Dropped the `trident-syscall-stubs-v2` fork; cargo reported it unused and both harnesses pass without it.
  - **CI.** New `fuzz-tests` job behind a `trident` paths filter, reusing the artifact
  `build-artifacts` already produces. ~7s per harness.
  
### Coverage Report

| Core | 26,189 | 27,559 | 95.0% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35115753374) |
| Indexer | 40,539 | 44,022 | 92.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35115753374) |
| Gateway | 1,653 | 1,918 | 86.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35115753374) |
| Auth | 1,755 | 1,962 | 89.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35115753374) |
| Withdraw Program | 129 | 241 | 53.5% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35115753555) |
| Escrow Program | 1,115 | 2,022 | 55.1% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35115753555) |
| E2E Integration | 17,341 | 22,571 | 76.8% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35115753374) |
| **Total** | **88,721** | **100,295** | **88.5%** | |

> Last updated: 2026-09-16 16:18:07 UTC by E2E Integration


private-channel-escrow-program/Makefile
```diff
@@ -79,6 +79,15 @@ integration-test: build integration-test-no-build
...
 all-test: unit-test integration-test
 
+# Run the Trident fuzz harnesses. Separate cargo workspace, so it builds on its
+# own; needs ../target/deploy/*.so and clients/rust from a prior build.
+# Each run picks a random master seed and prints it, so a failure replays with
+# TRIDENT_FUZZ_DEBUG=<seed>.
+fuzz:
+	@echo "Running fuzz harnesses for escrow program..."
+	@cd tests/trident-tests && cargo run --bin fuzz_escrow
+	@cd tests/trident-tests && cargo run --bin fuzz_rotate_bitmap
```
Huzaifa696: **Blocking: this job can never fail.**

When an `assert!` in a harness fails, Trident catches it, prints it, and keeps going. At the end it exits 0 anyway, unless you set `TRIDENT_WITH_EXIT_CODE`. So every check in both harnesses could be failing and CI would still be green.

```makefile
	@cd tests/trident-tests && TRIDENT_WITH_EXIT_CODE=1 cargo run --locked --bin fuzz_escrow
	@cd tests/trident-tests && TRIDENT_WITH_EXIT_CODE=1 cargo run --locked --bin fuzz_rotate_bitmap
```

That makes it exit 99 on failure, which make and Actions both read as a failure. I also added `--locked` so the lockfile in this PR actually gets used.

Side effect worth knowing: "both harnesses pass without it" (re: dropping the syscall-stubs patch) didn't tell us anything, since passing was the only possible outcome.

Refs: `flow_executor.rs:318-321` and `:409`.

.github/workflows/rust.yml
```diff
@@ -439,3 +444,34 @@ jobs:
...
+    name: Escrow Fuzz (Trident)
+    needs: [changes, build-artifacts]
+    if: needs.changes.outputs.trident == 'true'
+    runs-on: private-channel-runner-1
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup environment
+        uses: ./.github/actions/setup-environment
+        with:
```
Huzaifa696, line 470: **Blocking: nothing gets cached here, so every run rebuilds solana-sdk 2.x from scratch.**

The fuzz tests are their own separate cargo project, so they build into `tests/trident-tests/target`. `Swatinem/rust-cache` only knows about the main `target/` folder, so it never saves or restores this one. The "~7s per harness" is how long the fuzzing runs, not how long the job takes.

We can't just add the path inside `setup-environment/action.yml`, because rust-cache builds its cache key from every project listed there. Doing that would change the key for every job in the repo. Pass it in as an input instead:

```yaml
# setup-environment/action.yml
inputs:
  rust-cache-workspaces:
    required: false
    default: ". -> target"

    - uses: Swatinem/rust-cache@v2
      with:
        workspaces: ${{ inputs.rust-cache-workspaces }}
```

then here:

```yaml
          rust-cache-workspaces: |
            . -> target
            private-channel-escrow-program/tests/trident-tests -> target
```

(Leaving this here because the action file isn't part of this diff.)

.github/workflows/rust.yml
```diff
@@ -136,6 +137,10 @@ jobs:
               - 'auth/**'
               # Re-vendoring the DvP client runs its tests via ci-unit-coverage.
               - 'dvp-swap-program/**'
+            trident:
+              - '.github/workflows/rust.yml'
+              - '.github/actions/setup-environment/**'
+              - 'private-channel-escrow-program/**'
```
Huzaifa696, line 149: **Blocking: this filter wouldn't have caught the bug this PR is fixing.**

The escrow client takes `borsh`, `thiserror` and others from the root `Cargo.toml`, and the fuzz tests use the Rust version from the root `rust-toolchain.toml`. So someone can break the fuzz build by editing only those root files, without touching anything under `private-channel-escrow-program/`. This job wouldn't run, and we'd be right back here. That's what happened with #310.

```yaml
            trident:
              - 'Cargo.toml'
              - 'Cargo.lock'
              - 'rust-toolchain.toml'
              - '.github/workflows/rust.yml'
              - '.github/actions/setup-environment/**'
              - 'private-channel-escrow-program/**'
```

private-channel-escrow-program/Makefile
sebatustra: Confirmed and fixed

.github/workflows/rust.yml
sebatustra: Correct, and the input approach is right for the reason you give. Added rust-cache-workspaces to setup-environment defaulting to ". -> target", which is rust-cache's own default, so the other 13 callers keep identical keys

.github/workflows/rust.yml
sebatustra: Added all three paths

Huzaifa696 (approved):

PR #318, opened 2026-09-15
fix(indexer): check ledger liabilities against custody at runtime, pinned to the custody slot (issue-43)

Huzaifa696:
Runtime reconciliation only checked channel supply against custody plus the in-flight envelope. A landed deposit that is still `pending`, or `failed` and never minted, sits in custody without entering supply, so a custody drain of that size passed every tick. The shared ledger query had gaps of its own: it subtracted a withdrawal once its row read `completed`, which lags the payout and has no Solana slot, and it bounded withdrawal rows, which carry a private channel slot, by the Solana custody slot.

Runtime now also checks that ledger liabilities do not exceed custody plus the existing bps tolerance, with both sides pinned to the custody snapshot slot. The tick waits up to 30 s for the escrow indexer's checkpoint to reach that slot; if it does not, or no checkpoint exists, liabilities are unknown and the counter holds. In the shared query the slot bound applies to deposits only, and a withdrawal counts as released only when `observed_releases` has its release at or below the slot, so a later release cannot mask a drain and an unindexed one fails closed. Each invariant keeps its own counter on the same three-tick rule and halt path. Startup skips only the ledger comparison when its checkpoint is below the snapshot, since that ledger lacks recent releases. `get_mint_addresses` is gone because the ledger rows enumerate every mint.

Unit tests cover the check, the halt layer and every wait outcome; Postgres tests pin the query at its slot boundaries. This assumes greenfield deployments, so `observed_releases` holds every payout, and finalized-only ingestion. No new config, metrics, or per-status split.

### Coverage Report

| Core | 26,188 | 27,559 | 95.0% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35160449221) |
| Indexer | 42,340 | 45,825 | 92.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35160449221) |
| Gateway | 1,653 | 1,918 | 86.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35160449221) |
| Auth | 1,755 | 1,962 | 89.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35160449221) |
| E2E Integration | 17,630 | 22,848 | 77.2% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35160449221) |
| **Total** | **89,566** | **100,112** | **89.5%** | |

> Last updated: 2026-09-16 23:47:37 UTC by E2E Integration


indexer/src/operator/reconciliation.rs
```diff
@@ -261,18 +321,72 @@ async fn perform_reconciliation_check(
...
+            }
+            Err(e) => warn!("Checkpoint read failed while waiting for the ledger: {}", e),
+        }
+        if started.elapsed() >= LEDGER_CATCHUP_TIMEOUT {
+            warn!(
+                checkpoint = ?last,
+                slot,
+                "Escrow indexer checkpoint did not reach the custody slot in time; ledger liabilities unknown this tick"
+            );
+            return LedgerWait::Unknown;
```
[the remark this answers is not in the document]
Huzaifa696: Checked this against the geyser build we run, and the premise does not hold: a `blocks` subscription with `account_include` still receives one block message per produced block. `account_include` filters the transactions inside the message, not which blocks are delivered.

In `yellowstone-grpc-geyser/src/plugin/filter/filter.rs`, `FilterBlocks::get_updates` loops over the registered filters and pushes a `FilteredUpdateBlock` for every one of them on every `MessageBlock`. The only thing `account_include` does there is drive the `filter_map` that builds `transactions`; an empty list still ships a block update. So after the last escrow instruction the stream keeps delivering blocks, `handle_block` keeps emitting `SlotComplete`, and the checkpoint keeps advancing. `Slot` updates indeed forward nothing, but they are not what advances it.

This is also what the new end-to-end test exercises. In `a_real_drain_halts_the_runtime_refuses_the_next_startup_and_holds_while_unpinnable`, the drain is a Token-2022 permanent-delegate transfer, so it invokes no escrow instruction, and no escrow instruction lands after the deposit for the rest of the phase. The runtime still halts on the third tick, which it can only do if the checkpoint reached a custody snapshot slot taken after the drain. Phase 3 of the same test then stops the indexer, and only then does the arm read unknown and hold, which is the intended behaviour.

RPC polling is unaffected either way: it checkpoints each polled slot regardless of content.

No change made. If you can point at a deployment where the blocks filter suppresses empty blocks, I will revisit, since the wait budget would then need a different design rather than a longer timeout.

indexer/src/storage/postgres/db.rs
```diff
@@ -185,6 +185,45 @@ const DROP_STATEMENTS: [&str; 10] = [
...
+/// the bound.
+const RELEASED_BY_OBSERVATION: &str = "EXISTS (SELECT 1 FROM observed_releases r
+                         WHERE r.withdrawal_nonce = t.withdrawal_nonce
+                           AND r.slot <= $1)";
+
+/// The same, plus the operator's own record of a confirmed payout.
+const RELEASED_BY_OBSERVATION_OR_STATUS: &str = "(t.status = 'completed' OR EXISTS (
+                         SELECT 1 FROM observed_releases r
+                         WHERE r.withdrawal_nonce = t.withdrawal_nonce
+                           AND r.slot <= $1))";
```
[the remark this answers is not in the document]
Huzaifa696: The property is real, but its window is much narrower than the finding implies, and it is the deliberate trade-off this fallback exists for.

A withdrawal row reaches `completed` only after its release is confirmed **finalized** (a confirmed-but-not-finalized status is explicitly not settled, see `poll_in_flight_confirmed_not_finalized_does_not_settle`). So at the instant a row is completed, the finalized tip is already at or above its release slot. The custody snapshot is itself a finalized read. The only payout this fallback can therefore subtract early is one finalized *between* the custody reading and this query, which run back to back: seconds, and independent of how far the escrow indexer trails.

The alternatives are worse in the same direction:

- Deferring the whole comparison when the checkpoint lags, which is what this PR did before, masks 100% of a shortfall rather than a seconds-wide slice of one. Under `backfill.enabled = false` (`scripts/devnet/config/indexer-solana.toml`) the checkpoint lags on essentially every boot after downtime, so that removed startup's drain check entirely.
- Comparing strictly against a ledger that cannot be pinned reports every unindexed payout as a shortfall and fails healthy boots.

Custody cannot be re-read at a past slot and the yellowstone gap-fill is unbounded (`max_gap_slots`), so waiting for coverage is not available here either. Note also that this is strictly tighter than the pre-PR behaviour, which subtracted `completed` rows with a channel-slot bound applied to them; that bound could drop a genuinely paid withdrawal and false-abort.

Runtime reconciliation is unaffected: it holds out for the observed release and reports unknown instead of guessing, because it can afford to wait.

Pushed d032bac9, which states the window in the code at both sites rather than leaving it to be rediscovered. Not changing the behaviour.

indexer/src/operator/reconciliation.rs
```diff
@@ -229,12 +281,18 @@ async fn perform_reconciliation_check(
         *halted = flag.is_some();
     }
 
-    // Custody (Solana) plus the DB mint set enumerate the mints to check; the DB
-    // set keeps a zero-custody mint with outstanding supply in scope. The union is
-    // built once here and shared with both the supply fetch and the evaluation.
     let custody = fetch_on_chain_balances(rpc_client, escrow_instance_id).await?;
-    let mut mints: HashSet<Pubkey> = custody.keys().copied().collect();
-    mints.extend(fetch_db_mint_set(storage).await?);
+    let covered = match wait_for_ledger(storage, custody.slot, cancellation_token).await {
```
sebatustra, line 316: The supply invariant now compares custody and supply taken seconds apart.

The new ledger wait was inserted between the custody read and the supply read:
```
let custody = fetch_on_chain_balances(rpc_client, escrow_instance_id).await?;   // custody @ slot S
let covered = match wait_for_ledger(storage, custody.slot, cancellation_token).await { // blocks up to 30s
  ...
  match load_halt_inputs(storage, channel_rpc, &mints).await {                     // supply read HERE
```
load_halt_inputs (:419-426) reads the envelope and then one getAccountInfo per mint on the channel chain. So the supply arm now evaluates supply(t₁) vs custody(S) + envelope(t₁) where t₁ - t₀ is the whole ledger wait.

That wait is not rare. The indexer ingests at Finalized (datasource/rpc_polling/source.rs:618) and the checkpoint writer flushes every 5s (checkpoint.rs:157), so the committed checkpoint normally trails the RPC's finalized tip — meaning wait_for_ledger polls for seconds on a healthy tick, not just a degraded one.

Failure: a deposit finalizes after S, gets indexed, gets minted, and its row reaches completed — all inside the window. Supply is up; custody is the pre-deposit reading; and the envelope no longer covers it, because get_in_flight_amounts_by_mint only sums pending/processing/parked/pending_remint (db.rs:2813). The gap is a pure artifact. With the default 10 bps cushion, any such deposit above 0.1% of custody counts as a breach; three consecutive ticks quarantine every active withdrawal.

indexer/src/operator/reconciliation.rs
```diff
@@ -261,18 +321,72 @@ async fn perform_reconciliation_check(
...
+    let mut last: Option<u64> = None;
+    loop {
+        match storage.get_committed_checkpoint(&key).await {
+            Ok(Some(committed)) if committed >= slot => return LedgerWait::Covered,
+            Ok(Some(committed)) => last = Some(committed),
+            // No escrow indexer has ever committed, so waiting cannot help.
+            Ok(None) => {
+                warn!(
+                    slot,
+                    "No escrow indexer checkpoint; ledger liabilities unknown this tick"
```
sebatustra, line 394: The liability invariant silently switches itself off, with nothing to page on

Both give-up paths return Unknown, which makes liabilities None and turns the whole liability arm into a no-op for that tick:
```
  Ok(None) => {
      warn!(slot, "No escrow indexer checkpoint; ledger liabilities unknown this tick");
      return LedgerWait::Unknown;
  }
  ...
  warn!(checkpoint = ?last, slot, "Escrow indexer checkpoint did not reach the custody slot in time; ...");
  return LedgerWait::Unknown;
```

There is no metric anywhere in this module (grep metrics:: indexer/src/operator/reconciliation.rs → nothing), no health signal, no webhook, and no bound on how many consecutive Unknown ticks are acceptable. The counter holds, so the check just stops existing.

Why that matters: the states that produce Unknown are exactly the states where a drain is hardest to notice — escrow indexer down, backlogged, mid-resync, or its checkpoint deliberately frozen by the oversized-gap guard (datasource/yellowstone/source.rs:249-256, which freezes the checkpoint indefinitely until a human intervenes). In that last case the invariant is off forever and the only evidence is a warn! every 5 minutes.

indexer/src/storage/postgres/db.rs
```diff
@@ -185,6 +185,47 @@ const DROP_STATEMENTS: [&str; 10] = [
     "DROP TYPE IF EXISTS transaction_type CASCADE",
 ];
 
+/// A withdrawal counts as paid out when the escrow indexer recorded its release at or below
+/// the bound.
+const RELEASED_BY_OBSERVATION: &str = "EXISTS (SELECT 1 FROM observed_releases r
```
sebatustra: The ledger subtracts the intended withdrawal amount, not the released one

The escrow program computes the real custody delta and puts it in the event:
```
  // release_funds.rs:163-166
  let released = escrow_token_balance_before.checked_sub(escrow_token_balance_after)...
  let event = ReleaseFundsEvent::new(instance.instance_seed, *operator_info.address(), released, ...);
```

The parser already carries the instruction amount (parser/escrow.rs:191-195: ReleaseFundsData { amount, user, transaction_nonce }). The indexer throws it away:

```
  EscrowInstruction::ReleaseFunds { data, .. } => (
      None, None, None,
      Some(DbObservedRelease {
          withdrawal_nonce: data.transaction_nonce as i64,
          signature: signature.clone(),
          slot: instruction_meta.slot as i64,
      }),
  ),
```

observed_releases then holds only (nonce, signature, slot), and the aggregate subtracts t.amount from the transactions row:
```
  SUM(CASE WHEN t.transaction_type = 'withdrawal'
            AND EXISTS (SELECT 1 FROM observed_releases r
                        WHERE r.withdrawal_nonce = t.withdrawal_nonce AND r.slot <= $1)
           THEN t.amount ELSE 0 END)
```
release_funds validates the nonce bitmap but never checks args.amount against anything the row knows. So if a release moves less than the row says, liabilities drop by the full row amount while custody drops by the actual amount — the difference is headroom a real drain can hide in, permanently. Example: row for nonce N says 100_000; a release for nonce N moves 1; the ledger now believes 100_000 left escrow, and 99_999 of custody can go missing elsewhere without the check ever seeing it

indexer/src/operator/reconciliation.rs
```diff
@@ -398,7 +511,61 @@ async fn evaluate_and_maybe_halt(
...
+        // Unknown liabilities hold the counter: this tick is no evidence either way.
+        let Some(liabilities) = liabilities else {
+            if let Some(&held) = breach_counters.liability.get(&mint) {
+                next_counters.liability.insert(mint, held);
+            }
+            continue;
+        };
+        let c = *custody.get(&mint).unwrap_or(&0);
+        let owed = *liabilities.get(&mint).unwrap_or(&0);
+        let tolerance = insolvency_tolerance_raw(c, config.reconciliation_tolerance_bps);
```
sebatustra, line 613: Runtime and startup now share one liability model but not one threshold

Runtime:
  let tolerance = insolvency_tolerance_raw(c, config.reconciliation_tolerance_bps); // 10 bps of custody
  let Some(breach) = evaluate_liability_shortfall(c, owed, tolerance) else { continue };
Startup:
  .filter(|r| r.shortfall() > config.mismatch_threshold_raw)   // default 0 — any shortfall blocks

Same comparison, different units and very different defaults (config.rs:262 default 0 raw; config.rs:400 default 10 bps). A standing shortfall of, say, 0.05% of custody is invisible to the runtime check forever and refuses the next boot. Operators get a system that runs clean for months and then will not restart, with the runbook pointing at a halt reason that was never emitted.

indexer/src/operator/reconciliation.rs
Huzaifa696: Confirmed regression: the wait sat between the two readings the supply arm compares. The tick now reads custody, envelope and channel supply back to back and waits for the ledger after, which is main's ordering. `supply_is_read_before_the_ledger_wait` pins it: 83 checkpoint reads had landed before the supply request, now 0.

indexer/src/operator/reconciliation.rs
Huzaifa696: Both `Unknown` paths now advance a cross-tick counter, exported as `..._liability_dark_ticks` (0 when armed) plus `..._liability_unknown_total{reason}`, and every third dark tick logs `reconciliation_alert` and posts a webhook so a frozen checkpoint keeps firing. One gap stays, unchanged from main: a failed channel-supply read skips both arms with only a `warn!`.

indexer/src/storage/postgres/db.rs
Huzaifa696: Fixed: `observed_releases` gains a nullable `amount` from the release instruction, and both aggregates subtract `LEAST(COALESCE(r.amount, t.amount), t.amount)`. The cap matters as much as the amount, since subtracting the observed figure alone would let a release larger than its row net the theft to zero. An e2e covers it: a real 4,000 release on a row owing 40,000, then a 36,000 drain of the headroom.

indexer/src/operator/reconciliation.rs
Huzaifa696: Agreed on the failure, but I fixed the silence rather than the strictness: defaulting the boot to 10 bps weakens a guard shipping on main today, on the one path where strictness is free. Every shortfall is now exported per mint each tick, sub-tolerance included, so the state that refuses a restart is visible immediately. `reconciliation_tolerance_bps` is on `ReconciliationConfig` using the same formula, defaulting to 0; matching the defaults is a one-line change if you prefer.

indexer/src/bin/indexer.rs
```diff
@@ -44,6 +44,8 @@ struct StorageSection {
 struct ReconciliationSection {
     #[serde(default)]
     mismatch_threshold_raw: u64,
+    #[serde(default)]
+    reconciliation_tolerance_bps: u16,
```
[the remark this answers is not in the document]
Huzaifa696: Accurate, and deliberate. Closing it the other way means defaulting the boot to tolerate 10 bps of custody, which weakens a guard shipping on main today on the one path where strictness costs nothing. Instead every shortfall is now exported per mint each tick, sub-tolerance included, so a standing gap is visible long before a restart rather than surfacing as a refused boot. The knob uses the same formula on both sides, so matching the defaults is a one-line change if that is preferred.

sebatustra (approved):

PR #320, opened 2026-09-15
feat(escrow): support memo-required destinations on ReleaseFunds

sebatustra:
## What

Token-2022's `MemoTransfer` is an account extension: a user enabling it on their own ATA permanently bricked their withdrawals, since the escrow's `TransferChecked` CPI had no memo sibling. A client-side top-level memo can't fix it — it sits one stack frame above the CPI, so Token-2022 never sees it as a sibling.

The program now emits the memo CPI itself, immediately before the transfer, following [solana-foundation/dvp](https://github.com/solana-foundation/dvp).

## Changes

  - `ReleaseFunds` gains a `memo_program` account at index 13 (fixed prefix 13 → 14). Validated only on the path that uses it.
  - `Deposit` passes `None`: its destination is the PDA-owned escrow ATA, which can never carry the extension — only an owner-signed `EnableRequiredMemoTransfers` sets it, and no mint extension derives it.
  - Codama defaults `memo_program` alongside `systemProgram`/`tokenProgram`, so **no instruction builder call site changes**.
  - Operator's legacy-packet cap on transfer-hook extras drops 16 → 15: the new account eats the same 33 bytes an extra does.

Appending rather than inserting keeps indices 0..12 stable, so the indexer's `parse_release_funds` is untouched.

Deploy order: this is a breaking layout change — ReleaseFunds goes from 13 to 14 fixed accounts — so operator and program must be rolled out in that order: operator first, then program. A new operator against the old program passes memo_program in the slot the old code reads as the first transfer-hook extra: inert for plain mints (extras are only forwarded when the mint carries a hook), and for hook mints the extra account is forwarded to the hook during the window, which is untested and may fail. We accept that for the short deploy gap; hook mints are not in production use today. The reverse order is strictly worse: the old operator sends 13 accounts and every ReleaseFunds fails NotEnoughAccountKeys until the operator ships, i.e. a full withdrawal outage rather than a narrow one.

## Tests

62 program unit, 99 escrow integration (3 new: memo-required destination succeeds and logs the Memo CPI; wrong account in the slot → `IncorrectProgramId` with nothing moved; memo + transfer hook together, both firing), 1011 indexer unit. fmt + clippy clean.

### Coverage Report

| Core | 26,186 | 27,559 | 95.0% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35164772465) |
| Indexer | 40,854 | 44,335 | 92.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35164772465) |
| Gateway | 1,653 | 1,918 | 86.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35164772465) |
| Auth | 1,755 | 1,962 | 89.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35164772465) |
| Withdraw Program | 129 | 241 | 53.5% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35164772457) |
| Escrow Program | 1,115 | 2,048 | 54.4% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35164772457) |
| E2E Integration | 17,296 | 22,636 | 76.4% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35164772465) |
| **Total** | **88,988** | **100,699** | **88.4%** | |

> Last updated: 2026-09-17 12:45:01 UTC by Auth


private-channel-escrow-program/idl/private_channel_escrow_program.json
```diff
@@ -1610,6 +1610,15 @@
...
+          },
+          {
+            "docs": [
+              "SPL Memo program; only used when user_ata requires a memo"
+            ],
+            "isSigner": false,
+            "isWritable": false,
+            "kind": "instructionAccountNode",
+            "name": "memoProgram"
           }
```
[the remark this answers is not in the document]
sebatustra: Good catch, fixed

private-channel-escrow-program/program/src/processor/release_funds.rs
```diff
@@ -27,7 +27,7 @@ use pinocchio::{
 const INSTRUCTION_DATA_LENGTH: usize = 8 + 32 + 8;
 
 /// Fixed account prefix; anything past it is transfer-hook extras.
-const FIXED_ACCOUNTS_LEN: usize = 13;
+const FIXED_ACCOUNTS_LEN: usize = 14;
```
Huzaifa696, line 30: Breaking layout change with no safe deploy order: the old operator sends 13 accounts and now fails `NotEnoughAccountKeys`, while a new operator against the old program has this account misread as the first hook extra (inert for plain mints, untested for hook mints).

**Fix:** state the deploy order in the PR description (operator first, then program) and confirm or accept hook-mint breakage during the window.

indexer/src/operator/processor.rs
```diff
@@ -811,10 +811,13 @@ async fn check_withdrawal_preflights_inner(
...
 /// the sender builds a legacy message with no lookup tables, and the 1232-byte
-/// packet holds 699 bytes of release plus 33 per extra, so 16 fit and 17 does
+/// packet holds 732 bytes of release plus 33 per extra, so 15 fit and 16 does
 /// not. Raise it toward 32 if the sender ever moves to versioned transactions
 /// with a lookup table, or to v1 transactions with their larger limit.
-const MAX_HOOK_EXTRAS_LEGACY_TX: usize = 16;
+///
+/// The release grew by one account key plus its index byte, the same 33 bytes
+/// an extra costs, when ReleaseFunds gained its memo_program slot.
+const MAX_HOOK_EXTRAS_LEGACY_TX: usize = 15;
```
Huzaifa696, line 820: The 732-bytes-plus-33-per-extra math is correct, but it lives only in this comment, so the next account added to the release path becomes a production "transaction too large" rather than a failing test.

**Fix:** add a test asserting a 15-extra release serializes to <= 1232 bytes and a 16-extra one exceeds it.

private-channel-escrow-program/program/src/processor/deposit.rs
```diff
@@ -134,6 +134,11 @@ pub fn process_deposit(
         args.amount,
         mint_decimals,
         token_program_info.address(),
+        // No memo slot: the destination is the instance's own ATA, and required
+        // memos are only ever set by an owner-signed EnableRequiredMemoTransfers,
+        // which this program never issues, or derived from a mint extension,
+        // which never yields MemoTransfer.
+        None,
```
Huzaifa696, line 141: The `None` arm's safety rests entirely on the comment above it; nothing tests that it fails closed instead of silently skipping the memo.

**Fix:** add a deposit test into an instance ATA with `MemoTransfer` forced on, asserting an error. `set_token_2022_with_memo_account` already exists.

private-channel-escrow-program/program/src/processor/release_funds.rs
sebatustra: Agreed. added in PR description

indexer/src/operator/processor.rs
sebatustra: Added hook_extras_cap_is_the_widest_release_a_legacy_packet_holds in processor.rs. It builds a real release via build_release_funds, appends N distinct-key extras, runs it through the sender's own build_and_sign, and asserts the serialized transaction is ≤ 1232 bytes at the cap and > 1232 one past it.

private-channel-escrow-program/program/src/processor/deposit.rs
sebatustra: Added test_deposit_token_2022_memo_required_instance_ata_fails: set_token_2022_with_memo_account on the instance ATA, asserting NotEnoughAccountKeys and that neither balance moved. That's the memo_program.ok_or(...) in transfer_checked_cpi — the deposit dies before the transfer CPI rather than transferring memo-less.

Huzaifa696 (commented):
RELEASE_FUNDS_ACCOUNTS = 13 no longer matches the program's 14-account fixed prefix. It's a < floor so nothing breaks today, but the guard would wave through a 13-account release the program can no longer execute. 

bump it to 14.

indexer/src/indexer/datasource/common/parser/escrow.rs
```diff
@@ -26,7 +26,7 @@ const ROTATE_BITMAP: u8 = 8;
 // is not the configured one is dropped before it reaches storage.
 const CREATE_INSTANCE_ACCOUNTS: usize = 8;
 const BLOCK_MINT_ACCOUNTS: usize = 7;
-const RELEASE_FUNDS_ACCOUNTS: usize = 13;
+const RELEASE_FUNDS_ACCOUNTS: usize = 14;
```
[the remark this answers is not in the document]
sebatustra: this has not been deployed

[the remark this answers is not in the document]
sebatustra: the whole system has not been deployed in production, ever, only devnet and we dont care about indexing those

Huzaifa696 (approved):

PR #287, opened 2026-08-31
fix(gateway): reject token-account delegates on getSignaturesForAddress (finding-225)

sebatustra:
## Issue

The gateway's account gate authorized `getSignaturesForAddress` with the same check it uses for current-state reads. For SPL Token and Token-2022 accounts that check accepts either the token account's owner or its current delegate.

A delegate is a current spend authority, often temporary and allowance-scoped. Treating it as ownership let anyone controlling a delegate wallet read the address' full signature history, including activity from before the delegation existed and unrelated to the delegated allowance. For PrivateChannel users that discloses transaction timing and counterpart activity outside the delegation's scope. It is reachable in practice: SPL Token 8.0.0 runs in-channel as a precompile, so users can call `Approve` themselves.

## Fix

Split the policy by method. `OWNER_ONLY_METHODS` lists the gated methods a delegate must not unlock, currently just `getSignaturesForAddress`. `check_token_account_ownership` now takes the method and rejects those once the owner check fails, before the delegate check runs.

`getAccountInfo` and `getTokenAccountBalance` are unchanged: a delegate can spend the balance, so reading current state discloses nothing the delegation did not already grant.

The guard sits after the owner lookup, so approving a delegate never costs the owner access to its own history.
### Coverage Report

| Component | Lines Hit | Lines Total | Coverage | Artifact |
|-----------|-----------|-------------|----------|----------|
| Core | 17,750 | 19,421 | 91.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33522619006) |
| Indexer | 38,493 | 42,213 | 91.2% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33522619006) |
| Gateway | 1,648 | 1,907 | 86.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33522619006) |
| Auth | 1,312 | 1,506 | 87.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/33522619006) |
| Withdraw Program | - | - | - | - |
| Escrow Program | - | - | - | - |
| DvP Swap Program | - | - | - | - |
| E2E Integration | - | - | - | - |
| **Total** | **59,203** | **65,047** | **91.0%** | |

> Last updated: 2026-09-01 15:05:32 UTC by Auth

PR #321, opened 2026-09-16
Fix/issue 225 v2

sebatustra:
## finding-225

`getSignaturesForAddress` authorized a token account's history from its current owner. An SPL `SetAuthority { AccountOwner }` handoff therefore gave the new owner every signature from before the transfer. The earlier fix (#287) closed the delegate half of the finding and documented this half as accepted.

## Change

The node records every landed `AccountOwner` handoff in a new `token_account_owner_change` table, both sides of it, inside the same commit as the block. The gateway reads that chain, works out which slot windows the caller owned the address for, and passes them to the query so `limit` counts rows they may actually see. Cursors and paging behave as before.

A timeline that cannot be accounted for serves nothing rather than guessing: a chain that does not reach the current owner, one longer than the cap, or history older than the slot the table started recording at.

## Residual

A former owner still cannot read the history they made with an account they gave away, since they fail the current-state ownership check first. That is the pre-existing limitation already documented for closed accounts.

## Testing

Unit and integration coverage for the detector, the window builder, the scoped query and the gateway path, including handoff, handoff-and-return, same-slot ordering, over-long chains and the watermark.
### Coverage Report

| Core | 26,704 | 28,087 | 95.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35222998424) |
| Indexer | 42,340 | 45,825 | 92.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35222998424) |
| Gateway | 1,993 | 2,296 | 86.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35222998424) |
| Auth | 1,755 | 1,962 | 89.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35222998424) |
| Withdraw Program | 129 | 241 | 53.5% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35222998391) |
| Escrow Program | 1,115 | 2,022 | 55.1% | [unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35222998391) |
| E2E Integration | 17,748 | 23,085 | 76.9% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35222998424) |
| **Total** | **91,784** | **103,518** | **88.7%** | |

> Last updated: 2026-09-17 13:34:13 UTC by E2E Integration


core/src/accounts/owner_change.rs
```diff
@@ -0,0 +1,459 @@
...
+/// ```
+///
+/// Accounts are `[target, current_authority, ..multisig_signers]`. The program
+/// validates the signer against the token account's `owner` field, so account 1
+/// is that field's value going into the instruction — the previous owner, with
+/// no pre-state lookup needed.
+///
+/// Inner instructions are not scanned: CPI recording is off, so they are not
+/// available here. No admitted program CPIs `SetAuthority`, and a handoff that
+/// did slip through a CPI breaks the recorded chain rather than widening it.
```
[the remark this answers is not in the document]
sebatustra: fixed. You're right that the comment overstated the guarantee. The chain-break argument
  only holds for an account that already has a recorded handoff; on one with none,
  an empty table read as "never changed hands" and the history was served unscoped,
  so a CPI handoff there would have leaked it.

  Rather than rely on admitted programs never doing this, the gateway no longer
  treats an absent row as proof on its own. On the no-handoff path it checks the
  address still derives as the ATA of its current owner field. Users can only hold
  ATAs (`CreateAccount` is not admitted at ingress, so only the ATA program can
  create them) and the address is derived from the owner at creation, so an owner
  the address no longer derives to means the owner field was moved, whatever moved
  it. That case now serves no history.

  This does not depend on CPI recording or on enumerating which programs might CPI
  `SetAuthority`.

gateway/src/auth.rs
```diff
@@ -302,6 +320,194 @@ pub async fn check_account_data_ownership(
...
+        }
+        // A ledger recorded from genesis can vouch for the whole history.
+        if indexed_from == 0 {
+            record_scope_outcome("never_handed_on");
+            return Ok(None);
+        }
+        record_scope_outcome("watermarked");
+        return Ok(Some(vec![(indexed_from, i64::MAX)]));
+    }
+    if changes.len() as i64 > MAX_OWNER_CHANGES {
```
Huzaifa696: Records are never deleted and the chain only grows, so this is permanent. Whoever holds the account can pass it around 65 times and hand it back, and the real owner loses their history for good. Consider counting only from the caller's earliest owned window, so old churn they can't read anyway doesn't count against them.

gateway/tests/auth_integration.rs
```diff
@@ -1519,6 +1689,7 @@ async fn test_get_signatures_for_address_delegate_returns_403() {
 async fn test_get_signatures_for_address_token_account_owner_is_proxied() {
     let (pool, _url, _container) = start_postgres().await;
     db::init_schema(&pool).await.unwrap();
+    init_owner_change_table(&pool).await;
```
Huzaifa696, line 1692: This test's address doesn't derive from the owner, so the user actually gets an empty page here, and the test passes anyway because 200 comes back either way. Same in `test_get_signatures_for_address_owner_keeps_history_after_delegating` below. Use `untouched_ata_of(&owner_bytes)` and assert `forwarded_slot_ranges(&requests) == None`, like `..._unhandled_account_is_not_scoped` does.

gateway/src/auth.rs
sebatustra: Good catch, and the grief is worse than a length check can fix from the oldest end — the real owner is usually prev_owner of row 0, so counting from their earliest owned window still spans the churn. Made the cap a read horizon instead: owner_changes now reads the newest 64 rows (ORDER BY slot DESC, reversed), and past the cap we serve what we hold rather than nothing. Each window is attributed by the handoffs either side of it plus the current-owner anchor, so a suffix proves just as much; only the head range (0, first-1) is withheld, since a truncated read can't show row 0 is the account's first. Churn now just moves the horizon forward — the owner keeps the window from the handback on. New scoped_to_newest_handoffs metric label for when that happens.

gateway/tests/auth_integration.rs
sebatustra: Right, and the old chain in overlong_chain_serves_nothing didn't link either, so it was passing through the broken-link path. Both tests now use untouched_ata_of(&owner_bytes) with the recording backend and assert forwarded_slot_ranges ==  None. Confirmed load-bearing: with the arbitrary address put back, it fails left: Some(Array []), right: None

Huzaifa696 (approved):

PR #323, opened 2026-09-17
fix(indexer): unpack Token-2022 escrow accounts with StateWithExtensions

Huzaifa696:
Supersedes #77, rebased onto current `main`.

## Problem

`fetch_escrow_balances_by_mint` sweeps token accounts from both `spl_token` and
`spl_token_2022`, but unpacked every one with `spl_token`'s `Account::unpack`. That
requires exactly 165 bytes and rejects Token-2022 accounts carrying extensions
(minimum 166 bytes, with an `AccountType` discriminator after the base state).

The decode error is fatal to the whole sweep, not just the one account, so a single
extended Token-2022 escrow account aborts the entire reconciliation run.

## Fix

Dispatch on `token_program_id` and use `StateWithExtensions::<Token2022Account>::unpack`
for the Token-2022 program, which parses the base state and tolerates the extension
bytes that follow. The SPL Token path is unchanged.

## Credit

Original fix by @0xKyungmin in #77. The rebase needed a test rewrite because `main`'s
test module drifted after that PR was opened: `EMPTY_BODY` became `empty_body(slot)`,
`result_body` gained a slot argument, the function now returns `CustodySnapshot`, and
a slot-agreement retry was added across the two token-program calls. The production
change merged into `sweep_once` unmodified.

### Coverage Report

| Core | 26,707 | 28,087 | 95.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35340608380) |
| Indexer | 42,517 | 46,001 | 92.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35340608380) |
| Gateway | 1,993 | 2,296 | 86.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35340608380) |
| Auth | 1,758 | 1,962 | 89.6% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35340608380) |
| E2E Integration | 17,747 | 23,092 | 76.9% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35340608380) |
| **Total** | **90,722** | **101,438** | **89.4%** | |

> Last updated: 2026-09-18 12:24:28 UTC by E2E Integration


indexer/src/operator/escrow_sweep.rs
```diff
@@ -650,4 +660,63 @@ mod tests {
             read.reason
         );
     }
+
+    fn base64_token2022_account(mint: Pubkey, amount: u64) -> String {
```
sebatustra: Test fixture uses a layout Token-2022 never produces

indexer/src/operator/escrow_sweep.rs
```diff
@@ -650,4 +660,63 @@ mod tests {
...
+            amount,
+            delegate: COption::None,
+            state: spl_token_2022::state::AccountState::Initialized,
+            is_native: COption::None,
+            delegated_amount: 0,
+            close_authority: COption::None,
+        };
+        let mut buf = vec![0u8; Token2022Account::LEN];
+        account.pack_into_slice(&mut buf);
+        buf.push(spl_token_2022::extension::AccountType::Account as u8);
```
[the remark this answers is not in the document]
Huzaifa696: Good catch. The fixture now carries a real `TransferHookAccount` TLV entry and is sized by `ExtensionType::try_calculate_account_len`, so it exercises actual extension parsing rather than an empty TLV region.

indexer/src/operator/escrow_sweep.rs
Huzaifa696: You were right: 166 falls between the only two sizes Token-2022 allocates, 165 bare or >=170 once a TLV header is paid for. The fixture is now built by the library at 171, with a guard test pinning both legal lengths.

sebatustra (approved):

PR #326, opened 2026-09-18
fix: retune slot-time constants for the 250ms mainnet slot

Huzaifa696:
**Problem.** Several constants still assume Solana’s old 400ms slot time. Mainnet now runs at 250ms, so the indexer alert overstates lag by 1.6x: a “10-second” alert currently fires after about 6.2 seconds. The confirmation polling interval and related comments are also outdated.

**Fix.** The alert multiplier is updated from `0.4` to `0.25`, and dashboard descriptions now clarify that their values are raw slot counts. The default confirmation polling interval is reduced from 400ms to 250ms, including the two `operator-private-channel` configs. `operator-solana` remains at 100ms because it polls PrivateChannel rather than L1. `FINALITY_SAFETY_DELAY` remains unchanged because mainnet finality is still about 8 seconds until Alpenglow activates. Its comments and deadline test now derive from the actual constant. The outdated `example-indexer.toml` estimate is also corrected from approximately 45 days to 29 days.

The alert will now fire at 40 slots instead of 25, preserving the same 10-second threshold. The confirmation window drops from 1.6 seconds to 1 second, increasing `getSignatureStatuses` traffic, so RPC rate-limit headroom should be checked before a busy deployment. Verified with the deferred-remint deadline test, the full sender suite (`264 passed, 0 failed`), and `cargo fmt --check`.

### Coverage Report

| Core | 26,703 | 28,087 | 95.1% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35352781999) |
| Indexer | 42,520 | 46,005 | 92.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35352781999) |
| Gateway | 1,993 | 2,296 | 86.8% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35352781999) |
| Auth | 1,755 | 1,962 | 89.4% | [rust-unit-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35352781999) |
| E2E Integration | 17,752 | 23,092 | 76.9% | [e2e-coverage-reports](https://github.com/solana-foundation/solana-private-channels/actions/runs/35352781999) |
| **Total** | **90,723** | **101,442** | **89.4%** | |

> Last updated: 2026-09-18 14:38:29 UTC by E2E Integration


indexer/src/config.rs
```diff
@@ -389,14 +389,15 @@ pub struct OperatorConfig {
...
     #[serde(default = "default_confirmation_poll_interval_ms")]
     pub confirmation_poll_interval_ms: u64,
 }
 
-/// Default poll interval for `confirmation_poll_interval_ms`, matching Solana's ~400 ms block time.
+/// Default poll interval for `confirmation_poll_interval_ms`, tracking Solana's slot time,
+/// which stepped down from 400 ms and is 250 ms today (200 ms target).
 /// operator-solana overrides this to 100 ms since PrivateChannel confirms faster.
-pub const DEFAULT_CONFIRMATION_POLL_INTERVAL_MS: u64 = 400;
+pub const DEFAULT_CONFIRMATION_POLL_INTERVAL_MS: u64 = 250;
```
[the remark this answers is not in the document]
Huzaifa696: The arithmetic is right (5 attempts sleep 4 times, so 1.6s becomes 1.0s), but the non-idempotent path is not affected. `RetryPolicy::None` covers exactly one operation, `Mint`, which runs on the escrow operator against PrivateChannel at `confirmation_poll_interval_ms = 100`. This PR does not touch that value. Everything running at the retuned 250ms cadence (`ReleaseFunds`, `InitializeMint`, `RotateBitmap`) is `Idempotent`, so a premature `Retry` costs one resend guarded by the nonce, the already-initialized check, or `expected_generation`, not a manual review.

The window is also slot-denominated today: both operators get four sleeps of one block time, so 4 slots before and after is preserved. The fair residual is that RPC round-trip and indexing lag do not scale with slot time, so absolute headroom drops by 600ms. At a typical 50-200ms for that component, 1.0s still covers it.

Decoupling the timeout from the cadence is the right principle, but every way of doing it changes the PrivateChannel Mint window (to 0.7s by raising the attempt count, or 1.6s by deriving from a budget const). That is a semantics change to the one non-idempotent path this PR deliberately avoids, so it belongs in its own PR.

monitoring/dashboards/private-channel-indexer.json
```diff
@@ -313,6 +314,7 @@
       },
       "gridPos": { "h": 8, "w": 8, "x": 0, "y": 48 },
       "id": 30,
+      "description": "Lag in slots, not seconds. At the current 250ms slot time 40 slots is roughly the 10s point where the lag alert fires.",
```
sebatustra, line 317: gauge thresholds still encode 400ms. Panel 30's steps are yellow: 25, red: 75 — exactly the old 10s/30s marks (25×0.4=10, 75×0.4=30). This PR adds a description to that same panel saying "40 slots is roughly the 10s point where the lag alert fires", so the panel now turns yellow at 25 while its own description names 40, and the alert fires mid-yellow. Should be 40 / 120 to keep the 10s/30s meaning.
(in monitoring/dashboards/private-channel-indexer.json:306-307)

Huzaifa696: Good catch, and the description I added made it worse by naming 40 while the gauge turned yellow at 25. Fixed: 25/75 becomes 40/120, keeping the 10s/30s meaning at 250ms.

Left the health dashboard's gauge (50/200) alone since its thresholds are stated in slots with no seconds claim.

sebatustra (approved):

